PR #140, opened 2026-07-08
fix(runtime): restart finished subscriptions on every re-evaluation

akiomik:
## Summary

`Runtime::update_subscriptions` kept a hash of the subscription IDs from the previous re-evaluation and skipped `SubscriptionManager::update` whenever the ID set was unchanged. Because the hash was seeded on the *first* re-evaluation, every later re-evaluation with the same ID set was elided and the manager was never consulted.

This silently broke the manager's documented contract (`src/subscription.rs`):

> Subscriptions whose tasks have finished will be restarted if still present

A finite subscription (e.g. one backed by `stream::once`) that completed was never restarted, even though the application still requested it — the manager only saw it again on the first re-evaluation.

## Fix

The manager's diff is already ID-keyed and cheap, and `subscriptions()` is still only re-evaluated on dirty frames (after a message via `subscriptions_dirty`), so the hash layer saved almost nothing while introducing the bug. This removes the `subscription_ids_hash` field and its hashing logic, letting every dirty frame call `update` and restoring the restart. The re-evaluation gating (`subscriptions_dirty`) is unchanged.

## Verification

Added a regression test (`test_finished_subscription_restarted_with_unchanged_id`) that drives two message-gated re-evaluations of a fixed-ID one-shot subscription and asserts it restarts each time.

- ✅ passes with the fix
- ❌ fails with the hash cache reintroduced (confirmed by temporarily restoring the old behavior)
- The skip only bit on the *second* re-evaluation, which is why the test drives two rounds rather than one.

`cargo test`, `cargo clippy --all-targets`, and `cargo fmt --check` are all clean.

## Scope

Bug fix only — the broader `runtime.rs` refactor (render/subscription scheduler extraction, e2e test migration) is intentionally deferred to follow-up work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Finished subscriptions are now restarted on every re-evaluation while still
+  requested
+  - The runtime cached a hash of the subscription IDs and skipped
+    `SubscriptionManager::update` whenever the ID set was unchanged, which
+    suppressed the manager's documented restart of subscriptions whose tasks had
+    finished (e.g. a finite source completing, or a WebSocket disconnecting)
+  - **Behavior change:** while a subscription ID keeps appearing in
+    `subscriptions()`, a finished stream is restarted after the next message. To
+    stop a source once it finishes, drop it from the returned set; keeping a
+    finished WebSocket subscription in the set is now an implicit reconnect
+
 ### Added
 
 - `Command::without_redraw()` for updates that handled a message without
--- a/benches/subscription.rs
+++ b/benches/subscription.rs
@@ -2,9 +2,6 @@
 //!
 //! These establish a regression baseline for the runtime's subscription work:
 //!
-//! - **id hashing**: the hash the runtime computes over subscription IDs each
-//!   time it decides whether the subscription set changed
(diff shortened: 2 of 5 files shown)


akiomik:
Addressed review feedback in 5b9a034:

- **[CONFIRMED] Dead benchmark** — removed `bench_id_hashing` (and its `subscription_ids`/`DefaultHasher` helpers) from `benches/subscription.rs`; it measured the now-deleted hashing path. Module doc updated.
- **[CONFIRMED] Weak test assertion** — `test_event_loop_process_frame_tick_updates_subscriptions` now observes an actual subscription spawn (parked-stream source + spawn counter), instead of only asserting `!subscriptions_dirty`, which `process_frame_tick` clears unconditionally.
- **[CONFIRMED] Misleading trace event** — renamed `"subscriptions updated"` → `"subscriptions re-evaluated"` (it fires on every dirty frame even when nothing changed; the accurate signals are the manager's `"subscription started"/"stopped"`).
- **[PLAUSIBLE] Behavior change** — documented the restart-of-finished-subscriptions contract on `Application::subscriptions`, called out WebSocket's implicit auto-reconnect in the `websocket` module doc, and added a `Fixed` entry to the CHANGELOG noting the behavior change (drop a subscription from the set to stop it once finished).

`cargo test --all-features`, `cargo clippy --all-targets --all-features`, `cargo fmt --check`, and `cargo bench --bench subscription --no-run` are all clean.

REFUTED items (efficiency regression, `wait_for_spawns` latency, style nits, "no guard against cache reintroduction") not actioned, per the review.

PR #210, opened 2026-07-17
Resolve RFC 0006 open question 8: quit-under-backlog harness and acceptance matrix

akiomik:
## Summary

Resolves RFC 0006 §6 open question 8 (harness follow-up) — the first of the pre-implementation open questions for S4.

**Harness** (`benches/runtime_load.rs`):
- New `quit_*` multi-trial scenarios: `update` returns `Command::quit()` while the shared backlog is still deep, mirroring a real quit key press. The unbiased `select!` makes quit latency a distribution, so each scenario reports tail percentiles across 200 trials (20 for the keyed control).
- Two per-trial metrics: `quit->delivered` — the delivery instant itself, timestamped by a process-global tracing subscriber from the runtime's own `quit signal received` / `keyed quit signal received` debug events (target `tears::runtime`) — and `quit->exit`, which adds teardown (its backlog deallocation scales with depth and is reported separately so it cannot be misread as delivery). If the runtime's message strings change, trials fail loudly as "missing delivery event".
- Coverage: unkeyed quit at ~0 / ~50k / ~300k queued, unkeyed quit under an actively refilling producer, and a keyed-quit control through the private channel.

**RFC 0006 amendment**:
- F6: unkeyed quit delivery is backlog-independent — `quit->delivered` p50 0.11ms, p99 ≤ 0.62ms, max ≤ 0.84ms across 600 loaded trials, no separation between depths. This is the measurement basis for INV-L4: the statistical formulation (b) is satisfiable by the current implementation with no structural change, and a quit-branch scheduling regression is directly detectable because the metric is the delivery instant, not a proxy bound.
- F7: a keyed quit waits for the full shared drain (p50 1.300s at ~50k) — quantified status quo for open question 7.
- §5.1: bounded-vs-unbounded acceptance matrix with the unbounded baselines pinned; the bounded column is filled when the `RuntimeConfig` implementation lands. Queue-depth cells follow the harness's `produced - processed` definition and INV-L1's per-producer accounting: the observable bound is `capacity + concurrent producers` (`capacity + 1` for the single-flood-producer scenarios), not raw channel occupancy.
- F5's stale note (harness cannot measure quit under backlog) removed; OQ8 marked resolved.

The INV-L4 (a)/(b) formulation choice itself is deliberately left open; F6 removes the empirical uncertainty and the choice will be settled together with OQ7.

## Test plan

- `cargo clippy --bench runtime_load --all-features` clean; `cargo fmt` applied.
- All five `quit_*` scenarios run to completion with 0 timeouts and 0 missing delivery events (numbers in the RFC tables are from this run, Apple M1 Max reference machine).
- `steady_20k` re-run matches the RFC §2 reference numbers, confirming the added instrumentation does not perturb existing scenarios.

### DIFF
--- a/benches/runtime_load.rs
+++ b/benches/runtime_load.rs
@@ -9,11 +9,13 @@
 //!
 //! - **Queue depth**: pending messages (produced - processed), sampled while
 //!   the scenario runs; all runtime channels are unbounded today, so this is
-//!   the direct driver of memory growth under overload. `produced` counts
-//!   source emissions, not channel admissions, so under a future bounded
-//!   configuration the observable bound is `capacity + producers` — each
-//!   producer blocked in `send` holds one in-flight message outside the
-//!   channel (RFC 0006 section 5.1).
+//!   the direct driver of memory growth under overload. `produced` counts a
+//!   message when the source stream yields it and `processed` counts it when
+//!   `update` begins (once it has left the channel), so under a future
+//!   bounded configuration the observable bound is `capacity + producers`:
+//!   each producer blocked in `send` holds one in-flight message outside the
+//!   channel, while the message currently inside `update` is already
+//!   excluded (RFC 0006 section 5.1).
 //! - **Update latency**: message emission to `Application::update`.
 //! - **Render latency**: message emission to the first `Application::view`
 //!   call that observes it (input-to-screen staleness).
@@ -68,6 +70,7 @@
 
 use std::fmt::Debug;
 use std::num::NonZeroU32;
+use std::process::ExitCode;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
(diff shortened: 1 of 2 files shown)


akiomik:
Addressed both review findings:

1. **`quit->last-work` was only a lower bound on delivery** — replaced with `quit->delivered` (261cde2): a process-global tracing subscriber timestamps the runtime's `quit signal received` / `keyed quit signal received` debug events, i.e. the instant the event loop's quit branch fires. A quit-branch scheduling regression now shows up directly in the metric, and a change to the runtime's message strings fails trials loudly as "missing delivery event" instead of skewing numbers. Re-measured: loaded p99 ≤ 0.62ms, max ≤ 0.84ms across 600 trials at ~8k–300k depth (still depth-independent); keyed p50 1.300s. RFC tables, F6/F7, and INV-L4 updated (643390d).

2. **Queue-depth acceptance contradicted the harness's depth definition** — the §5.1 cells now state the bound as `app_channel_capacity + concurrent producers` (`capacity + 1` in the single-flood-producer scenarios), matching `produced - processed` semantics and INV-L1's per-producer accounting, with a note below the matrix spelling out the compliant vs. regression readings. The harness documents the same accounting at the queue-depth metric and `Metrics::queue_depth`.

akiomik:
Addressed both findings from the second pass:

1. **[P1] Depth accounting missed the consumer's in-flight message** — `processed` is now counted at `update` entry, once the message has left the channel (d95d5f5), so the depth excludes the message being processed and the observable bound stays `capacity + producers` (`capacity + 1` here). Without this, a compliant bounded implementation could read `capacity + 2` (full channel + producer refilling the freed slot + in-update message) and be flagged as a regression. The §5.1 note now spells out both in-flight positions explicitly: producer-held message *included*, in-update message *excluded*, and why counting at update completion would misflag (bd6de5f). Quit latency and reported depth values are unchanged by the reorder.

2. **[P2] Failed trials no longer exit green** — `main` returns `ExitCode::FAILURE` when any quit trial times out or misses its delivery event, so a tracing-message change (or any partial sample behind the RFC's "≥ 200 trials" condition) turns the CI Benchmarks check red instead of producing percentiles from fewer trials. An argument list matching no scenario also fails now. Verified locally: a bogus scenario name makes `cargo bench` fail with exit status 1; the success path (`quit_idle`, `quit_backlog_50k`, 400 trials, 0 failures) still exits 0.

PR #214, opened 2026-07-17
docs(rfc): resolve OQ6 — no cross-channel fairness policy

akiomik:
## Summary

Resolves RFC 0006 open question 6 (cross-channel fairness): **bounded keyed-delivery latency is not a goal — no fairness policy is added, and INV-14 stays exactly as RFC 0003 states it, in both delivery modes.** This settles the last of the three contracts the decision-scope note flagged for settling before the post-0.10.0 implementation (OQ3 and OQ7 were the other two).

## The decisive fact

Shared-first pull *is* the cancellation mechanism, not a scheduling preference layered on top of one. A shared message is opaque `Msg` at the pull point; a cancellation exists only once `update` has consumed that message and its cancel list reaches `enqueue_command`. So "deliver keyed output unless a cancel is pending" is not implementable: honoring a cancel still queued as an unprocessed shared input requires processing every ready shared input first — which is precisely INV-14. Any policy that *guarantees* a keyed-delivery bound must therefore deliver keyed output past inputs that could still cancel it, trading exactly the property keying exists to buy. The question's constraint ("relax INV-14 while preserving cancel-before-delivery") has no non-vacuous solution.

Candidate policies considered and excluded in §4.7: shared-pull quota per batch window (the question's own example), deadline/aging, weighted round-robin, cancel-aware scan (not implementable — cancels are invisible in messages), and per-delivery liveness check (already exists; sees only dispatched cancels).

## The trade, stated explicitly (revised after review round 1)

- **What is forgone: keyed liveness under sustained shared readiness, in both modes.** Bounded capacity does not restore it — a waiting producer refills each freed slot, so shared readiness persists at any capacity (§4.3); backpressure bounds memory and shared drain-side latency (INV-L1/L3), not keyed deferral. The deferral is indiscriminate about the shared message's source (user input, subscription output, unkeyed command output all travel the shared channel, §1.1) — any of them can be the one whose `update` returns the cancel, so a still-wanted keyed output can wait indefinitely behind unrelated hot-subscription traffic. This is an accepted, stated cost: cancellation correctness over keyed liveness.
- **Where applications live, no policy is needed**: in-capacity keyed delivery is sub-ms (`keyed_steady` p50 0.10ms); under unbounded overload the whole UI is seconds stale anyway (F3); under bounded overload the persisting deferral is the visible cost — remedied per command and at the source, not by a policy.
- **The API already prices the trade per command**: unkeyed output waits its arrival-order turn in the shared FIFO (INV-L3-bounded drain-side in bounded mode); keying opts into cancellability and with it deferral. Same structure as §4.6's quit guidance.
- **A policy would add contract surface without adding a remedy the API lacks**; keyed liveness under load comes from source pacing (application-owned, same trust class as the producer-count premise, §4.5).

## Enforcement

No new invariant (nothing new to pin): the content is INV-14 (imported via INV-L5; RFC 0003 already disclaims bounded shared/keyed fairness) plus INV-L10/INV-L11. Policy absence is checked structurally at the seams a policy would occupy — both `AppInputs` pull paths and the micro-batch loop — with the shared-wins unit tests as regression checks, not proofs (a quota larger than any finite test's pull count passes it).

## Changes

- New §4.7 (rationale, excluded candidates, explicit forgone-liveness statement, consequences); amendment header, decision-scope note, F4, §4.3, §4.6 latency bullet, INV-L3, §5 closing paragraph, §5.1 `keyed_overload`/`keyed_isolation` rows, §6 OQ6 resolution.
- New unit test `shared_input_wins_the_nonwaiting_pull_when_keyed_output_is_also_ready`: the quit pair already covered both pull paths (INV-L11), but the keyed-*message* form only had the blocking-path test — the seam-coverage gap the pre-review checklist flags.
- Pre-review checklist: two new prompts from review round 1 (one-row implicature under item 1; own-derivation consistency under item 4).

## Pre-review checklist

Run per docs/rfcs/pre-review-checklist.md: quantifiers checked against the channel/pull-site inventory; adversarial candidates recorded in §4.7; enforcement classes declared with named seams; code claims re-verified against `core.rs` (cancel visibility), `app_input.rs`/`runtime.rs` (pull sites, batch loop), and tokio-stream 0.1.18 `StreamMap` (randomized start position); cross-reference sweep for "open question 6" done; `typos`, `git diff --check`, fmt, clippy, and the full test suite are clean.

### DIFF
--- a/docs/rfcs/0006-runtime-load-control.md
+++ b/docs/rfcs/0006-runtime-load-control.md
@@ -26,18 +26,30 @@
   private-channel routing, keyed-quit ordering, and shared-first
   precedence this rests on are pinned as INV-L10/INV-L11, and bounded-mode
   keyed-quit latency is capacity-dependent, so F7's regression check is
-  scoped to the unbounded default (sections 4.2, 4.6, 5, 5.1)
+  scoped to the unbounded default (sections 4.2, 4.6, 5, 5.1). 2026-07-17
+  — open question 6 resolved: no cross-channel fairness policy — bounded
+  keyed-delivery latency is not a goal of this contract. Cancellation is
+  an `update` decision, so a pull point cannot know which ready shared
+  inputs are cancels; shared-first pull (INV-14) is therefore the
+  implementable form of cancel-before-delivery itself, and any policy
+  that guarantees a keyed-delivery bound (quota, aging, weighted
+  round-robin) must deliver keyed output past inputs that could still
+  cancel it — trading exactly what keying buys. INV-14 stays unrelaxed in
+  both delivery modes, the keyed-probe scenario stays measurement-only,
+  latency-critical output is directed to the unkeyed path, and the
+  shared-wins unit tests now cover both pull paths for keyed messages
+  (sections 4.3, 4.7, 5, 5.1)
 
 > **Decision scope.** Section 3 (the release-gate verdict) is the part of this
 > draft that 0.10.0 depends on, and it is final unless the contract design in
 > sections 4–6 turns out to be unimplementable without breaking the public
 > API. Sections 4–6 fix the direction of the load-control contract but their
 > details (capacities, batching caps, per-source classes) remain open until
-> implementation. In particular, cross-channel fairness (section 4.3)
-> remains a contract to settle before the post-0.10.0 implementation; the
-> quit delivery path and the exact scope of the memory bound (INV-L1) were
-> the other two such contracts and are now settled (open questions 7 and 3,
-> sections 4.6 and 4.5). None of them gates the 0.10.0 release.
+> implementation. The three contracts flagged for settling before the
+> post-0.10.0 implementation — the exact scope of the memory bound
+> (INV-L1), the quit delivery path, and cross-channel fairness — are all
+> now settled (open questions 3, 7, and 6; sections 4.5, 4.6, and 4.7).
+> None of them gated the 0.10.0 release.
 
 ## Summary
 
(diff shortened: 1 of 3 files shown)


akiomik:
Review round 1 addressed in 07ef153.

**P1 (bounded mode does not dissolve keyed starvation)** — accepted as a merge blocker and re-derived rather than patched (checklist item 5). The "regime bounded mode exists to exit" bullet and the §6 echo are gone. The declining-the-goal block now leads with the forgone half of the trade, stated explicitly: keyed liveness under sustained shared readiness is given up in **both** modes — bounded capacity does not restore it (§4.3's refill argument), and backpressure bounds memory and shared drain-side latency (INV-L1/L3), not keyed deferral. The bounded-overload case is now described honestly: UI staleness bounded by capacity (INV-L3) while keyed deferral persists as the visible cost, remedied per command (unkeyed) and at the source (pacing), not by a policy. Propagated to the amendment header, §6, and the documentation-guidance bullet.

**P2 (shared traffic ≠ user controls)** — accepted. The "exactly the user's controls" sentence is replaced with the inventory-honest statement: the deferral is indiscriminate about the shared message's source (terminal-event subscription output, other subscription output, unkeyed command output — §1.1), any of which can be the input whose `update` returns the cancel; a still-wanted keyed output can therefore wait indefinitely behind unrelated hot-subscription traffic. The incompatibility argument never needed the UX claim; the UX claim is no longer made.

Both defect classes are added to the pre-review checklist in the same commit, per its living-document rule: a "one-row implicature" prompt under item 1, and an own-derivation-consistency prompt under item 4 (a clause crediting a mechanism with resolving a failure mode must be re-checked against the section that analyzed that mechanism).

PR #218, opened 2026-07-18
feat(command,subscription): add Command::scoped and Subscription::scoped (RFC 0005 Phase B)

akiomik:
## Summary

- Implements RFC 0005 Phase B: `Subscription::scoped` and `Command::scoped`, the composition-boundary operators that qualify a subscription's or command's lifecycle identity with one structural scope segment.
- `SubscriptionId` and `CommandId` each gain an ordered `ScopePath` alongside their existing structural key. Scope segments are order-sensitive and collision-safe (same guarantees as Phase A's structural key), and are stored in a field separate from the local key so a scoped identity can never be confused with a plain local key of matching shape.
- `Command::scoped` qualifies the keyed spawn id (if `cancellable`/`cancellable_with` was already called) and every explicit cancel id already present at its call boundary — it does not retroactively cover ids attached by a later modifier, consistent with RFC 0003's last-call-wins rule. `cancellable`/`cancellable_with` rustdoc documents both call orders with runnable examples.
- Additive change; unscoped subscriptions and commands keep their existing 0.10.0 behavior unchanged.

## Test plan

- [x] `cargo test --all-features` (341+ tests, including new structural, subscription-manager, and runtime-level scope-isolation tests)
- [x] Sanity check: temporarily neutered `Command::scoped` to confirm the new tests fail as expected, then restored the real implementation
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo test --test api_surface -- --ignored` (public API shape/reachability)
- [x] `cargo bench --bench subscription --features bench-internals -- --test` (new scoped steady-state case compiles and runs)
- [x] `cargo doc --all-features --no-deps`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `Subscription::scoped` and `Command::scoped` qualify a subscription's or
+  command's lifecycle identity with one structural scope segment, so composed
+  child instances that reuse the same local source/key or `CommandId` no
+  longer alias each other's lifecycle (RFC 0005 Phase B)
+  - `Command::scoped` qualifies the keyed spawn id (if `cancellable`/
+    `cancellable_with` was already called) and every explicit cancel id
+    already present at the call boundary; it does not retroactively cover
+    ids attached by a later modifier call — see the ordering examples on
+    `Command::cancellable`'s rustdoc
+  - Scoping is additive: unscoped subscriptions and commands keep their
+    existing 0.10.0 behavior unchanged
+
 ## [0.10.0] - 2026-07-17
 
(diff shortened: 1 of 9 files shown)


akiomik:
Addressed the reported gap (8958786).

- `src/subscription.rs`: added tests using `ManagerCollisionKey` (a constant-hash scope type) to verify that the same local id starts independently — and stops independently — across two hash-colliding scopes
- `src/runtime.rs`: added `CollidingScope` and three tests verifying that command replacement (`CancelInFlight`), `KeepInFlight` suppression, and explicit cancellation each stay isolated across hash-colliding scopes

To validate the tests themselves, I temporarily degraded `ScopePath` equality to a hash-only comparison (the exact implementation the RFC forbids). All 5 new tests failed while the existing non-colliding-scope tests still passed, confirming these tests actually close the gap. The implementation has been restored to its correct state.

PR #220, opened 2026-07-18
docs(rfc): add RFC 0007 draft — RuntimeConfig API and acceptance parameters

akiomik:
## Summary

RFC 0006 names "the separate `RuntimeConfig` RFC" as the sole remaining prerequisite of the load-control implementation PR (its §§3.2, 6). This draft is that document. It discharges every delegated obligation, enumerated as an inventory in §1 with the section that closes each one:

- **Public API (§2)** — `RuntimeConfig` is the runtime's construction-time configuration: it carries the frame rate plus RFC 0006 §4.1's three load controls under their working names. Private fields; `RuntimeConfig::new(frame_rate)` as the sole constructor (deliberately no `Default`: the crate has no default frame rate, and inventing one inside a derive would smuggle in a policy value); infallible consuming setters in the crate's combinator convention. `Runtime::with_config(flags, config)` consumes it; `Runtime::new` is unchanged and becomes a literal delegation — the structural seam INV-L6's check goes through. Rejected shapes (RFC 0006's three-argument example, an optional frame rate with a defaulted value) are recorded with reasons.
- **Recommended defaults, documentation only (§3; RFC 0006 OQ1 + the default half of OQ2)** — sizing rule (capacity × per-message drain cost) plus starting values `app = 1024`, `keyed = 16`, `batch_max_messages` unset, all derived from the RFC 0006 §2 measurements; lands as rustdoc together with the §§4.5–4.7 guidance notes.
- **Restart-rate control (§4; OQ5)** — stays subscription-level; `RuntimeConfig` carries no restart-rate field and reserves none.
- **Bounded acceptance-run parameters (§5)** — one configuration under test (equal to the documented starting point, fixed here before implementation); bounded quit scenarios redefined around blocked-producer count (`quit_blocked_1`, `quit_blocked_64`) and channel-full churn (bounded `quit_overload`) per the RFC 0006 review follow-up; trial counts carried from INV-L4's normative floor; `keyed_isolation` pinned at 8 saturated keys.
- **CI smoke profile (§6)** — adopted: a `--smoke` harness profile gating on completion and lossless bookkeeping only, no latency assertions, so a slow CI machine cannot fail it on speed.

INV-C1..C4 pin the surface with declared enforcement classes; the load-control semantics stay in RFC 0006 by reference, so this RFC cannot amend them.

## Pre-review checklist

Run per `docs/rfcs/pre-review-checklist.md`. Notable outcomes: one code-claim finding was caught and fixed before this PR (the `api_surface` test enforces canonical-path/prelude rules, not a surface snapshot — INV-C4's check is review-only and the text now says so); the §2 constructor-integration clause was re-derived from premises after review feedback moved the frame rate into the config type, and the checklist was re-applied to the rewritten section.

## Review focus

- §2.1's scope decision: the frame rate lives inside `RuntimeConfig` (a type named for runtime configuration should carry the runtime's settings); the alternative shapes and their rejection rationale are in §2.2.
- §5.1's choice to make the configuration under test equal to the documented starting point.
- §5.2's bounded quit scenario definitions and producer counts (1 / 64).

### DIFF
--- a/docs/rfcs/0007-runtime-config.md
+++ b/docs/rfcs/0007-runtime-config.md
@@ -195,24 +195,40 @@ one exception that the sizing rule below must stay consistent with RFC
 ### 3.1 Sizing rule and starting values
 
 The documented rule: a bounded channel's capacity buys burst absorption and
-costs worst-case latency — once the queue is full, a newly accepted
-message waits up to `capacity × per-message drain cost` before reaching
-`update`. Per-message drain cost is application-dependent and measurable
-(RFC 0006 §2: ~2.2µs/message at 2µs `update` cost, ~26µs/message at 25µs —
-drain time tracks `update` cost, which the application controls).
+costs queueing latency — once the queue is full, a newly accepted message
+waits roughly `capacity × per-message drain cost` before reaching
+`update`, where the drain cost is the *observed average* service time of
+the application's own loop, application-dependent and measurable (RFC 0006
+§2: ~2.2µs/message average at 2µs `update` cost, ~26µs/message at 25µs —
+drain time tracks `update` cost, which the application controls). The
+product is an estimate on the measured workload, not a worst-case bound:
+an average cannot bound the tail, and neither RFC 0006 nor this RFC
+defines a per-message worst-case service time to build a hard bound on.
+An application that needs one derives it from its own measured tail
+service time, not from this rule.
 
-Documented starting values, derived from the section 2 reference numbers:
+Documented starting values, each with its basis stated:
 
-- **`app_channel_capacity = 1024`.** Large enough that the in-capacity
-  regime never engages backpressure (`steady_200k` peaks at depth 400), and
-  small enough that a full queue adds at most ~27ms (~1.6 frame periods at
-  60 FPS) even at the harness's heavy 25µs `update` cost. Applications with
-  larger expected bursts scale up by the rule above; `burst_200k` shows the
-  cost of undersizing is producer wait, not loss.
-- **`keyed_channel_capacity = 16`.** Keyed outputs are per-command result
-  streams, low-rate by construction in the measured workloads
-  (`keyed_steady`); 16 bounds each command's buffer share of `m × capacity`
-  (RFC 0006 R1) while leaving headroom above the harness's probe cadence.
+- **`app_channel_capacity = 1024`.** Measurement-derived at both ends:
+  large enough that the in-capacity regime never engages backpressure
+  (`steady_200k` peaks at depth 400), and small enough that a full queue
+  adds ~27ms of estimated queueing latency (~1.6 frame periods at 60 FPS)
+  at the harness's observed average drain rate under its heavy 25µs
+  `update` cost — an estimate per the rule above, not an upper bound.
+  Applications with larger expected bursts scale up by the same rule;
+  `burst_200k` shows the cost of undersizing is producer wait, not loss.
+- **`keyed_channel_capacity = 16`.** A margin choice, stated as such —
+  not measurement-derived: in the measured workloads a keyed channel never
+  buffers more than one message (`keyed_steady`'s probe fires every 25ms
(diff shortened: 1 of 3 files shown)


akiomik:
## Review findings: verification and fixes (21ea69e)

All six findings verified against the RFC text, the harness's depth definition, and `ci.yml`; all six confirmed valid and fixed. Per finding:

| # | Finding | Verification | Fix |
| --- | --- | --- | --- |
| 1 | P1 — depth bound wrong for 64 producers | Confirmed: harness depth is `produced - processed` and counts each blocked producer's in-flight message; RFC 0006 §5.1's own accounting is `capacity + concurrent producers` | §5.2 now separates channel occupancy (≤ capacity) from observed depth (≤ capacity + producers); `quit_blocked_64` explicitly instantiates to ≤ 1088; §5.3's `1024 + 1` is scoped to single-flood rows and names the quantity it bounds |
| 2 | P1 — smoke lossless check non-discriminating | Confirmed: "accounted for at shutdown" was undefined, and quit scenarios legally discard undelivered messages, so `produced - processed` cannot separate legal discard from illegal drop | §6 pass/fail split by discriminating power: draining scenarios assert `produced == processed == scripted total` (an INV-L2 violation leaves `processed` short); quit scenarios assert completion only, with the indistinguishability stated as the reason |
| 3 | P1 — average drain rate presented as worst-case | Confirmed: ~26µs/message is the observed average over a 50k drain (F7); `capacity × average` bounds nothing | §3.1's rule is now an estimate on the measured workload; "at most ~27ms" removed; the text states a hard bound requires a measured worst-case service time neither RFC defines |
| 4 | P2 — smoke rationale ignores current CI | Confirmed against `ci.yml`: the Benchmarks job already runs the full harness on every push (`cargo test --bench runtime_load`, custom `main` ignores test filtering) | §6 rewritten: the smoke profile *replaces* the Benchmarks job's `runtime_load` invocation (the `subscription` bench is unchanged); rationale is now wall-time growth of the full scenario set, not absence from CI |
| 5 | P2 — INV-C3 signature check admits panicking setters | Confirmed: the adversarial setter with an infallible signature and a `panic!` on a chosen value passes the stated check | INV-C3's structural check extended to the four function bodies: plain field writes, no branching on argument values, no panic paths |
| 6 | P2 — `keyed_channel_capacity = 16` not measurement-driven | Confirmed: `keyed_steady`'s 25ms-cadence probe never buffers two messages, so the cited measurement cannot distinguish 16 from 1 | §3.1 restates 16 as a margin choice with the trade-off spelled out (≤16-message burst absorption vs. 16 messages per command of the `m × capacity` total); a keyed-burst scenario is named as the additive path to a measured basis |

## Could the pre-review checklist have prevented these?

Four of six fall under existing items — these are process failures (the checklist pass on this draft was too shallow on §§3, 5, 6), not checklist gaps:

- **Finding 1 → item 1** (quantifiers against the inventory): walking the bound claim against every scenario row — including the rows this RFC itself added with different producer counts — catches it. Measurement/scenario tables are explicitly named as inventories there.
- **Finding 2 → item 2** (minimal-effort adversary against definitions of done, the PR #215 prompt): "processed or accounted for" is satisfied by any implementation, including one that drops under backpressure.
- **Finding 4 → item 4** (code claims verified against code): the claim "the harness otherwise runs only on the reference machine" was about `ci.yml` and was never checked against it.
- **Finding 5 → item 2** (adversarial counterexample per invariant): the panicking-setter adversary is exactly the "does the named thing with arbitrary values" shape.

Two are defect classes the checklist did not name; per its living-document rule they are added in this PR (same commit), under item 4 as "measurement citations prove only what they measured":

- **Finding 3**: an observed average or p50 supports an estimate, never an *at most* — `n × average cost` does not bound a tail.
- **Finding 6**: a value presented as measurement-driven must cite a measurement that *distinguishes* the chosen value from the alternatives it rejects; a workload indifferent between the candidates sizes nothing.

PR #232, opened 2026-07-24
refactor(observability): emit producer gauges outside the counter lock

akiomik:
## Summary

The producer-gauge path in `src/runtime/load.rs` updated a counter and emitted the `tears::runtime::load` event under the same `std::sync::Mutex`. That serialization was two footguns:

- **(a) producer stall** — a subscriber's event handler ran *under* the lock, serializing every producer and the event loop behind a slow handler.
- **(b) self-deadlock** — a handler that re-entered the runtime deadlocked on the non-reentrant `Mutex`.

This snapshots the four-field `Gauges` (`Copy`) under the lock and emits **after** releasing it. Snapshotting under the lock preserves the value-fidelity guarantee (the event carries the value the change reached, not a later re-read); emitting outside it keeps the subscriber's handler off the counter lock. No `seq` number is introduced — cross-producer arrival order is explicitly *not* guaranteed.

Caught before release: RFC 0006/0007 are Implemented but still `[Unreleased]`, so this ships in the same additive 0.10.x patch. (Was tracked as the "#\2 gauge event delivery under the lock" review-hold item.)

## Contract change (RFC 0006)

The lock/snapshot is *mechanism* and stays in code comments; the *observable guarantees* are contract, so §4.4 and INV-L13 now pin them:

- **Value fidelity** — each change emits one event carrying the value it reached, so order-insensitive aggregates (high-water marks) are exact. Enforcement split: **behavioral** for the single-threaded emitted-sequence, **structural** for the concurrency component (a single-threaded test cannot refute a lone-atomic re-read).
- **Non-blocking emission** — emitting holds nothing a handler could contend for; **structural**, reviewed at the two emit sites (`LoadObserver::step`, `set_keyed_entries`).
- Both state their **negative space**: cross-producer arrival order is not guaranteed.

## Docs (second commit)

The amendment surfaced a reusable scope distinction, recorded where future RFC authors decide what to write:

- `README.md` — new **"What an RFC pins"** section: mechanism-out / observable-property-in, the from-scratch reimplementation test, and two corollaries (pin negative space; property and mechanism can carry different enforcement classes).
- `pre-review-checklist.md` — item-2 adversary for the concrete defect this amendment's self-review caught (a concurrency-only hazard behind a single-threaded-deterministic test).

## Verification

- `cargo test --lib runtime::load` — 3 pass (existing value-fidelity / all-fields / levels tests unchanged and green)
- `cargo clippy --all-targets`, `cargo fmt --check`, `typos`, `git diff --check` — clean
- pre-review checklist run against the changed claims; the behavioral-vs-structural finding above was caught and fixed in this branch

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     (`pulled`/`updated`/`shared_pending`), a bounded-mode capacity-wait event
     (`channel`/`wait_us`), and producer-count gauges (`subscriptions`,
     `unkeyed_commands`, `keyed_commands`, `blocked`)
+  - The producer-gauge events carry two subscriber-facing guarantees: value
+    fidelity (each change emits the value it reached, so high-water marks and
+    other order-insensitive aggregates are exact) and non-blocking emission (a
+    subscriber's handler may run long, or re-enter the runtime, without
+    stalling or deadlocking producers); cross-producer arrival order is not
+    guaranteed
   - Additive and non-breaking: `Runtime::new` is unchanged, and a
     load-control-unset `RuntimeConfig` reproduces the previous unbounded
     delivery path exactly
--- a/docs/rfcs/0006-runtime-load-control.md
+++ b/docs/rfcs/0006-runtime-load-control.md
@@ -491,9 +491,24 @@ as INV-L13 (section 5):
   (active forwarding tasks), `unkeyed_commands` (running unkeyed command
   tasks), `keyed_commands` (active keyed entries), and `blocked`
   (producers currently awaiting capacity; always 0 in unbounded mode).
-  These gauges are how the application-owned producer-count premise stays
-  observable rather than enforced (section 4.5), including the
-  blocked-producer anti-pattern named there.
+  Two delivery guarantees hold for these events, so that an implementation
+  mistake in the emission cannot silently degrade a subscriber. *Value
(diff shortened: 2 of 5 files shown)


akiomik:
Closing without merging. Review found the emit-outside-lock change is unsound at the design level, not just in wording:

- The gauge event stream provides **ordered, last-event-is-current** delivery, and that is a load-bearing (though previously unpinned) contract. Two repo consumers depend on it: the bench teardown barrier `await_quiescence` (`benches/runtime_load.rs`, last-writer-wins on `LIVE_PRODUCERS`) and the RFC 0007 §5.2 acceptance predicate ("observe the gauges at the quit instant; a barrier may not substitute for the observation").
- Emitting outside the lock lets a stale snapshot arrive after the final one, which can permanently wedge `await_quiescence` (5s timeout aborts the whole bench) and skew §5.2 trial validity. The amendment's own "cross-producer arrival order is not guaranteed" directly contradicts §5.2, which this PR did not touch.

Removing this property to mitigate two low-severity footguns (slow-subscriber producer stall; re-entrant deadlock) is a net-negative trade. Preserving the footgun fixes while keeping ordered/current-value observation requires sequence numbers or a direct `LoadObserver` read API plus a §5.2 protocol revision — a runtime-contract redesign that belongs in S10, not a small additive patch.

Deferred to S10. The valuable finding — that ordered/current-value gauge delivery is an implicit load-bearing contract coupling §4.4 and §5.2 — is recorded as an S10 input.

PR #236, opened 2026-07-24
refactor(runtime): dispatch gauge events off the lock behind a re-entrancy funnel

akiomik:
## Summary

Moves the producer-gauge `tracing` dispatch out from under the runtime's gauge mutex and makes that move re-entrancy-safe, completing what RFC 0006 §4.4 anticipated when it introduced the gauge-event `seq` ordering field.

Previously the gauge event was captured **and** dispatched while holding the gauge lock, so a slow subscriber stalled every producer on that lock and a subscriber that re-entered the runtime deadlocked on it. RFC 0006 §4.4 had flagged two re-entrancy hazards as prerequisites to moving the dispatch off the lock:

- **Unbounded recursion** under a global `tracing` dispatcher, when a subscriber that itself causes a gauge change re-enters the emit path.
- **Nested-event drop** under a scoped dispatcher's re-entrancy guard, which breaks value fidelity (a reached value gets no event).

## Approach

- Split the emit path into a **snapshot-under-lock** (`Gauges::capture`, which bumps `seq`) and an **off-lock** `GaugeSnapshot::dispatch`. The `seq`/snapshot capture stays atomic under the lock; only the dispatch moves off it.
- Funnel all dispatch through a **single drainer**: the first producer to find no drainer running claims the role and dispatches in a loop, while every concurrent or re-entrant producer only enqueues its snapshot in `seq` order and returns. Delivery is thus iterative and never nested inside a `tracing` dispatch, so neither hazard can arise.
- A `DrainGuard` releases the drainer role if a subscriber panics mid-dispatch, so a panic cannot wedge the funnel shut — the off-lock analogue of the existing poisoned-lock recovery in `lock`.

## Contract & docs

The observable schema is unchanged: every gauge change is still delivered as its own `seq`-ordered event, and consumers still order by `seq`, never by arrival. The `[Unreleased]` CHANGELOG entry already states that, so **no CHANGELOG change** is needed.

The now-stale prose in RFC 0006 §4.4 and RFC 0007 §5.2 — which described gauge emission as happening under the lock, with arrival order coinciding with `seq` — is updated to present tense, keeping the delivery mechanism as an implementation concern and stating only the observable guarantee. Per the pre-review checklist's coincidental-property item, the two consumers that read gauges by greatest `seq` (`benches/runtime_load.rs` teardown barrier, RFC 0007 §5.2 predicates) were confirmed already migrated off arrival order.

## Tests

- New gauge-layer test: a subscriber causes a gauge change while handling one; the re-entrant event must be delivered with a distinct `seq` and its reached value — which the old under-lock dispatch would deadlock on and a naive off-lock dispatch would drop.
- `set_default_subscriber` added to the crate test support so a bespoke subscriber installs cache-safely (the same callsite-interest handling `TraceRecorder::set_default` uses).

## Verification

`cargo fmt --check`, `cargo clippy --all-targets --features bench-internals` (0 warnings), full `cargo test` (0 failed), `cargo doc` (public + CI mode), `git diff --check`, and `typos` all clean.

### DIFF
--- a/docs/rfcs/0006-runtime-load-control.md
+++ b/docs/rfcs/0006-runtime-load-control.md
@@ -504,23 +504,23 @@ as INV-L13 (section 5):
   of this contract**, so a consumer that reads "the value on the most
   recently *arrived* event" is relying on an ordering the schema does not
   provide. Ordering by an explicit `seq` rather than by arrival is
-  deliberate: because the snapshot-and-`seq` capture is atomic while only
-  the emit need move, it lets a later change emit gauge events without
-  holding the runtime's gauge lock across the `tracing` dispatch, so a slow
-  subscriber can no longer stall producers on that lock and a subscriber
-  that re-enters the runtime can no longer deadlock on it — without
-  breaking any `seq`-ordered consumer. That move does **not**, on its own,
-  make a subscriber that itself *causes* a gauge change safe: such a
-  subscriber re-enters the emit path, risking unbounded recursion under a
-  global `tracing` dispatcher, or — under a scoped one — a nested event
-  silently dropped by `tracing`'s re-entrancy guard, which breaks value
-  fidelity. Resolving that re-entrancy is a prerequisite of any future
-  off-lock change and is out of scope for the `seq` field, which secures
-  only the ordering; it is recorded here so the off-lock change cannot read
-  the stall/deadlock sentence as a claim that re-entrancy is already safe.
-  The initial implementation still emits under that lock, so `seq` and
-  arrival order coincide there; consumers must not depend on the
-  coincidence.
+  deliberate: it lets the runtime dispatch a gauge event without holding
+  the gauge lock across the `tracing` dispatch — the snapshot-and-`seq`
+  capture stays atomic under the lock while only the dispatch moves off
+  it — so a slow subscriber no longer stalls producers on that lock and a
+  subscriber that re-enters the runtime no longer deadlocks on it, without
+  breaking any `seq`-ordered consumer. Off-lock dispatch is
+  re-entrancy-safe by construction: a subscriber that itself *causes* a
+  gauge change re-enters the emit path, and dispatched inline that would
+  recurse without bound under a global `tracing` dispatcher, or have its
+  nested event silently dropped by a scoped dispatcher's re-entrancy guard
+  (breaking value fidelity); the runtime instead delivers a nested change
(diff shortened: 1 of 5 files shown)


akiomik:
Thanks for the review — all four addressed in a7d98c4.

**Point 1 (DrainGuard misfire) — fixed.** Replaced `thread::panicking()` with an `armed` flag that the drain loop disarms the instant it relinquishes the role normally, so a still-armed drop means *and only means* an unwinding `dispatch`. This stays correct when `emit` runs during an unrelated unwind (`GaugeGuard::drop`), which `thread::panicking()` could not distinguish. The one-drainer invariant and value fidelity are no longer at risk in the race you described.

**Point 2 (arrival vs seq) — fixed.** You're right: the single-drainer funnel serializes dispatch in `seq` order, so arrival order *does* still coincide with `seq` in the current implementation. Corrected all three spots (module doc, RFC 0006 §4.4, RFC 0007 §5.2) to the contract statement — the schema does **not guarantee** they coincide, so consumers must order by `seq` even where an implementation happens to deliver in `seq` order — rather than the false "does not coincide."

**Point 3 (cross-thread attribution) — added.** Module doc now notes that a snapshot is dispatched by whichever thread is draining, so the event fires under that drainer's thread and current span context: a global subscriber sees only a schema-preserving attribution change (INV-L13 unaffected), while a thread-scoped dispatcher can see another thread's gauge change or miss its own.

**Point 4 (panic-recovery test) — added.** New `a_subscriber_panic_mid_dispatch_does_not_wedge_the_funnel`: a subscriber panics while dispatching the first gauge event; the test asserts a later change still dispatches (`seen_after == [2]`) rather than piling up behind the unwound drainer. It fails if the `DrainGuard` recovery is removed or the arming is broken.

One caveat worth flagging: this recovery test pins the *recovery-on-panic* direction, but it passes under both the old `thread::panicking()` and the new `armed` guard — in a single-threaded panic both recover. 指摘 1's actual misfire (false recovery during an *unrelated* unwind corrupting a *concurrent* drainer's state) is inherently a multi-threaded race with no corrupting single-threaded manifestation, so it isn't deterministically unit-testable; the `armed` fix is correct by construction (armed-at-drop ⟺ dispatch unwound). Happy to add a loom test if you'd prefer that pinned too.

Verification: `cargo fmt --check`, `clippy --all-targets --features bench-internals` (0 warnings), full `cargo test` (0 failed), `cargo doc` (CI mode), `git diff --check`, `typos` all clean.

PR #254, opened 2026-07-25
feat(testing): give TestStore a controlled time context and advance (stage 2)

akiomik:
## Summary

Implements the RFC 0008 stage-2 amendment (#253), consuming RFC 0009's Clock DI contract. Stacked on #253 — review that first.

- **Store-owned controlled time context.** `TestStore::new` builds a current-thread executor context with the clock started paused and only the time driver enabled (no I/O). Every poll under the §4.1 budget enters it, so a time-gated leaf finds the paused clock. The INV-T10 construction check is behaviorally unchanged (any entered runtime still panics) with its rationale revised: the store's own context must be the only executor context in play.
- **`advance(duration)`.** Anchors first — its anchoring scan polls every pending leaf without buffered output exactly once in enqueue order (INV-T13) — then moves the clock by exactly `duration`. It delivers nothing; readiness is observed at the next scan that polls the leaf. Because `advance` is the only clock mover and anchors before moving, a timeout leaf's deadline counts from its enqueue-time virtual now, scan-order-independent.
- **Timeout / retry-backoff leaves** become ordinary pending leaves driven through `receive` flow (INV-T12); the cannot-become-ready class narrows to I/O-dependent leaves, which keep the missing-reactor failure at first poll.
- **`test-util` feature** joins the unconditional `tokio` dependency features (RFC 0009 §5.1); the duplicate dev-dependency entry is dropped. Pausing stays opt-in at runtime construction, so production is unchanged (RFC 0009 INV-C4).

## Tests

New: INV-T12 (timeout pending across failing scans and sub-deadline advances; exact-deadline delivery; retry-backoff gating), INV-T13 (first-poll vs construction anchoring), the `advance` half of INV-T4's poll budget, equal-deadline enqueue-order delivery (§4.2 / RFC 0009 §3.4), anchoring-scan buffering, the unadvanced-timeout finish leak, and post-quit `advance` failure. The stage-1 reactor-dependent tests are recast over I/O-dependent leaves per the amended §4.3.

## INV-C4 verification (test-util flip)

- `cargo clippy --all-targets -- -D warnings` and the full test suite (incl. doctests) green.
- RFC 0007 §6 smoke profile (`cargo bench --bench runtime_load -- --smoke`) passes after the flip (exit 0).
- Full RFC 0006 load harness run recorded below.

_(Full load-harness numbers appended once the run completes.)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `tears::testing::TestStore`, a deterministic, executor-free test harness
-  that drives an `Application`'s `update` transitions and immediately ready
-  effects synchronously (RFC 0008, stage 1)
+- `tears::testing::TestStore`, a deterministic test harness that drives an
+  `Application`'s `update` transitions, immediately ready effects, and —
+  through a store-owned paused time context — time-dependent command
+  effects, synchronously and with no wall-clock waiting (RFC 0008;
+  RFC 0009)
   - `send` applies one message through `update`; `receive`, `receive_matching`,
     and `receive_quit` assert the next deliverable effect output — under a
     canonical per-leaf delivery order that is the store's own contract, not a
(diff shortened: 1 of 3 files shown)


akiomik:
## INV-C4 load-path regression check — full RFC 0006 harness results (after the `test-util` flip)

`cargo bench --bench runtime_load` (all scenarios), local reference machine:

- **Load scenarios**: all report `run: ok` with normal throughput (e.g. `steady_20k` ≈ 20k/s sustained, `burst_20k_bounded` ≈ 390k/s, bounded depth capped at 1025 as configured); no seq-integrity violations.
- **Quit scenarios**: every unkeyed scenario 200/200 valid trials (keyed control 20/20), 0 predicate misses, 0 timeouts, 0 missing deliveries. Representative: `quit_blocked_1` quit→delivered p99 = 0.904 ms; `quit_overload_bounded` p99 = 0.914 ms; `quit_idle_bounded` p99 = 0.721 ms — comfortably inside RFC 0006's criteria.
- **`keyed_isolation`**: `status: ok`, per-key raw yields all `capacity + 1 = 17`, keyed delivered to update = 0, concurrent shared depth = 1025 (= `app_channel_capacity + 1`), `isolation: true`; exit 0.

The RFC 0007 §6 smoke profile also passes standalone (exit 0). This satisfies RFC 0009 INV-C4's requirement that RFC 0006's acceptance apparatus continue to pass after the feature flip, under its own measurement conditions and gating.

PR #270, opened 2026-08-05
docs(rfcs): runtime consolidation bundle (RFC 0010-0012, 0006/0007 amendments)

akiomik:
## Summary

The runtime-consolidation bundle, accepted together per RFC 0010 §1.8:

- **RFC 0010 — Runtime consolidation** (new, Accepted 2026-08-05): the decision record — root verdicts (§3–§7), the full 531-row reaffirmation ledger (§8: reaffirmed 522 / redesigned 9 / delegated 0 / open counterexamples 0), the 115-row additivity audit (§9: A 35 / C 40 / X 13 / adopted(X) 15; interactions pass 7 / excluded 5 / fail 0), and the delegation register with the anti-catalog adoptions (§10).
- **RFC 0011 — Runtime lifecycle** (new): the lifecycle phase machine and termination contract.
- **RFC 0012 — Subscription execution** (new): the source-execution contract, including the quiescence barrier and re-admission semantics.
- **RFC 0006 / RFC 0007 amendments** (semantics-bearing): gauge-schema `runtime_id`/`seq`, INV-L8 rescope, §8 preserved/relaxable partition; `Copy`-derive removal.
- Semantics-neutral retensing/sync across RFCs 0001–0005/0008/0009, and code conformance for documented contracts (fail-before-reuse identifier allocators, `RuntimeConfig` guidance rustdoc).

## Verification

- Acceptance gate (RFC 0010 §1.8): every audit row terminal, no unprocessed breaking verdicts, every remaining conditional delegated externally with owner and both delegation flags; machine-verified end to end.
- Both drift gates closed (RFC 0010 §1.10): code-side endpoint upstream `5c05011`, docs-side endpoint `a6ce2d1`.
- Local CI: fmt, clippy (`-D warnings`), 457 tests passing, doc build, typos, whitespace check.

## Merge instructions

**Please merge with a merge commit — do not squash or rebase.** RFC 0010 §1.10 cites commits `ab313b2` and `a6ce2d1` on this branch as drift-gate anchors; rewriting history would orphan those references and require re-pinning the endpoints and re-running the drift gates.


akiomik:
Review response — both findings addressed in 19669f1.

**[P1] LG-0007-012 reclassified.** The row is now `redesigned` / `changed_to=0007 §2.1`, with a note in the ledger's established style: adoption of the leaf-I41 verdict removes the `Copy` derive at 0.11.0, so the recorded compatibility acceptance is dissolved rather than preserved (redesign under the 1 ID = 1 verdict rule). Both tally sites (§2.4 and the §8 preamble) now read **reaffirmed 522 / redesigned 9 / delegated 0 / open counterexamples 0**; the population (531) and the 478 owner / 53 pointer split are unchanged, and §2.4's destination enumeration already covers the row ("the `Copy`-derive removal, RFC 0007").

**[P2] RFC 0007 status references retensed.** All four present-tense "Draft" references in RFC 0006 (Decision scope, §3.2, §4.1, §6) now state the accepted-with-open-deliverable status, matching 0007's header. The same sweep also corrected the direct status claims the acceptance commit missed inside 0010 itself: §7.5 ("the RFC 0007 Draft"), §7.8 ("the RFC 0006 Draft amendment"), the LG-0007-003 note's status clause, the LG-0006-070 note's tense, and N16's "Draft amendments".

**Deliberately unchanged**, with rationale:
- The **"Draft overlay"** term wherever it cites §1.1's scheme — §1.1's definition was retensed at acceptance ("carried Draft status through the audit … transitioned together at the bundle acceptance"), so the term names the document set, not a current status.
- The **0006 status clauses are retensed, not deleted**: they are dependency statements, not duplicated headers — §3.2/§6 must establish that no prerequisite of the implementation PR remains open, and naming 0007's one remaining deliverable is what makes that negative checkable rather than merely asserted.
- Forward guard: the four 0006 sites plus 0010 §7.5 are the exact text to re-sweep when the `Copy` removal lands and 0007 transitions to Implemented.

Verification: `redesigned` rows grep-counted at 9 (522 reaffirmed, sum 531); no `523` remains anywhere in `docs/`; a case-insensitive `draft` sweep across the five bundle RFCs leaves only the §1.1 defined term, dated process records, and verb usage — zero current-status claims.

akiomik:
Second-round review response — both acceptance-evidence findings addressed. No tracked-file changes were needed; HEAD remains 19669f1.

**[P1] Verifier chain resynced end to end.** The canonical audit series flagged in the review (kept outside the tracked tree) has been updated through every tier: the ledger-replay canon's LG-0007-012 row now records the redesigned verdict with `changed_to=0007 §2.1` and a note in the established style; the §8 skeleton was regenerated with a fresh source hash; the translated series was byte-synced to the RFC rows (LG-0006-070, LG-0007-003, LG-0007-012, and §9's N16) and recorded as a new versioned entry in the tamper-audit changelog; and the generator's and canonical verifier's expected tallies moved from 523/8 to 522/9. The verifier now prints OK against HEAD 19669f1, idempotently under skeleton regeneration.

**Hardening beyond the finding** (from our own adversarial pass): tamper-testing showed the chain previously bound only the §8/§9 table rows and the §10 register — the prose tallies and status prose (exactly where rounds 1–2's defects lived) were unchecked, so "machine-verified" overstated coverage. The verifier gained two new check classes: (1) prose-tally binding — both §2.4/§8 replay-tally sentences (occurrence count fixed, so deletion or rewording also fails), the 531/478/53 population statements, and §9's "Tallies (machine-verified)" line are now bound to counters computed from the actual table rows; (2) a status-prose guard over all five bundle documents rejecting current-status Draft claims, including the copula-elided "Draft for" form that occurred in round 1. Every round-1/2 defect phrasing, when reinstated into a scratch copy, now fails the verifier; the unmodified files pass. The "machine-verified" wording in §9 and the PR body's gate claim are accurate under the extended checker.

**[P2] PR body tally synced.** The Summary now reads reaffirmed 522 / redesigned 9 (single-line edit, verified by diff).

**Deliberately unchanged, disclosed for the next pass:** dated audit-history records keep their original wording — the Japanese canonical notes for LG-0006-070/LG-0007-003, the §9 manifest's dated version-history entry that still phrases N16's condition in pre-acceptance terms, and historical 523/8 occurrences inside dated log entries. These are dated snapshot records rather than consumed tallies; the chain deliberately binds structure (verdicts, grounds, id/span sets) and now prose tallies, not narrative history, and re-anchoring history entries would cascade regeneration for zero evidentiary gain.

akiomik:
Third-round review response — both §F findings fixed, and the verification design was restructured rather than re-patched. No tracked-file changes; HEAD remains 19669f1.

**Process change first.** Since machine-verification gaps reached external review twice, the internal gate was restructured: guard changes are now attacked by an independent bypass-hunter agent that is not shown the implementer's tests, whose mission is to construct inputs the checker accepts while a stated guarantee is violated, property by property. The two rounds of hunting below drove the redesign.

**[P1] F1 per-section binding.** Each prose fact is now bound inside its home section slice (count = 1, values equal the table-derived counters) with a whole-document occurrence guard on top, so corrupting §2.4 while appending a correct tally elsewhere fails twice over. Heading discovery was hardened alongside (unique-heading assertions, chapter-numbering skeleton pin 1..11, any-level heading terminators, fence/comment-aware slicing) after the hunter showed decoy and demoted headings could relocate slices.

**[P1] F2 default-deny.** The phrase blocklist is gone. The status gate now verifies each of the five bundle documents syntactically — exactly one Status declaration across bullet/bold/plain shapes, case-insensitive, canonical `- Status: Accepted`, with the full list-item value checked against a revocation deny-list — and every `draft` token in the normalized text must be fully covered by one of eleven exact pinned phrases (the corpus's real occurrences, each also pinned to exactly one occurrence). "The RFC remains in Draft status.", "is still Draft", "is a Draft RFC", decorated and lowercase variants, and every anchor-abuse construction the hunter produced now fail.

**Structural termination of the arms race.** Two hunt waves (nine bypass classes, then twenty new constructions against the hardened mechanisms) showed that regex hardening alone keeps opening new surface — fenced text invisible to scanning, clause-grafting onto allowed phrases, semantic negation around intact numbers. Since the accepted bundle is frozen, the evidence chain now pins all five RFC files by content hash as the primary tamper barrier: any byte change to the bundle fails verification outright, which retires every textual-tampering class at once. The analytic layer (anchors, tallies, status/draft gates) remains as the drift check that a future legitimate amendment must pass before re-pinning, under the same versioned-changelog discipline the chain already uses. Verified: the pinned hashes equal the committed bytes at 19669f1; all thirty hunter reproductions fail against the pin; twenty-eight of thirty also fail analytically with the pin disabled.

**Machine-verified now means, precisely:** byte identity of the five accepted documents against the pinned evidence, two-tier anchor correspondence of the §8/§9 tables and §10 register, and prose tally/status consistency of the pinned text with the table data. Disclosed residual, for the re-derivation path only: paraphrases that avoid both digits and the draft token (e.g. spelled-out numbers), edits to unbound narrative prose such as chapter titles, and encoding-level attacks — these are covered by the corpus pin in the steady state and by human review of amendment diffs when re-pinning.

PR #275, opened 2026-08-05
test(runtime): add RFC 0011 lifecycle contract tests

### LINKED ISSUE #276: Flaky: lifecycle bootstrap-panic test exceeds its settle bound under host load
## Symptom

`tests/lifecycle.rs::a_bootstrap_subscriptions_panic_propagates_and_reaches_both_postconditions` (added in #275, commit 6788b6b) failed once on CI — Test (macos-latest, stable): https://github.com/akiomik/tears/actions/runs/31019183037/job/92351144432 — with:

```
thread 'a_bootstrap_subscriptions_panic_propagates_and_reaches_both_postconditions' panicked at tests/lifecycle.rs:146:9:
the executor must dismantle the keyed init effect once the requested cancellations are processed
test result: FAILED. 11 passed; 1 failed; 2 ignored
```

The three sibling matrix jobs (macos beta, windows stable/beta) were cancelled by fail-fast, not failed; both ubuntu jobs passed.

## Flakiness evidence (local, macOS, same commit)

| Condition | Runs | Failures |
| --- | --- | --- |
| single test, isolated (`--exact --test-threads 1`) | 300 | 0 |
| full `lifecycle` binary, sequential | 100 | 0 |
| full binary, 4-way parallel (CPU contention) | 200 | **1** (same shape) |
| full binary, 4-way parallel, second session | 240 | 0 |

So the failure needs host-level CPU contention and hits well under 1% of runs — consistent with the single CI occurrence on a shared runner.

## What the assertion means

The test panics `Application::subscriptions` during bootstrap, expects the unwind to tear the runtime down, and then drives `assert_two_stage_postconditions` (tests/lifecycle.rs): a settle loop bounded at 1000 `yield_now` iterations that re-checks the no-further-transitions counters each pass and exits once the producer gauges are zero **and** every registered task-drop witness has flipped. The drop-flag witness exists because the keyed gauge is count-based and publishes zero at `KeyedCommands::Drop`, before the executor dismantles the task futures (RFC 0011 INV-LC7's quiescent postcondition, witnessed via the future's drop). The failure is the loop exhausting its bound with the keyed init effect's future still not dropped.

## Open questions / suspected direction

The runtime under test is `current_thread` + `start_paused`, which should make scheduling deterministic — the sensitivity to *host* load is the puzzle. Plausible mechanisms worth checking: the iteration-bounded loop racing tokio's internal dismantling of an abort-requested task (the abort is a request; the future drops on a later executor poll), interaction with the paused clock's auto-advance, or the task being dismantled on a path `yield_now` does not drive. Not root-caused yet.

Fix directions (do NOT weaken the assertion — no `#[ignore]`, no retry, no loosened postcondition):

- Replace the fixed 1000-iteration bound with a virtual-time bound (e.g. poll the witnesses under `tokio::time::timeout` on the paused clock), keeping the per-iteration no-further-transitions re-check.
- Or add a stronger synchronization point for "runtime-owned tasks fully dismantled" and await it — this is exactly the aggregated outstanding-task observable already noted as the pending-work-aggregation follow-up input in #275's description; fixing this flake and that follow-up may be the same change.

## Environment

- CI: macos-latest, stable toolchain, run 31019183037
- Local repro: macOS (Darwin 25.5.0), stable per rust-toolchain.toml

akiomik:
## Summary

Drafts the contract-test layer for RFC 0011 (runtime lifecycle) ahead of its implementation work: a coverage map of every §8 invariant against the existing suites, 20 new tests closing the gaps the map found, and an explicit list of the invariant halves that need a new observation hook before they can be tested at all. Tests only — no production source, no `docs/rfcs/`, no CHANGELOG (matching the repo's test-only-commit convention).

## What lands

**White-box phase-order tests** (`src/runtime.rs` `mod tests`): a recording application whose `view`/`subscriptions` log the state they observe, driven through `process_input_batch`/`process_frame_tick` directly, asserting INV-LC1 (input batches neither render nor re-evaluate; a frame pass does each once), INV-LC2 (render before re-evaluation, both observing the same latest state, suppression scoped to render only), and INV-LC4's eligibility half (first frame pass renders with no message processed, ignoring the init command's redraw directive).

**Integration lifecycle tests** (`tests/lifecycle.rs`, all `current_thread` + `start_paused`, no sleeps): the ten INV-LC5/LC6 termination rows — unkeyed quit, keyed quit, render error (via a `FailingBackend`, since `TestBackend`'s error is `Infallible`), run-future drop, and panics in `update` / `view` / bootstrap `subscriptions` / steady `subscriptions` / a source constructor — each asserting both INV-LC7 postcondition stages through a bounded settle loop that re-checks the no-further-transitions counters on every yield and uses task-drop-flag witnesses where the count-based keyed gauge publishes zero ahead of the executor dismantling the futures. Plus the three INV-LC8 containment rows (unkeyed, keyed, forwarder panic → application keeps running with surviving producers).

**Two post-conformance tests, `#[ignore = "RFC 0011 §3.4 conformance lands with the 0.11.0 lifecycle change"]`**: inert construction and never-run drop. Both verified to fail for exactly the intended reason when run with `--ignored` on current `main` — they encode the §3.4 target contract, not current behavior.

## Deliberately not covered (needs hooks or structural review — input for the pending-work-aggregation follow-up)

- INV-LC9 (driver exclusivity): structural by the RFC; review, not observation.
- INV-LC4 ordering half: no stable observable phase exists between init dispatch and the effect's first poll (§3.3 records this as negative space).
- INV-LC6 synchrony half: distinguishing "cancellation requested within the drop" from "requested one poll later" needs a synchronously readable outstanding-task observable.
- The keyed-gauge quiescence blind spot above is the concrete motivation for an aggregated pending-work observable.

## Verification

- `cargo fmt --all -- --check` clean; `cargo clippy --all-targets --all-features -- -D warnings` clean
- `cargo test --lib --bins --tests --examples --all-features`: 436 lib; `tests/lifecycle.rs` 12 passed / 2 ignored; 0 failures anywhere
- Mutation checks (applied temporarily, reverted): swapping render/re-evaluation order, breaking `PendingWork`'s initial redraw, and removing the subscription-manager drop abort each fail exactly the intended tests; removing `run()`'s shutdown call fails nothing — the tests pin the RFC's postconditions, not the mechanism.

---

**Update**: the macOS CI failures were root-caused to an un-armable drop witness (constructed in the future's body, so a cancelled-before-first-poll init task dropped the un-started future without ever arming it) and fixed in `727f4d4` — witnesses are now armed on the future value, and the settle bound is a documented scheduler-pass count. Full analysis and corrected reproduction data in the linked issue.

Fixes #276

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -549,8 +549,8 @@ mod tests {
     use std::future::pending;
     use std::hash::{Hash, Hasher};
     use std::num::NonZeroU32;
-    use std::sync::Arc;
     use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+    use std::sync::{Arc, Mutex};
 
     use futures::stream::{self, BoxStream, iter};
     use ratatui::backend::TestBackend;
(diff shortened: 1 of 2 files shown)


akiomik:
CI note: the Test (macos-latest, stable) failure is a flaky settle-bound race in the new bootstrap-panic test, reproduced locally at <1% only under host CPU contention — tracked in #276 with evidence and fix directions (the other three red matrix jobs are fail-fast cancellations). Re-running the failed jobs.

PR #282, opened 2026-08-16
docs(rfcs): adopt the reducer-first core and scope teardown bundle

akiomik:
## Summary

This bundle lands the outcome of the final core architecture selection for the runtime: the reducer-first kernel is the composition core, recorded as contracts across the RFC corpus.

- **RFC 0014 — reducer-first core** (new, Accepted at the spike tier its §13.1 defines): a `Program`/`Reducer` protocol over a runtime kernel — one origin-tagged data lane with revocation filtering at the delivery decision point, a control lane for producer-originated quits, a normative fixed pass pipeline (exit reflection → control drain → bounded input batch → frame), a park/wake contract (INV-RC16) with lane topology pinned (INV-RC15), and a twelve-row supersession register. `Application` is preserved as the single-feature facade: `Runtime<App>::run` and its result contract are unchanged.
- **RFC 0013 — scope teardown** (new, Accepted): `Command::teardown` semantics — prefix selection over every run kind, revocation at the application point, subscription participation by immediate stop, cleanup participation, INV-ST1–ST8 and R1–R9 — co-designed with RFC 0014 §4.
- **RFC 0010**: §1.9 counterexample returns #3–#8 recording the selection outcome on the affected roots (append-only; verdict bodies, snapshots and ledgers unamended), and the §10.1 delegation register discharged for both new RFCs.
- **Owner amendments** across RFC 0003, 0005, 0006, 0007, 0008, 0011 and 0012: the per-invariant successor correspondence for every reached clause, forward-gated on RFC 0014 §13.1's open tier — with RFC 0012 carrying the one immediately-normative additive amendment (the teardown stop cause and the barrier's subject scoping).
- A test-only serialization guard for the `cell_core` loom models (independent fix).

## Acceptance state

RFC 0014 is accepted at the **spike tier**: a twelve-series conformance suite (nine series pass-unit driven, three on the park-boundary probe) is green and repeat-stable on the `spike/b-kernel` prototype, with a mutation battery confirming each park witness detects the arming defect it exists for. The prototype is test-only and is **not** mainlined by this PR.

**Implementation mainlining stays gated** on RFC 0014 §13.1's implementation-acceptance tier (production-loop measurement, the bounded-lane extension of §13.3, the INV-L4 re-derivation of §13.5). This PR accepts contracts, not an implementation.

## Process

The selection ran as a two-candidate gate (Application-centric vs reducer-first) with blind fresh-context drafting, a common driving contract proven by spike, maintainer-pinned counting rubrics and a frozen verdict; reducer-first won on the first lexicographic metric. The assembled bundle then passed three independent adversarial review passes; every blocking finding was resolved and re-verified, and the remaining editorial notes are recorded as deliberately unapplied.


akiomik:
Round-1 review response — all six findings are addressed on the branch:

| Finding | Disposition | Commit |
|---|---|---|
| 1. dangling `G-6` | Dropped, leaving the citation on RFC 0013 R5. Restoration was attempted first: RFC 0010's G-series identifiers are `leaf-G35`/`leaf-G36`/`G-34` (none an admission-coupling demand), the nearest clause (C-15 (f)) demands non-interference rather than accepting the coupling — substituting it would misattribute the claim — and `G-6` turned out to originate in an untracked working artifact, a citation class this repo forbids. | `6089703` |
| 2. `docs/testing.md` vs the loom guard | A third category added: tests that deliberately raise panics take the guard even though they never touch the hook — the rule is about who raises panics, not only who swaps hooks. The concrete hazard (the unfiltered counting hook in `panic_hook.rs`) is named in the section, and the `cell_core.rs` comment now matches. | `c9f6204` |
| 3. `app_channel_capacity` | Renamed at the kernel landing to **`data_lane_capacity`**, with **no compatibility alias**: the control's meaning widens from sizing the shared application-message channel to sizing the one data lane every producer shares, so keeping the old name would assert a compatibility that does not hold. Owned by RFC 0014 §9 row 2 (row 9 stays observability-only); RFC 0007 §7.1 re-derives INV-C2/C3/C4/C6 under the new name, INV-C4's review now checks the old name's absence. `Default` stays explicitly undecided in §7.1, to be settled as a deliberate additive decision in the implementation PR. | `87361d1` |
| 4. return numbering | Unified to per-root numbering: the fixture-correction counterexample is returns #8 (§3.2), #9 (§5.3) and #10 (§6.2), cross-referencing each other in the same form return #4 uses toward #3. §1.9's rule text needed no change. | `dec5276` |
| 5. bounded-mode input admission | Recorded as an explicit loss in RFC 0014 §3.2 (*input admission is not prioritized over producer output — not preserved, newly consequential in bounded mode*) and mirrored in RFC 0006 §5.2, each citing §3.3's no-bound statement as the consequence of the shared lane. | `6089703` / `dec5276` |
| 6a/6b. comment bound syntax, 86-char line | The §2.5 comment now describes the impl in prose (no type-equality in bound position); the long line is rewrapped — both bundle RFCs now have no prose line over 80 columns. | `6089703` |

**6c (the `hook_guard()` duplication)** is deliberately deferred to a follow-up PR: this PR is an RFC bundle plus a single test-only guard fix, and folding a test-support consolidation across seven files into it would dilute what reviewers here are being asked to accept. The follow-up will add `test_support::hook_guard()` and migrate the seven call sites.

akiomik:
Round-2 response — all four findings addressed:

**A** (`e7b0e48`): The rule and the guard stay; the rationale now stands on measured ground. The section states the measurement first (a succeeding loom model never reaches the hook on loom 0.7.2 / generator 0.8.9 — verified against this crate's `Cargo.lock` — and only a failing assertion reaches it, once), then the three reasons the guard is kept anyway (a failing model's panic does inflate the unfiltered counter — one red test turning a second test red is a bad diagnostic; the measured behaviour belongs to dependency versions nothing here pins; the cost is four models serializing). The original flake is recorded as it was observed — the `silent_scope_restores_…` test counting three hook calls where one was expected, attributed by thread-name instrumentation to panics from other tests' threads, twelve consecutive green after serialization — closing with "treat this as an observed flake with an unconfirmed mechanism, not as a mechanism to reason from." The "every explored execution" assertion is gone from both the doc and the `cell_core.rs` comment.

**B** (`ef6cb59`): Row 9 now records why the rename principle deliberately does not apply to the gauge schema: a renamed config field fails the build that uses it, a renamed telemetry field breaks dashboards and parsers silently, off the compiler's path — so the observability face priorities schema stability, re-reads its names, and amends only meanings and the `channel` value domain.

**C** (`ef6cb59`): The CHANGELOG's isolation entry now carries the consequence inline — "terminal input included, so in bounded mode an input's admission can queue behind effect output" — inside the existing entry, not as a new loss.

**D** (`ef6cb59`): RFC 0006 §5.2's absent-parts enumeration grew a third part: §6's resolved questions are records of decisions already taken, and where question 3 restates the buffer-total formula and the isolation it pinned, it follows R1's, INV-L1's and INV-L9's correspondence rows and carries none of its own. §6 itself stays untouched, matching how §3's verdict is handled.

Verification: analysis layer green at each stage, re-pin #11 (0006 only; the other four pinned documents re-measured identical), normal-mode verifier green with the failure path exercised.

akiomik:
Round-3 response: the optional wording item is applied (`d2bdeed`) — the §5.2 sentence now reads "…question 3's buffer-total formula for R1 and INV-L1, and the per-command isolation it pins as INV-L9, **among them** — …", fixing the exemplary reading while keeping the general rule that reaches every restatement (question 2's batching window and frame branch, and questions 6/7's INV-14 / INV-L10 / INV-L11, already named via the §4.6/§4.7 bullets). "Among them" was chosen over "for instance" to avoid planting a hedge-scan token next to a normative clause. Re-pin and both verifier modes green.

akiomik:
Second-reviewer round-1 response — all three findings addressed, taking the minimal option in each case (each is a recorded decision, open to override in review):

**P1** (`47b0838`): Every combinator boundary now takes a **read/write projection pair** — `state`/`state_mut`, `rows`/`rows_mut`, `slot`/`slot_mut`, on the `get`/`get_mut` convention. §2.5 derives why a pair is required (the two `Reducer` methods borrow the parent differently: `reduce` needs the mutable projection, `subscriptions` the shared one — a combinator holding only the mutable accessor would have to fabricate an aliasing borrow or drop the child's declarations), reconciles it with the INV-SE6 purity transfer, and INV-RC2's aggregation clause now states the declarations are read through the shared projection. No lens trait is introduced.

**P2 hook guard** (`ee02722`): Fixed at the recorder, which is the side that scales: the hook-restoration counter now filters by thread name — the same filter `HookProbe` already applies, reusing the module's `on_thread` helper — so no unrelated panic can inflate it, whether from `#[should_panic]` tests or anything else. `docs/testing.md` is reorganized around that as the primary defence ("there is no way to enumerate — let alone guard — every test that panics on purpose"), with the loom-model guard kept as a defensive second layer and round-2's measured facts intact. `cargo test --lib` 400 passed, `--all-targets --all-features` 473 passed, zero failures.

**P2 `CleanupEffect`** (`47b0838`): The phantom type is gone — `Command::on_teardown(finalizer: impl Future<Output = ()> + Send + 'static)` takes the future directly. The bounds are copied from the code's own `Command::future` (`src/command/core.rs`) rather than invented, differing only in `Output = ()`, which the text notes closes the message-return path in the signature itself while the structural review keeps covering capability handing. Zero occurrences of `CleanupEffect` remain in the corpus.

akiomik:
Second-reviewer round-2 response (`7fc4ba6`) — the finding is resolved, though the investigation ended somewhere more interesting than either prior account:

**Your source citation is correct, and so is the earlier zero-reach measurement — `gen_impl.rs` is the missing piece.** `done()` does raise `panic_any(Error::Done)` (`yield_.rs`), and loom's scheduler does end coroutine bodies with it. But generator itself manages the process-global hook around its own unwinds: tearing a coroutine down it calls `take_hook`, installs a no-op, and reinstalls the previous hook afterwards (`gen_impl.rs`). Re-measured on this crate's locked versions (both with and without `--cfg loom`): a succeeding `cell_core` model explored 27 interleavings and reached an installed, **unfiltered** hook zero times — with the counter's liveness proven in the same run by a deliberate panic it did count.

**The guard's justification is now stated on that verified ground, and it is stronger than the "future versions" framing:** a loom model *is a hook swapper* — the take/set/restore sequence a dependency performs is still a hook swap happening concurrently with any hook-holding test — so the loom rule is an instance of the section's first rule rather than a category resting on a frequency claim. The historical diagnosis (which named generator's completion path as the primary cause, with thread-name instrumentation) is recorded as it happened, beside the re-measurement that does not reproduce it; the two observations are left as they are rather than reconciled by argument. The "unconfirmed mechanism" understatement is gone.

`cargo test --features loom-core --lib` 401 passed / 0 failed; both verifier modes green; no pinned document touched.

PR #283, opened 2026-08-16
refactor(test-support): consolidate the panic-hook guard helper

akiomik:
## Summary

The follow-up deferred from #282's review (first reviewer, finding 6c): the `PANIC_HOOK_GUARD.lock().unwrap_or_else(PoisonError::into_inner)` idiom, duplicated across the crate's hook tests, is consolidated into one helper.

- `test_support::hook_guard() -> MutexGuard<'static, ()>` now owns the lock and the poison recovery, with the reasoning documented once: poisoning is the ordinary case for this guard — the tests it serializes panic on purpose — so recovering keeps one test's panic from failing every later hook test.
- Eleven call sites migrate to it (`panic.rs` ×5, `subscription.rs`, `runtime/core.rs`, `runtime/keyed_commands.rs`, `test_support/panic_hook.rs` ×2, and `cell_core.rs`, whose private copy is deleted). The guard static is no longer re-exported; the helper is the single way to take it.
- `tests/common/panic_hook.rs` is deliberately untouched: the `#[cfg(test)]` test-support module does not exist in an integration binary's view of the crate, and the documented duplication policy already records the deeper difference — the integration copy uses `tokio::sync::Mutex` (held across `.await`), which has no poisoning, so the consolidated idiom never occurred there.
- Three remaining `unwrap_or_else(PoisonError::into_inner)` sites (`runtime/load.rs`, `command/effect.rs` ×2) guard unrelated mutexes and are left alone.
- `docs/testing.md` points at the helper; the rule's substance is unchanged.

## Verification

`cargo test --lib` 400 passed, `--features loom-core --lib` 401 passed, `--all-targets --all-features` 473 passed — zero failures; `clippy --all-targets --all-features -D warnings` clean; `fmt`/`typos`/`diff --check` clean. `docs/rfcs/` is untouched.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -155,10 +155,14 @@ across the two copies.
 
 ## Process-Global Panic Hook Tests
 
-Use `crate::test_support::PANIC_HOOK_GUARD` for crate-internal tests that directly
+Use `crate::test_support::hook_guard()` for crate-internal tests that directly
 call `std::panic::set_hook` or `std::panic::take_hook`. The panic hook is
 process-global and tests run in parallel by default, so hook-swapping tests can
-otherwise clobber each other or observe an unrelated panic.
+otherwise clobber each other or observe an unrelated panic. The helper locks the
+process-global `PANIC_HOOK_GUARD` and recovers from poisoning in one place —
+poisoning is ordinary here, since the tests it serializes panic on purpose, and
+recovering keeps one such panic from failing every later hook test with a poison
+error instead of its own assertion.
 
 Hold the guard for the full critical section: install or take the hook, trigger
 and catch the panic if needed, restore the previous hook, and then inspect any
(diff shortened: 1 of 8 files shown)


akiomik:
Review response — all four items applied in `fce020e`, plus one transparency note about the base commit.

1. **Rustdoc vocabulary**: the distinction you drew is applied exactly — the one caller-facing instruction (`HookProbe`'s "callers hold [`PANIC_HOOK_GUARD`]…") now says "callers call [`hook_guard`] and hold it…", while the three internal descriptions (`with_silent_panic_hook`'s across-`.await` note, the static's own doc, and `hook_guard`'s "Locks […]") keep the static's name, since the static is what they are talking about.
2. **`cell_core.rs` comment**: moved directly under `mod tests {`, ahead of the imports with a blank line — no longer readable as a comment on `use loom::sync::Arc`, and outside rustfmt's import-sorting path.
3. **`docs/testing.md`**: the three affected paragraphs are reflowed, no wording changed.
4. **`PANIC_HOOK_GUARD` is now module-private** (taking the optional item): with the re-export already gone, dropping `pub` makes "the helper is the only way in" compiler-enforced rather than conventional. `cargo doc --no-deps` is clean; the two unresolved links `--document-private-items -D warnings` surfaces are pre-existing and unrelated (`runtime/channel.rs:31`, `runtime.rs:418` — identical on the base tree).

**Transparency note**: the base commit was amended (`c9acfdb` → `4429d12`, force-pushed with lease) to correct a figure in its message — it claimed "473 passed" for `--all-targets --all-features`, a number carried from the pre-consolidation branch; the measured value on this branch is **485 passed / 0 failed** (17 result lines), and the follow-up commit was held until its own run finished rather than reusing a number. No code differs from what you reviewed; only the message text changed.

All checks measured on the final tree: `--lib` 400, `loom-core --lib` 401, `--all-targets --all-features` 485 — zero failures; clippy `-D warnings` 0; fmt/typos/diff-check clean.

PR #284, opened 2026-08-16
docs(rfcs): fold the bundle review lessons into the checklist

akiomik:
## Summary

The reducer-first bundle's review cycle — three adversarial pre-review passes and five human review rounds on #282, plus #283's round — surfaced defect classes the pre-review checklist should have caught. Per the checklist's own meta-rule (fold restatements into existing items; add only genuinely new classes; prune what no longer earns its length), this lands as **two additions, seven folds, and one prune**:

**Added (new defect classes):**
- *Item 1 — a cited invariant's own clauses are an inventory*: a successor mapping that promised an answer for every clause walked all sixteen invariants but answered only half of the one that pinned two properties.
- *Item 1 — a rationale that generalizes is quantified too*: a principle used to justify one decision reaches every same-shaped surface; apply it there or record where and why it divides (the config-rename vs telemetry-schema split needed its reason written down).

**Folded (sharpenings of principles already present):**
- Item 2's mandated-test bullet: the joint-satisfiability walk includes the document's own evidence rules — an invariant whose only admissible instrument those rules exclude has no witness.
- Item 4's dependency bullet: one correctly read source site is not yet the mechanism — a neighbouring routine can invert its visible effect (generator's `done()` raises `panic_any`; its teardown wrapper silences the hook around it).
- Item 4's measurement block, three sub-bullets: an instrument proves its own liveness in the same run (an unvalidated counter's zero measures nothing); two verified observations that disagree are juxtaposed, not reconciled by argument, and the normative rule grounds on what needs neither; a number in a commit message or PR body is a measurement citation of the exact tree it describes.
- Item 4's opener: an outside measurement is the same class as a reviewer's description — a hypothesis to verify, never an authority over the repo's own record, which is itself re-verified before being asserted as current.
- Item 6: the fix's own rationale scopes the changed-claims sweep — a commit that names its defect class puts every corpus instance of that class on the list (the class fixed in one document regenerated as the next round's P1 from the siblings, and once more from a fourth site in the same document).
- Item 7's citation bullet: every cited identifier token resolves inside the tracked corpus — an identifier inherited from untracked working notes is a finding even when the sentence reads plausibly (`G-6` survived six rounds).
- Item 7's audience bullet: a visibility change makes caller-facing instructions corrected claims while internal descriptions correctly keep the internal name — sweep by instruction-vs-description.

**Pruned:** item 4's PR #211 example loses its second half; the first half teaches the principle completely and eight later examples now carry the same discipline.

### DIFF
--- a/docs/rfcs/pre-review-checklist.md
+++ b/docs/rfcs/pre-review-checklist.md
@@ -74,6 +74,16 @@ externally completed), and a post-quit no-poll exception named only
 the finish/drop checks while send/receive were also polling sites
 that could reach a leaf on their way to failing.
 
+A cited invariant's own clauses are an inventory. A correspondence or
+successor claim that quantifies over invariants — *each becomes*, *all
+hold*, *every clause is preserved* — is checked clause by clause inside
+each one, because an invariant that pins two properties needs two
+answers. *From PR #282:* a successor mapping opened "every clause above
+holds as stated; each becomes the following", walked all sixteen
+invariants, and answered only the arming half of the one that also
+pinned a receiver-exhaustion result — the second half had no successor
+anywhere, in a section whose lead promised one for each.
+
 A universal claim about executions quantifies over the behavior other
 parties supply. A harness, runtime, or wrapper cannot promise a
 property of the whole execution when part of the execution is the
@@ -101,6 +111,18 @@ presented as enforcing "every time read" missed `UNIX_EPOCH.elapsed()`
 the fix text then claimed the list was "derived by walking `std`'s
 surface" while `std::net`'s nameable timed I/O was still absent from
 it.
+
+A rationale that generalizes is quantified too. When a decision is
+justified by a principle rather than by its own particulars — *a
+silently accepted old name would misstate what it bounds* — that
+principle reaches every surface of the same shape, so either apply it
+to all of them or record where it divides and why. An unrecorded
+division reads as an arbitrary decision on re-reading. *From PR #282:*
+a renamed configuration control and a re-read telemetry schema were
+decided oppositely on the same page; the reason is real — a renamed
+config field fails the build that uses it, a renamed telemetry field
+breaks dashboards silently, off the compiler's path — but it was
+recorded only after a reviewer asked why the principle stopped.
 
 A universal negative over a set that contains the rule's own
 instrument is checked against that member first. A contract that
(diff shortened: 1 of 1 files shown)


akiomik:
Review response (`ce55e84`) — all five items addressed:

**P1 (the "because")**: Your reading is confirmed at the source, and it goes one step further: `raw_cancel`'s silencing window wraps only the resume it performs, its callers (`cancel`, `Drop`) guard on the coroutine *not* being done — so a coroutine that completed takes the early return and the window is unreachable on precisely the path a succeeding model ends through. The paragraph no longer argues a refutation: it now presents the two sites as "two sites, two paths, hook handling that differs between them", states that neither reading establishes what an installed hook observes, and credits the in-repo measurement with settling it — consistent with the disagreeing-observations bullet and with `docs/testing.md`'s deliberately unreconciled wording (untouched, as is the `cell_core.rs` comment; both were already scoped to teardown).

**P1 (the ranking)**: "The order is:" is gone. The opener now states a verification order, not a truth authority — measure here first; where impossible, consult the repository's own record; treat an outside report as standing until one of those bears on it — and explicitly defers to the measurement-citation bullets for what to write when verified observations disagree, so it can no longer instruct anyone to discard a record.

**P2**: The second full narration of the hook incident is folded away; the opener keeps one principle sentence and points at the bullet that carries the rule and the concrete example.

**P3 (wrap)**: The orphaned line is rewrapped, plus one 84-column line the same insertion had left. The file's four 76–77-column lines are pre-existing at the base commit; this PR introduces no line over the convention.

**P3 (untracked citation)**: The item-2 example now cites the tracked record — RFC 0014 §7.2 (the park-boundary probe and its citation scope) and §13.1 (the per-instrument suite split) — both verified present on main, restoring consistency with the identifier-resolution rule this PR itself adds.

akiomik:
Re-review response (`91e9a14`) — all items addressed:

**P2**: The sweep rule is bounded and given its relief. The scope is defined in place — "the tracked documents its subject reaches — the RFCs plus the repository documents they cite, which is where a contract claim's siblings live" — covering what the real incident spanned without stretching to rustdoc or the whole tree; and the escape hatch is now symmetric with item 1: "Sweep that scope, or record in the changed-claims list which boundary was swept and why the remainder is out of it; an unrecorded boundary is the same defect as an unrecorded division under item 1." One deliberate choice to flag: item 7's "tracked corpus" was left as is — there the word contrasts tracked against whatever notes were open while drafting, which defines it in place — say the word if you'd rather both sites share one wording.

**P3 "standing"**: now "hold any outside report as an open hypothesis until one of those two bears on it".

**P3 (three tellings)**: the prune is applied now, in the minimal form you sketched — the dependency paragraph keeps the class it teaches (two sites, two paths, neither settling the effect) and closes with "What settled it, and how the result was written, are the measurement-citation bullets below." One telling plus two pointers; no class loses its example.

**Figures**: your inability to reconstruct 8/7 was the finding — the numbers were wrong. Re-derived from the six tense-sync commits on main (`75e1246`, `ce3c744`, `9f74691`, `3ee74d4`, `da8b029`, `e5aec19`): **nine sites across six documents**, of which the next pass's enumeration found eight (RFC 0005's fourth surfaced during the fix — a sentence the example now carries). Every figure in the passage is reproducible with `git show`.

**Nit**: both rewrap seams are closed, including one my own edit had introduced; the only remaining >75-column lines are the four pre-existing at the base commit.

On the unverifiable non-blocker: correct — the "473" figure was amended away in the then-unpushed `c9acfdb`, so no pushed history carries it; the public record of the slip and its correction is the transparency note in PR #283's thread, and the checklist prompt it produced is what makes the incident durable beyond that thread.

PR #285, opened 2026-08-17
docs(rfcs): land the stage-3 driving layer in RFC 0008

akiomik:
## Summary

RFC 0014 §9 row 11 and §13.2 delegate the stage-3 driving layer's API
body to RFC 0008. This PR lands that amendment: a new **§9 "The stage-3
driving layer"** defining the `TestDriver` / `ParkProbe` surface —
same-topology driving of the production kernel with scripted pass
initiation and send grants — plus the store parity-extension delegation
(§9.10). The former §9 (Open questions) and §10 (References) renumber
to §10/§11; RFC 0010's ledger row `LG-0008-047` and the README index
row follow the renumber.

The surface is **contract-only until RFC 0014 §13.1's
implementation-acceptance tier closes**: nothing defined here exists in
the crate today, and the `Added` CHANGELOG entry ships with the kernel
landing (§9.1). Stages 1–2 (§§2–8) are byte-identical to `main`.

## What §9 pins

- **Same topology (INV-RC13, §9.2)**: construction through the
  production path; the five prohibited shapes have no constructor.
- **The API body (§9.3)**: `TestDriver` (`new` / `boot` / `step_pass` /
  `grant` / `confirm` / `settle` / `accepted` / `intents`),
  `WakeSource { Data, Control, ProducerExit }` as the whole scripting
  vocabulary, `StepReport`, `GrantToken`,
  `Confirmed { Accepted, Reclaimed }`, `AnonymousRun`, `ParkProbe`;
  a 3-state × 8-call legality table.
- **Origin naming (§9.4)**: keyed and subscription runs by their
  RFC 0005 identities; anonymous runs by a driver-minted opaque name
  that reaches no kernel identity surface (RFC 0013 §9 resolution 3 /
  §10's auto-keying rejection honored).
- **Pass initiation (§9.5)**: `boot` is the production intake plus the
  continuation pass that consumes the pending first render, so the
  scripted seam is exactly the three wake sources of RFC 0014 §3.5 —
  no fourth reason a pass begins.
- **The grant handshake (§9.6)**: at most one outstanding grant
  driver-wide, so raw grant order is unrepresentable (RFC 0014 §7.2 /
  INV-RC14's structural half); the gate covers producer sends on both
  lanes, quits included; a grant resolves on the released send's end —
  `Accepted` (the send got into the lane) or `Reclaimed` (it never
  will) — revocation being a delivery-side filter that does not
  prevent admission (RFC 0014 §3.1, INV-RC5, §4.3); `accepted()` is
  the acceptance ledger (admission, not delivery; entries carry origin
  and lane), `intents()` the non-guaranteed pre-gate ledger; `settle`
  supplies bounded executor turns and is misuse while a grant is
  outstanding; executor-turn non-selectivity is stated as a class fact
  over the four turning calls.
- **`ParkProbe` (§9.7)**: evidence scope is INV-RC16 alone; its wake
  vocabulary is the same `WakeSource`.
- **Determinism (§9.8)**: scoped to a current-thread executor and
  unbounded lanes; the bounded extension stays open with RFC 0014
  §13.3's two protocol conditions named.
- **Evidence surface and citation rule (§9.9)**; **store parity
  delegation (§9.10)**; **coverage, adversarial models, and excluded
  claims (§9.11)**.

## Decisions to review explicitly

1. **The ledger is named `accepted()` / `AcceptanceLedger`.** It
   records admissions past the gate, which delivery never sees in the
   revoked case, so the name states what the ledger holds rather than
   defending a "deliveries" spelling in prose. This follows RFC 0014
   §9 row 2's precedent (a control whose meaning moved was renamed,
   not explained). The rename was decided during the pre-submission
   review process described below, not first derived by the RFC's own
   argument — flagged so reviewers weigh it on its merits.
2. **§9.6 deliberately does not settle** how RFC 0014 §13.3's trailing
   ack-correlation clause reads where a grant resolves as `Reclaimed`
   (no commit to correlate). That reading is assigned to §13.3's
   resolution, which lands as a future addendum to this same section
   (§9.8 keeps it open).
3. **`Confirmed::Accepted` can follow the run's revocation.** The arms
   partition the released send's ends, not the run's states; INV-RC5
   is checked on pass behavior, not on either ledger (§9.11).

## Open items left to the future RFC 0008 addendum

1. The bounded-lane determinism verification pass and its recorded
   scope (§9.8).
2. How §13.3's trailing ack-correlation clause reads for a `Reclaimed`
   grant (§9.6).
3. Which of §13.3's two ack-correlation disjuncts the verified surface
   rests on (§9.6 supplies both; the resolution records which was
   verified).
4. The driver-progress condition's verification — the shape satisfies
   it, but a shape is not a verification (§9.8).
5. §9.10's named delegation: store intake for teardown entries and
   multi-keyed batch children, §5.1's INV-11 bullet, and the
   shared-first parity sites, all owned by the change that lands the
   kernel-side lowering.

## Process

- Pre-review checklist: all 8 items with per-item verdicts, re-run
  after every fix round.
- Independent adversarial review (mission: find grounds for No-Go):
  five rounds, blockers converging 5 → 2 → 1 → 1 → none; the final
  round's non-blocking residuals were applied before this PR.
- Note for readers of the commit history: some intermediate commit
  messages state positions later rounds corrected (an earlier
  ParkProbe-series driving claim; an earlier premise that revocation
  prevents acceptance; superseded names). This description states the
  final text's contract; the history is preserved as history.


akiomik:
Thanks — all eight addressed in `c5ed8ce`.

**1. `grant(&self)` vs the `&mut self` universal.** Took option A:
`grant` now takes `&mut self` (§9.3). You were right that the receiver
was never what §13.3's driver-progress condition needed — that turns
on the token being `'static`, so the rationale in §9.6 is rewritten
onto the token's own property. The universal is restored across all
five driving methods, and interior mutability for the outstanding slot
is no longer implied. One knock-on the sweep caught: the block's
normativity note listed "parameter forms" as latitude, which would
have made the receiver unenforceable — it now pins receivers as
normative and names what they carry (INV-LC9, and the
unrepresentability of a borrowing token).

**2. Armed grant that never sees an intent.** Agreed, and it is
reachable the way you describe. `Confirmed`'s subject is now the
*grant* rather than the released send: `Reclaimed` means this grant
will never put anything into the lane, established by either the
released send ending without getting in, or the origin being gone with
no send released at all (§9.6). *Unresolvable grant* in §9.11 now
names both cases. Grant legality was not narrowed to "runs that may
still send" — your reason against it (exit timing is not predictable
by the test) holds.

**3. Which intent a grant releases.** Stated in both the rustdoc and
§9.6: the next unreleased intent at that origin, *including* one
already waiting at the gate, otherwise the first to arrive after.
§9.6 adds why it matters — a grant is issuable ahead of the producer,
which is what lets a script fix an order before the run reaches its
send point.

**4. `settle` missing from the script.** Fixed in §9.8 — a script is
now the ordered driving-call sequence (each `step_pass`'s `WakeSource`
plus the positions of `grant`, `confirm`, and `settle`) together with
application-side inputs and readiness. `settle` belongs because its
position changes what a later `ProducerExit` step has to reflect,
even though it contributes no entry to the guaranteed sequence. The
same paragraph notes this instantiates RFC 0014 §7.2's tuple rather
than widening it.

**5. Probe's future provenance.** §9.7: the test calls
`ProgramRuntime::run` itself and pins the returned future. No
`TestDriver` hands one out, which is the same rule §9.2's
manual-effect-polling row turns on.

**6. Render observation.** Recorded as negative space in §9.11: no
render-observation surface — the driver owns its terminal and never
returns it, and `StepReport` carries only `started`/`terminated`.
Render evidence is application-side (a `view` recording its own
calls), which is where INV-RC10's redraw row is read.

**7. Sibling pointers.** Deliberate, and now permanently recorded in
§9.1: RFC 0014 §13.2, RFC 0012 §6.2 and its §12 second question, and
RFC 0010's TS-1/TS-3/TS-6 close at the landing rather than at this
section — a contract stated is not a surface delivered, and those
rows are conditioned on the delivery. Sibling documents are untouched
in this PR. Separately: those rows call this surface a "driving
store" where §9 defines a driver beside the store — noted as a
follow-up for when they are closed.

**8. Wraps.** Reflowed by paragraph; an automated ragged-wrap scan
over §9 is clean.

akiomik:
All five in `4bd594b`.

**A. `Reclaimed`'s second fact vs. `confirm`'s no-pass rule.** Good
catch, and it broke exactly the case the arm was added for. Kept the
kernel-bookkeeping definition as you suggested, and expanded "the test
steps instead" into both step-requiring routes (§9.6): a commit that
needs the lane drained, and a grant whose origin has exited but whose
exit no pass has reflected — the latter resolved by interposing
`step_pass(WakeSource::ProducerExit)` before `confirm`. No contract
change is needed for either, since the token is detached and survives
the step, and §9.9's route enumeration now carries the same
requirement so the two read together.

**B. `GrantToken` doc.** Now "correlated to the send it releases, and
through it to that send's commit where one happens" — a commitless end
being a normal outcome since the last round.

**C. `ProgramRuntime`.** Added to §9.1's third division, noting it is
there because §9.7 names it.

**D. Receiver scope.** Scoped to `TestDriver` at all four sites. You
were right that this is the same failure class as the original
receiver finding — unscoped, the universal reads against
`ParkProbe::poll(&self)`.

**E.** `boot` is in §9.8's script with the note that the state table
fixes its position, so it is not a free variable. §9.11's render line
now quotes INV-RC10's own enforcement wording ("a scripted flood with
an interposed probe and a pending redraw") and identifies the
interposed probe as the application-side instrument. Seven ragged
wraps reflowed — the four you listed plus three the newer edits
introduced; both the width and ragged-wrap scans over §9 are clean.

akiomik:
Both in `12401df` — §9.9's route requirement is now stated per route ("one always … another when its establishing fact is an origin's exit") with no count asserted then qualified. Sweeping that class turned up one more in the same bullet: its closing sentence ranged "neither"/"both" over three calls (`grant`, `confirm`, `settle`), now "none of the three". The four short lines are reflowed at paragraph granularity; width and ragged-wrap scans over §9 are clean.

akiomik:
All three in `8119b6b`.

**1. `Origin` vs. run instances.** You're right, and the supersession
case is the clean counterexample: the identity slot is free for the
successor at the revocation's application point while the revoked run
can still send late (RFC 0014 §3.1 / RFC 0003 INV-8), so
`Origin::Keyed(CommandId)` names both. `AnonymousRun`'s three rules
are generalised to a per-run `RunName` minted for every kind, with
`RunKind` carrying the logical identity where there is one. §9.4 now
grounds it explicitly: this is the test-surface counterpart of the
kernel's per-run token discipline, not a second kernel identity —
RFC 0005's model is untouched, and the anonymous case still adds no
key where the kernel has none. Renamed since it no longer names only
anonymous runs; an *Identity as run* model records the conflation it
excludes.

**2. Ledger read surface.** Added: `SendRecord` (run + lane) and, on
both ledgers, `len`/`is_empty`/`iter` in their stated order. One
record type for both, since the content is identical and only the
worth of the order differs. Noted in §9.11 that this is what makes
INV-RC14's sequence assertable — without it the invariant had a
contract and no instrument.

**3. `settle` completion.** Now takes a caller predicate, evaluated
before the first turn (so an already-true condition costs none) and
at turn boundaries after, bounded as every wait here is. §9.6 gives
the reason for the predicate form: "until the executor is idle" is
not observable through any surface here and implementations disagree
on it, so it would differ between conforming kernels, while a
predicate's evaluation points are fixed at turn boundaries. §9.8
records that the predicate reads application-side state and so adds
no free variable to a script.

akiomik:
Both in `4b438ce`.

**1. Turn budget and unit.** Caller-supplied budget:
`settle(max_turns, until)`, exhaustion failing with the turns
consumed. And a turn is now defined normatively — the driver task
yields once, and every task ready at that yield gets the executor
before the predicate is re-evaluated — with the reason in the text:
two drivers whose turns admit different numbers of ready tasks reach
different states on the same call. It is satisfiable by a FIFO ready
queue on the current-thread executor, so it stays inside the
determinism claim's verified range and claims nothing outside it, and
it does not touch RFC 0011 §2.3 — a turn is a unit of opportunity,
not progress. Added the converse for `confirm`, since the asymmetry
would otherwise be the next question: its completion condition is not
the caller's, so any two bounds agree wherever the resolution arrives
and the bound is only a hang guard.

**2. Predicate as a free variable.** You're right, and the
counterexample is now in the text. A settle's predicate and
`max_turns` are script elements. On the §7.2 tuple the instantiation
claim is dropped rather than stretched: adding `settle` at all
already went past that tuple, and the arguments ride the same
extension — with an explicit statement of what the extension does not
do, so §7.2 isn't weakened. The determinism is derived rather than
asserted: application state evolves deterministically under a fixed
script, the predicate is a function of that state at defined
boundaries, so its evaluation sequence is fixed too.

Sweeping for the deleted claim found two places that still asserted
it — §9.11's coverage row, and §9.3's "the bound's value is
mechanism" — both corrected.

akiomik:
One more, in `64f0510`.

An implementation pass against this contract found
`ParkProbe::armed()` and `woken_by()` unimplementable, and it is worth
being direct about the direction: this was over-specification on the
contract's side that implementability exposed, not a kernel limitation
to work around. The kernel parks on a single select across both lanes
and the exit notifications — one waker registered, and a wake carries
no tag. Answering per-source needs per-source instrumentation inside
the kernel, i.e. a branch, and RFC 0014 §7.2 forbids exactly that of
this instrument in the same sentence that gives it its waker and poll.

Both accessors are removed; `ParkProbe` is `poll` plus a wake count.
What §7.2 says the probe observes is unchanged, because those are
observations rather than accessors — §9.7 now states how each is
established, which is the shape §13.1 already requires of the three
series. That it parked: a re-poll returning `Pending` with neither
ledger moving and the wake count still, then a count that moves
exactly once after the arrival. Which source: the script arranges one
source's arrival and no other, so the wake can have come from nothing
else.

INV-RC16 is unaffected — its arming half was always structural at the
park site (no finite test proves a registration for a source it did
not exercise), and its behavioral half is the three series, one per
armed source, carried as before.

akiomik:
Four, in `e8b7651` — and this one touches RFC 0014, flagged up front.

**1. Turn definition.** You're right that the property isn't
available: no ready-set observation, and a bare yield may re-poll the
yielding task. A turn is now defined by *construction* — the driver
spawns a fresh no-op task onto its own executor and awaits it, spawn
and join, public primitives only. That gives what the property was
for (two conforming drivers build the same turn, so no test passes
under one and fails under the other) without asserting anything about
scheduling; the ready-task behaviour is demoted to an informative
note about the verified range's executor. Spawning one ordinary task
adds no hook and reads no scheduler state, so §7.2's
uninstrumented-scheduling clause is intact.

**2. `confirm`'s budget.** The asymmetry paragraph was wrong and is
gone — your counterexample is decisive, and `confirm(max_turns,
token)` is now symmetric with `settle`. Both budgets are script
elements; only the completion condition differs (`settle`'s is the
caller's predicate, `confirm`'s is the gate's resolution).

**3. RFC 0014.** This is the first edit to an owner RFC in this PR,
and it is one bullet: §7.2's script enumeration now opens in exactly
one direction — the driving vocabulary a stage-3 surface adds and
those calls' arguments, with what is added fixed by RFC 0008 §9 and
every addition still bound by the clauses that follow. INV-RC14's own
text does not restate the tuple, so it needed no change — that is
what kept the edit to a dozen lines. RFC 0008 §1.3's enumeration is
synced. The checklist ran independently over the RFC 0014 diff (items
4 and 7 in particular), and a corpus sweep for the tuple's wording
finds exactly two sites, §7.2 and §1.3, now consistent.

**4. Park witness.** Correct — that execution has no ledgers. The
witness is now a re-poll returning `Pending`, an unmoved wake count,
and the application's own instrumentation staying silent, with a
sentence saying why the ledgers cannot be among the witnesses.

One knock-on the sweep caught: *Selective turning* was still closing
on the demoted property. It now closes by construction, which is
stronger — a turn is a spawn and a join, so there is no per-task
control to be selective with.

akiomik:
Both in `e17efd2`, with a follow-up in `70a043d`.

**1. `confirm`'s zero-turn case.** Undefined as you say, and it split
conformance twice — on whether `confirm(0, token)` can succeed at
all, and on whether the turn a resolved grant did not need moves the
intent ledger or carries a run to an exit. Both calls now check their
completion condition before the first turn, symmetric with what
`settle` already did, so `confirm(0, token)` is the way a test
asserts a grant is already resolved. Sweeping "budget"/"caller" over
the section for the same class found the one stale site you pointed
at — the waiting rule still said only `settle`'s bound was the
caller's; it now names both, and records that `boot` and `step_pass`
wait on no condition at all.

**2. Edit remnant.** Removed, and the section was scanned for other
truncated or repeated sentences — there are none.

`70a043d` is a process miss worth naming: the first commit left that
paragraph at 101 columns, the width check reported it, and the
failure was read past because a success banner printed below it.
Reflowed, and the verification now prints only on failure so there is
no banner to misread.

PR #288, opened 2026-08-17
feat(kernel): land the reducer-first kernel as private scaffolding

akiomik:
## Summary

This PR lands the reducer-first kernel of RFC 0014 as **private
scaffolding**: the kernel, its delivery accounting, the stage-3
`TestDriver`/`ParkProbe` of RFC 0008 §9, and the conformance suite
RFC 0014 §13.1 names — all `pub(crate)` or `#[cfg(test)]`. Nothing
changes for users of this crate today.

- **Production behavior and the public API are unchanged.**
  `cargo public-api --all-features` between `main` and this branch is
  identical, and the nightly `tests/api_surface.rs` checks pass. The
  old runtime remains the production owner of every path.
- Production entry stays where it is: switching `Runtime` to this
  kernel, publishing the new API, and removing the old topology are
  later changes, gated on RFC 0014 §13.1's implementation-acceptance
  tier as the RFC requires.

## What lands

1. **Command/identity groundwork** (production files, behavior
   preserved): effect leaves carry key/scope/kind metadata that the
   old runtime's consumption path structurally never sees
   (`into_execution_parts` returns the same 3-tuple with the same
   content); `Command::batch` pushes child keys down into leaves while
   keeping its warning and its `key: None` result;
   `ScopePath` becomes root-first (a bijection on paths — the equality
   relation, and therefore `CommandId`/`SubscriptionId` identity, is
   unchanged, and no `Debug` output prints segment order), so
   RFC 0013 §3.1's prefix selection reads literally as
   `starts_with`; a `pub(crate)` teardown constructor.
2. **The kernel** (`src/kernel/`): one FIFO data lane (constructed
   through the observed channel, so capacity-wait events and the
   `blocked` gauge work unchanged), an unbounded control lane, the
   reservation→commit/release delivery accounting with its loom
   mirror, per-run registry with scope tracking and revocation
   (`stop_request` as the single transition), teardown selection over
   every run kind, producer harness on one `JoinSet` with contained
   producer panics, the fixed four-stage pass of RFC 0014 §3.5,
   park/wake over exactly the three wake sources (INV-RC16), bootstrap
   with the synchronous init-quit short-circuit, and the two-stage
   termination postconditions.
3. **The stage-3 driving layer** (`src/testing/driver.rs`,
   `pub(crate)` until the kernel lands publicly): the eight calls of
   RFC 0008 §9.3 with their normative receivers, `RunName`/`RunKind`
   per-run naming, the grant handshake (driver-wide ≤ 1, both
   `Confirmed` resolutions), `settle(max_turns, until)` with the
   constructive turn definition, the acceptance/intent ledgers (inert
   in production construction — no records, no locks — recording only
   under a scripted gate), and `ParkProbe` as `poll` plus a wake
   count.
4. **The conformance suite** (`src/kernel/conformance/`, 42 tests):
   the twelve series of RFC 0014 §13.1 under their own names — nine
   pass-unit driven, three on `ParkProbe` — plus the RFC 0013 kernel
   teardown rows. The series-to-invariant mapping is in
   `src/kernel/conformance.rs`'s module doc, together with an explicit
   list of what this surface does not reach (the
   implementation-acceptance rows that land with later changes, and
   the halves the invariants themselves declare structural).

## Evidence

- **Suite**: 615 lib tests under `--all-features` (565 default), all
  targets green, clippy `-D warnings` clean, `#[ignore]` zero among
  additions, no sleeps/timers/wall-clock reads in the kernel or the
  driver (one bounded, clock-free yield-counted wait in the
  mid-batch handshake fixture).
- **Repeat stability**: the conformance suite 10× identical; the
  threaded rows 15×.
- **Mutation battery** (each applied, observed, reverted): stage
  reorder, revocation-filter removal (data lane, control lane, and
  both), park-to-spin, dirt-condition widening, init-quit
  short-circuit removal, control-lane accounting removal — every
  mutation is caught, and the park and accounting mutations are
  caught by exactly the tests that claim them. An independent review
  pass reproduced four of these from scratch.
- **Structural reviews** (RFC 0014's structural halves): the kernel
  has one `select!`, arming all three wake sources, with no
  per-source priority, quota, or ordering state anywhere
  (INV-RC16, §3.5's unbiased initiation); no `biased;` directive
  exists in the kernel.
- **loom**: the delivery-accounting mirror (line-for-line with the
  implementation) passes its four models beside the existing
  query-cell models; the two historically rejected encodings still
  reproduce their counterexamples when substituted (negative
  control).
- **Four threaded rows** (mid-batch quit, cancel-beats-quit,
  buffered-quit retraction, and the simultaneous-readiness producer
  face) run on a multi-worker executor with their
  determinism supplied by an explicit handshake, not the scheduler;
  they cite no INV-RC14 guarantee (that claim's verified range is
  current-thread) and stay pass-unit driven, so they sit inside
  RFC 0008 §9.9's evidence surface.

## Deliberate exceptions and their grounds

- Seven unit tests inside the crate-private `structural_key` module
  were mechanically rewritten for the root-first inversion (the one
  place tests changed; every assertion keeps its meaning, and the
  public-facing identity tests are untouched and green).
- `Command::actions` (a `#[cfg(test)]` constructor) exists because a
  producer-originated quit is not otherwise constructible from
  today's `Command` surface; its public spelling belongs to the
  effect-constructor split RFC 0014 §3.4 implies, decided at the
  switch.

## Points for review discussion

1. **`TestDriver::on_worker_threads`** (`#[cfg(test)] pub(crate)`) is
   a construction knob RFC 0008 §9.3's block does not name: same
   construction path, same seams, only the executor differs. It is
   what makes the four threaded rows constructible. Whether a public
   form should exist belongs to the RFC 0008 addendum that resolves
   RFC 0014 §13.3 (executor-independent driving).
2. **`step_pass` takes one executor turn** between its readiness check
   and the pass — the turn production takes at its park, but taken
   unconditionally. Whether this is part of the send-gate seam's
   handshake machinery or a third driving differential to record is a
   question for the same addendum.
3. **`ParkProbe` series poll `Kernel::run`** today; RFC 0008 §9.7
   names `ProgramRuntime::run`, which lands at the switch — the
   witness moves to the named route then.

## Relation to the contract documents

RFC 0008 §9 (merged in #285) is the driving-layer contract this
implements; two of its rounds were shaped by this implementation
(the `ParkProbe` accessor reduction and the grant-reclamation route),
which is the direction the process intends — implementability
findings went back into the contract rather than being worked
around. RFC 0014 §13.1's implementation-acceptance tier stays open:
cleanup hooks, the combinator surface, the observability rows, and
the load re-derivation land with later changes, and the RFC status
transitions wait for the switch.


akiomik:
Thanks — all seven are addressed in `85fc2fb` and `83d9829`.

**1. `ExitReport::joined`.** You are right that the doc had it
backwards. `exit_buf` holds exits that were *observed* but not yet
reflected; a reflected exit left the join set when it was observed
and left the buffer when the pass reflected it, so this drain sees it
in neither place. The field doc and the comment now say that, and
name the two groups the count actually covers.

**2. Row count.** Fixed at the second site, and the whole suite was
re-swept for the same class of claim
(`(one|two|…|twelve) (rows|series|cases|arms|halves|members)`) rather
than only the line you named. Thirteen hits, one stale — the one you
found. The PR description's "three threaded rows" is also corrected
to four, naming the simultaneous-readiness producer face.

**3. The bounded "not yet committed" half.** Your diagnosis was
exactly right, and following it turned up a second constraint worth
recording. `try_confirm` lands as a test-only internal
(`peek_resolution` on the gate, non-consuming; the `take_resolution`
discipline is untouched), but it does not on its own rescue the
assertion: `grant` turns nothing, and the *only* call that supplies
turns while a grant is outstanding is `step_pass`, which always
drains. So there is no observation point at which the lane is still
full and the producer has run — `try_confirm == None` placed there
would have been as empty as the ledger read it replaces.

What carries the claim instead is a negative row per site: with no
dequeue, `confirm` spends its whole budget and fails
(`#[should_panic]`). That is the contrapositive spent for real, and
it is what rules out a release that would have committed anyway. The
positive rows keep `try_confirm` for the one thing they can say at
that point — the dequeue freed the slot but did not itself commit the
send. The zero-budget form was verified unreachable: after
`step_pass`, the drain wakes the blocked sender but nothing polls it
inside that call, so `confirm(0, …)` exhausts. `try_confirm`'s public
form joins `on_worker_threads` on the RFC 0008 addendum's agenda.

**4. `lower()`'s placeholders.** The module doc now carries their
lifecycle: unreachable before the switch (nothing outside the crate
constructs the kernel), and deleted at the switch, which lands
together with the effect-constructor split that makes both shapes
unconstructible. The two are not separable — landing the switch alone
is precisely what would expose the debug-build panic, and that
coupling is now a recorded acceptance criterion for the switch. The
fallback was not taken: logging and lowering the first carrier would
pick a meaning for a shape that has none, silently, which is the
failure class RFC 0007 INV-C5 exists to prevent.

**5. `starts_with`.** Swapped to `allow`, with the reason rewritten
to the transitive-reachability fact rather than the stale one. The
lint set here requires `expect` over `allow`, so the two sit paired
on the item; both the test and non-test builds are clean under
`-D warnings` on the pinned toolchain.

**6. `Command::batch`'s warning.** No code change; the comment above
it now records that it stops being true — and is removed — with the
switch that lands RFC 0014 §9 row 3.

**7. Prefix walk.** No code change; `select_prefix`'s doc notes that
its cost in live runs is mechanism (RFC 0013 §3.7) and is measured
where load acceptance is re-derived on this topology (RFC 0014
§13.5, under RFC 0006's ownership).

Suite: 613 lib tests green, clippy `-D warnings` clean for both the
test and non-test builds on the pinned toolchain, fmt clean, and the
four bounded rows repeated ten times.

akiomik:
Thanks for running the mutations — M3 found a real hole, and both
points are addressed in `a8db66b` and `bf09026`.

**A.** Both call sites asserted `None`, which a peek that reads no
state at all satisfies equally well — and returning `None`
unconditionally from `peek_resolution` did leave the whole suite
green, exactly as you found. Two rows close it.

The first takes your construction: a marking run's end makes
`ProducerExit` ready, so the step is admitted and its own leading
turn hands the parked run its release, which commits before the pass
runs. `try_confirm` then reads `Accepted`, and the `confirm` after it
reads `Accepted` too — the same row witnessing that the peek took
neither the token nor the grant.

The second pins the property that explains the two `None`s and that
this API's callers are most likely to trip over: a refused step
drives nothing, not even a turn. An empty journal is equally what a
step that spent a turn and delivered nothing would leave; an
unresolved grant is not, and the confirm at the end shows one turn
was all the release was waiting for.

Re-running M3 against the updated suite: the mutation now fails
exactly one test — the new positive row — and the suite with that row
skipped is still entirely green under it, which confirms both the
hole and that this is what closes it. Reverted, tree clean.

**C.** Agreed, and changed. The `&mut self` was reasoned from
RFC 0008 §9.3's normative receiver, but the call is outside that
block by its own documentation and drives nothing — no turn, no token
taken, no grant cleared — so it now sits with `accepted` and
`intents` on `&self`. The impl-level `needless_pass_by_ref_mut`
expectation stays fulfilled by the driving calls; both the test and
non-test builds are clean under `-D warnings` on the pinned
toolchain.

**B.** The description's counts are updated (42 conformance tests,
615 lib tests under `--all-features`, and the two remaining "three
threaded rows" now say four).

Suite: 615 lib tests green, fmt clean, and the two new rows repeated
ten times.

akiomik:
All four are in `3df8578` and `57b0c57`.

**1 & 2 — identity.** Both were real: run tokens count from one in
every kernel and grant sequences from one per gate, so two drivers'
first runs compared equal and one driver's token matched the other's
outstanding grant by number alone. Each now carries the identity of
the driver that minted it, from a process-wide monotone counter — no
clock, no entropy, and nothing observable about the value beyond
"not the other driver's". Using either elsewhere is misuse rather
than a silent success: a foreign name denotes a run this kernel never
started, which is the bookkeeping rule's own case, caught one step
earlier. Three rows cover it — the two first runs are not equal,
granting at a foreign name panics, confirming a foreign token panics.

**3 — the bootstrap panic.** Reproduced, then fixed at the source:
the kernel records a run's kind when it starts it, and the report
carries that rather than a lookup performed after the continuation
pass may already have retired the entry. The mint reads no
bookkeeping at all now, which also retires its old ordering
constraint against `settle`.

Worth noting on the regression row: the bare construction only
reproduced about one run in six, so it would have been a guard in
name only. The fixture now holds the *initial reconcile* until the
init run has ended — bootstrap's order makes that the right seam —
with a counted tail of scheduler yields to close the gap between a
producer's own mark and the task behind it becoming joinable, which
no application surface reports. Fail-first was eight of eight before
the change, and it is deterministic after it.

**4 — the nested carrier.** Fixed by moving the check onto the
carriers, with one wrinkle worth recording. Grouping by key id alone
flags a shape §3.4 explicitly allows — two children each keyed with
the same id, lowering to consecutive dispatches with the second a
replacement under its own policy — and after `batch` has pushed a key
down, that shape and "one key across two carriers" are
indistinguishable from the carriers. So the push records what that
one attachment reached, on each carrier it fills, and the lowering
asserts over those plus the command-level key's own reach. Nesting
depth stops mattering because nothing is asked of the command. The
recording is `debug_assertions`-only, like the assertions it serves.

The assertions stay at the lowering site rather than moving into
`attach_key`: that call runs in every build, including the current
runtime's, where the documented behaviour is to warn and ignore —
asserting there would turn that into a debug-build panic for today's
users. The lifecycle note already on the module (unreachable before
the switch, deleted with the type split at it) applies unchanged.

Suite: 621 lib tests green, clippy `-D warnings` clean on the pinned
toolchain for both builds, fmt clean, and the six new rows repeated
ten times.

akiomik:
You're right, and thanks — that row was not scripted, it was just
likely. The mark happens inside the run's body, before the task
carrying it is finished and joinable, and the counted yields after it
establish nothing about the worker; a worker left unscheduled long
enough and the superseded mint passes the row too.

Replaced with your option (b), in `a3c8db2`. The state is now
constructed outright on the current-thread executor: a pass starts
the run and leaves its token in the kernel's `started` log — nothing
in that block is a driving call, so nothing drains it — turns run
until the kernel's *own* readiness read reports the exit, which is
the fact the wait is waiting for rather than a proxy for it, and the
next pass reflects and retires the entry. The log is then drained by
hand and minted from, with the registry asserted empty for that run
first. No scheduling order is load-bearing anywhere in it.

It is marked `whitebox_` and says so: driving stages directly puts it
outside the evidence surface, and what it guards is one property of
the harness — that the mint reads the kind recorded at the start and
consults no bookkeeping a pass is free to retire — rather than a
contract. The conformance suite's own `whitebox_` note now points at
it so "this suite has none" is not read as a crate-wide claim.

Fail-first against the superseded lookup-based mint: six of six
panicked, and because the construction is deterministic that is the
whole story rather than a rate. The raced `on_worker_threads` row is
deleted along with the fixture it needed; the field record of
reproducing it there stays in the commit history.

Suite: 621 lib tests green, clippy `-D warnings` clean on the pinned
toolchain for both builds, fmt clean, new row repeated ten times.

PR #289, opened 2026-08-18
feat(reducer): land the cleanup hooks and the composition surface

akiomik:
## Summary

Second scaffolding change of the kernel mainlining: **cleanup hooks
(RFC 0014 §4.4, INV-RC8) and the composition surface (RFC 0014 §2.5,
INV-RC2/RC3)** — `Command::on_teardown` and the cleanup ledger on the
kernel side; `ScopeValue`, `Keyed`/`Slot` with removal journals, and
`ReducerExt`'s `scope`/`for_each`/`presented`/`into_program` on the
reducer side. Everything is `pub(crate)`; production behavior and the
public API are unchanged (`tests/api_surface.rs` green, public API
diff empty), and the old runtime remains the production owner.

The combinator signatures are RFC 0014 §2.5's block verbatim — no
aliases, no reshaping — with one deliberate exception recorded in the
findings below (`ScopeValue`'s bound list).

## What lands

1. **Cleanup hooks.** Registration travels the command as its own
   carrier (qualified by `scoped` and the combinators — RFC 0005
   INV-18's coverage), arms in the spawn phase after the same
   command's cancel phase (§3.4's teardown-and-reregister rule), and
   starts when a teardown selects its scope. A cleanup run joins the
   kernel's one `JoinSet` but counts toward no gauge, neither joins
   nor triggers the admission barrier, marks no dirt, and has **no
   route back into the runtime**: its harness holds a join set, a
   task index, and the gate — no lane sender, no directive path. That
   absence is the no-output clause's structural review (INV-RC8),
   written at the construction site. Termination discards unfired
   registrations and cancels running finalizers, with no grace
   window.
2. **Collections with removal journals.** `Keyed` and `Slot` record
   *removals* — `remove`, `dismiss`, insert-over-occupied,
   present-over-occupied — never state diffs, so a same-update
   remove-and-reinsert still yields the old instance's teardown and a
   fresh successor in one command (INV-RC3). Building a collection
   (`from_iter`, literals) records nothing: an entry that only ever
   existed in an iterator never placed a run, so there is nothing to
   tear down.
3. **Combinators.** Each boundary qualifies every identity carrier of
   its child's returned command — spawn keys, cancel ids, teardown
   prefixes, cleanup registrations — and the child's subscription
   declarations, with its segment (INV-RC2); equal local ids under
   sibling boundaries never alias; stacks nest (a whole
   `for_each`/`presented` stack composed as another boundary's child
   is exercised end to end). Journal drains merge one teardown per
   removal into the update's own command **without constructing a
   batch**: the entries are appended to the command's teardown
   carrier directly, so a boundary adds identity carriers and
   nothing else — the update's redraw directive (including
   `without_redraw`) survives, and no spurious diagnostics fire.
   Origination stays one surface (RFC 0013 R8): `Command::teardown`
   is the only constructor of a teardown entry crate-wide.

## Evidence

- **Suite**: 687 lib tests under `--all-features` (+66 over `main`),
  all targets green, clippy `-D warnings` clean in debug and release,
  fmt clean, no sleeps/timers/clock reads, `#[ignore]` zero among
  additions. New and changed rows repeated ×10.
- **Counterexamples killed, with the mutation run** (applied,
  observed, reverted — an independent review pass re-ran them):
  - *diff-based removal detection*: a diff-equivalent journal fails
    the kernel recreate row on its **unkeyed-run** assertion (an
    unkeyed run has no identity to supersede, so only the journal's
    teardown reaches it) plus five unit rows; the fixed suite
    passes.
  - *batch-fold merge* (the shape the boundary drain deliberately
    avoids): restoring it fails exactly the two directive/diagnostic
    rows — `without_redraw` preserved, no false "child key ignored"
    warning (with an in-recorder control showing the zero is
    silence, not suppression).
  - *cancel-phase cleanup registration*: moving the registration to
    the cancel phase fails the three spawn-phase rows.
- **RFC 0013 §7.2's kernel rows** are all present, including the
  ST8 routing pair (a key-addressed message after a same-key
  remove-and-reinsert reaches the **new** instance's state; one for
  an absent key is discarded) and the final-update row (teardown and
  quit in one update; termination fires no hook). The final-update
  row reads the two RFC 0011 §4.4 postconditions through the
  driver's terminated report plus the bounded settle; the
  postconditions proper are owned by the existing termination row.
- **R8 one-surface review** re-verified by grep after the merge
  redesign: one origination site; `merging_teardowns` and `batch`
  extend already-originated entries.

## Deliberate exceptions

Two pre-existing test-code edits, both mechanical, no assertion
semantics changed: an exhaustiveness-forced match arm in a lowering
test helper (the new dispatch step variant), and an `.expect(...)`
where a test-surface constructor became `Option`-returning
(structural replacement of an `unreachable!`).

## Findings for the contract documents

1. **INV-RC12 (a)'s cleanup half cannot be behavioral.** A
   stop-requested cleanup run is not constructible outside
   termination — teardown selection excludes the kind, cancel and
   supersession address keyed slots, re-evaluation addresses
   subscription runs, and at termination no admission site remains.
   Verified independently by enumerating every stop-request call
   site. The reachable half (an in-flight cleanup run defers no
   admission) is behavioral here; the rest is carried structurally by
   the barrier predicate reading subscription runs only. RFC 0014
   §12 marks all of (a) behavioral — an enforcement-class amendment
   candidate for the RFC, not an implementation change.
2. **A boundary-merged teardown does not redraw; a hand-written
   `Command::teardown` does** (it inherits the default directive).
   Both are intended — directives belong to the update, and a
   boundary adds identity carriers only — but the sentence belongs in
   RFC 0014 §2.5 when this surface goes public at the switch.
3. **`ScopeValue` drops `PartialEq`**, which `Eq: PartialEq` makes
   redundant. RFC 0014 §2.5's block still writes it; a wording sync
   for the RFC when this surface goes public, not an implementation
   change.
4. `Keyed` iterates in insertion order off a linear scan —
   determinism is what the driving contract rides on; the O(n)
   lookup is stated mechanism (RFC 0013 §3.7) and joins the
   load-measurement list for the acceptance re-derivation.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -25,6 +25,7 @@
 //! ```
 
 mod cancellation;
+mod cleanup;
 pub(crate) mod core;
 mod effect;
 mod kernel_parts;
@@ -34,6 +35,7 @@ mod runtime_parts;
 
 pub(crate) use cancellation::CancellableCommand;
 pub use cancellation::{CancelPolicy, CommandId};
+pub(crate) use cleanup::CleanupRegistration;
 pub(crate) use core::Command;
 pub(crate) use kernel_parts::{KernelParts, SpawnEntry};
 pub use retry::{RetryBackoff, RetryContext, RetryError, RetryPolicy, RetryStopReason};
--- a/src/command/cleanup.rs
+++ b/src/command/cleanup.rs
@@ -1,0 +1,66 @@
+//! The cleanup-hook carrier: what [`Command::on_teardown`] puts on a
+//! command.
+//!
+//! A registration is command *metadata*, like a teardown prefix and unlike
+//! an effect leaf: dispatching the command arms it and starts nothing. What
+//! starts it is a later teardown whose prefix covers its scope, at that
+//! teardown's application point (RFC 0014 §4.4, RFC 0013 §5).
+//!
+//! The finalizer's `Output = ()` is the signature half of INV-RC8's
+//! no-output clause: there is no message for a cleanup run to return, so the
+//! path back into the runtime is closed in the type rather than by the
+//! kernel remembering to ignore something. The other half is structural and
+//! lives at the run's construction site, where the kernel hands a cleanup
+//! run no lane sender and no directive capability
+//! (`kernel::producer::CleanupHarness`).
+//!
+//! [`Command::on_teardown`]: super::Command::on_teardown
+
+use futures::FutureExt;
+use futures::future::BoxFuture;
+
+use crate::structural_key::{ScopePath, StructuralKey};
+
+/// One armed finalizer and the composition boundary it is registered
+/// against.
+///
+/// It carries no message type: a cleanup run produces none, so a
+/// registration survives [`Command::map`](super::Command::map) unchanged
+/// rather than being re-wrapped at every boundary that maps a message type.
+pub struct CleanupRegistration {
+    /// The scope this registration is anchored at — the path a teardown
+    /// prefix has to cover to consume it (RFC 0013 INV-ST1).
+    ///
+    /// Empty at construction and qualified by
+    /// [`Command::scoped`](super::Command::scoped) and by the combinators,
+    /// exactly as a teardown prefix is (RFC 0005 INV-18's coverage).
+    pub scope: ScopePath,
+    /// The finalizer itself.
+    pub finalizer: BoxFuture<'static, ()>,
+}
+
+impl CleanupRegistration {
+    /// A registration anchored at the root, for a finalizer under
+    /// `Command`'s ordinary effect bounds.
+    pub fn new(finalizer: impl Future<Output = ()> + Send + 'static) -> Self {
+        Self {
+            scope: ScopePath::empty(),
+            finalizer: finalizer.boxed(),
+        }
+    }
+
+    /// Qualifies this registration with one already-erased boundary segment.
+    ///
+    /// Takes the erased segment rather than the scope value so that one
+    /// `scoped` call can share a single erasure across every carrier at its
+    /// boundary — the spawn key, the explicit cancels, the teardown
+    /// prefixes, and this — without the scope type itself being `Clone`
+    /// (RFC 0005 §8.1).
+    #[must_use]
+    pub fn scoped_with(self, segment: StructuralKey) -> Self {
+        Self {
+            scope: self.scope.prefixed_key(segment),
+            finalizer: self.finalizer,
+        }
+    }
+}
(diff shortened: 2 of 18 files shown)


akiomik:
Thanks — all six addressed in `74d3c2c`.

**1. `ForEach::reduce`'s rationale was wrong.** You're right that a
child cannot reach the collection: it is handed the projected
`&mut C::State` and nothing else. So only the parent branch can
record, and that branch drains what it just recorded. What the drain
on the *other* branch is actually for is an entry recorded **before**
the reduce — by a mutation outside one, which the collection module's
note discourages but nothing prevents. Such an entry is owed its
teardown whichever branch the next message takes. The doc now says
that and names `a_pending_removal_survives_a_discarded_message` as
its row; `Presented`'s "for the same reason" points at the same
thing.

**2. The corollary that follows from it.** A boundary drains only
when its own `reduce` runs, so a message an outer boundary claims
does not drain the boundaries in its parent chain. This is invisible
for entries a boundary recorded itself (recorded on the branch that
drains) and observable only for the outside-a-`reduce` case above.
Nothing is lost — the next message that reaches the boundary pays it
in full — but the teardown is deferred. New module section plus
`an_outer_claim_leaves_an_inner_boundary_s_pending_removal_undrained`,
which builds exactly that shape: the outermost `presented` claims the
message and the `for_each` below it never runs, then the next message
reaching it yields the pending teardown. Recorded as the wall a
future INV-RC3 extension would meet — making the timing unconditional
would mean draining every boundary in the stack per message, which is
a different contract from the one §2.5 states.

**3.** `merge` takes an `IntoIterator` now and `Presented` passes a
lazy `repeat_with(...).take(dismissals)`, so an update that dismissed
nothing clones nothing.

**4.** `PartialEq` dropped from `ScopeValue` and its blanket impl.
RFC 0014 §2.5's block still writes it; the code doc notes the wording
is synced when that surface is made public at the switch.

**5.** The four shapes being the whole mutation surface is now stated
as a decision rather than left implicit, with the requirement on any
future bulk operation: one journal entry per removed instance, or
§11's diff-based adversary walks back in through a convenience
method.

**6.** The ledger lookup's `expect` now says what makes it total — a
cleanup run is constructed with no lane sender at all, so it can
never record — and that widening `CleanupHarness` and this lookup
have to move together.

Full suite green (687), clippy clean in debug and release, fmt clean,
new and touched rows stable over 10 repeats.

akiomik:
The description is synced: the "verbatim" sentence now carries its one deliberate exception, a fourth finding records the `ScopeValue` bound divergence as an RFC wording sync (not an implementation change), and the suite count is updated to 687.

PR #290, opened 2026-08-19
feat(kernel): close the implementation-acceptance evidence and its contract text

akiomik:
## Summary

Third scaffolding change of the kernel mainlining: **the
implementation-acceptance evidence RFC 0014 §13.1 still listed as
open** — the observability vocabulary on the kernel (RFC 0014 §9
row 9), the remaining RFC 0011/RFC 0012 conformance rows, the
production/driver divergence recount, the load-acceptance
re-derivation (RFC 0014 §13.5 → RFC 0006 §5.3), and the bounded
half of the driving contract's determinism question (RFC 0014
§13.3 → RFC 0008 §9.12) — plus the RFC text those resolutions
land in. Production behavior is unchanged; the kernel remains
crate-private; the old runtime remains the production owner.

## What lands

**Code.**

1. The kernel's input batch reports the existing batch event —
   same target, same field names, same schema (the files owning the
   schema have a zero-line diff). `pulled` counts dequeues against
   the cap (revoked discards included, RFC 0006 INV-L12's unit),
   `shared_pending` reads the data lane's residue including the
   park buffer. Capacity waits from all three producer kinds report
   `channel = "data"` (INV-RC15's behavioral neighbour, three
   rows).
2. The remaining INV-RC11 rows (RFC 0011's lifecycle suite on the
   kernel: render-once/re-evaluate-once, render observes the
   batch's final state, construction inertness, panic containment
   for every producer kind, and more) and INV-RC12's admission
   suite (RFC 0012's INV-SE1–SE5 rows, including the mandated
   supersession window, constructed for real — see below).
3. Conformance additions the review cycle demanded: a full-lane
   **replay determinism** row (two keyed producers through a
   one-slot lane, three genuine capacity waits asserted, eight
   replays per script face), and the INV-SE4 window row on the
   multi-worker harness (a source whose dismantling is held on a
   worker keeps the stopped run un-quiesced while the successor
   generation is installed; both generations admit zero inside the
   window; the successor admits after quiescence).
4. Benches: `kernel_load` (the RFC 0006 scenario set plus the two
   quit routes and a decomposition series) and `kernel_scan` (the
   registry's per-dispatch walks). Acceptance rows pin their
   parameters explicitly (`batch_max_messages = 1024`, render cost
   500 µs); no bench carries a pass/fail latency assertion.

**Contract documents.**

1. **RFC 0006 §5.3** re-derives INV-L4 on the kernel topology, per
   quit route (the two routes are separate contract objects after
   RFC 0014 §9 row 10). The producer route's acceptance is
   quantified over RFC 0014 §3.3's constructive bound, and the
   measured decomposition matches the predicted batch remainder
   within 0.5%. The synchronous route's tail is a one-sided bound
   (2.5× the p50 predictor). The instrument is an upper bound
   (termination return), stated with its conservative direction; no
   new schema is introduced. The registry-walk figures land as an
   informative paragraph. The render-count change (frame pacing
   removal, RFC 0014 §6.3) is recorded and explicitly not gated.
2. **RFC 0008 §9.12** resolves RFC 0014 §13.3's bounded half: the
   scripted-determinism claim's verified range becomes **a
   current-thread executor on either lane mode**, with the replay
   row named as the verification pass; executor independence stays
   open. The reclaimed grant's ack-correlation reading is fixed
   (the next grant follows the previous grant's resolution; a
   reclaimed one leaves no commit to correlate). `try_confirm` and
   `on_worker_threads` are promoted into §9.3's block; `step_pass`'s
   unconditional pre-pass turn is recorded as the third driving
   differential rather than removed (RFC 0008 §9.2).
3. **RFC 0014 / RFC 0012**: INV-RC12 (a)'s cleanup clause is
   reclassified structural in both owners at once (a stop-requested
   cleanup run is not constructible outside termination — every
   stop-request call site enumerated in both documents); the
   in-flight cleanup row stays behavioral as the clause's
   neighbour. §13.3/§13.5 close with pointers to their resolutions.
4. **RFC 0006 §5.2** records the one reach change in the batch
   event's firing, which follows from RFC 0014 §3.3's quit
   synchronization (row 4), not row 9: the rule is unchanged, but
   the two quit routes moved in opposite directions — an
   update-returned quit now ends a batch that emits no event, while
   a producer quit no longer cuts a batch at all — so neither
   suppressed set contains the other.

## Evidence

- Suite: 713 lib tests under `--all-features`, all targets green,
  clippy `-D warnings` clean in debug and release, fmt clean,
  `cargo doc` clean under `-D warnings`, loom 8 models, repeat ×10
  identical (the threaded window row ×20–30).
- Production/driver divergence recount: `pass_cycle` has one implementation; the
  mode-dependent branches in the crate are exactly three, all
  inside the send-gate seam; the driving differentials are the two
  seams plus the recorded pre-pass turn.
- INV-L4 figures from the reference machine (RFC 0006 §2), ≥200
  valid trials per quit row (the bounded-depth load rows are
  single-run), 2026-08-19. The batch-remainder decomposition
  matches prediction within 0.5%. Backlog independence is decided
  by its own validating run, 2026-08-24, whose isolation window is
  bracketed at both ends: slope −0.1200 ns/envelope against a
  bound of +0.5 and span 0.102 ms against 0.6, over depths 1,024
  to 299,999. The verdict is `da5c70a` (RFC 0006 §5.3).
- An independent adversarial pass ran twice (sixteen findings in
  the first round, all closed and re-verified in the second; the
  second round's two additions are in this head). Its recomputation
  of the tail-bound ratios and the decomposition arithmetic matched
  the record to two decimals.

## API note

No additions to the default-build public API. Under
`--all-features` the branch adds four `#[doc(hidden)]` re-exports
behind `bench-internals` (the bench harness's kernel handles);
`tests/api_surface.rs` builds with `all_features(true)` and is
`#[ignore]`d pending nightly, so those four are unchecked by CI.

## What this closes and what remains

RFC 0014 §13.1's implementation-acceptance tier is now closed for:
cleanup hooks, the combinator surface, the observability
vocabulary, the load re-derivation, and the remaining behavioral
rows. What remains before the switch: the production arbitration
policy's final structural walk on the switched tree, the
store-parity extension, and the switch itself with its
public-surface changes — the next change in the sequence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,18 +67,25 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
       - uses: taiki-e/install-action@just
-      - name: Run subscription benchmark
+      - name: Run criterion benchmarks in test mode
         # `--bench <name>` (singular, per-target) rather than `--benches`
         # (plural): the plural form implicitly also runs the lib's unit
         # tests, duplicating what the `test` job's matrix already covers.
-        run: cargo test --bench subscription --all-features
-      - name: Run runtime-load smoke profile
-        # The full `runtime_load` scenarios leave CI: they are deliberate
-        # acceptance/regression runs (RFC 0006 §5.1, RFC 0007 §5–6), and their
-        # latency numbers gate nothing on a CI machine. CI runs only the
-        # reduced, latency-assertion-free smoke profile, which gates on
-        # completion. Invoked via `just` so the local and CI invocations are
-        # identical (RFC 0007 §6).
+        # Criterion's own test mode is the smoke profile for these two: one
+        # iteration each, no sample, so what it gates is that they still
+        # compile and run.
+        run: |
+          cargo test --bench subscription --all-features
+          cargo test --bench kernel_scan --all-features
+      - name: Run the load harnesses' smoke profiles
+        # The full scenarios of both load harnesses leave CI: they are
+        # deliberate acceptance/regression runs (RFC 0006 §5.1, RFC 0007
+        # §5–6, RFC 0014 §13.5), and their latency numbers gate nothing on a
+        # CI machine — a shortened row's percentiles are not the quantity the
+        # acceptance matrices are stated over. CI runs only the reduced,
+        # latency-assertion-free smoke profiles, which gate on completion and
+        # sequence integrity. Invoked via `just` so the local and CI
+        # invocations are identical (RFC 0007 §6).
         run: just bench-smoke
 
   loom:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,10 +52,11 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Run tests
         # Benches are excluded from this matrix (and run once, on Linux
-        # only, in the `bench` job below) since `runtime_load`'s custom
-        # harness ignores `cargo test`'s `--test` flag and always runs its
-        # full scenarios; matrixing that across every OS/toolchain makes
-        # Windows CI disproportionately slow for no additional coverage.
+        # only, in the `bench` job below) since the load harnesses'
+        # (`runtime_load`, `kernel_load`) custom mains ignore `cargo test`'s
+        # `--test` flag and always run their full scenarios; matrixing that
+        # across every OS/toolchain makes Windows CI disproportionately slow
+        # for no additional coverage.
         run: cargo test --lib --bins --tests --examples --all-features
 
   bench:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,19 @@ name = "gauge"
 harness = false
 required-features = ["bench-internals"]
 
+# The kernel's load-acceptance re-derivation (RFC 0014 §13.5). Reaches the
+# kernel through the `bench-internals` handles; `benches/runtime_load.rs` is
+# the superseded topology's harness and stays as the before column.
+[[bench]]
+name = "kernel_load"
+harness = false
+required-features = ["bench-internals"]
+
+[[bench]]
+name = "kernel_scan"
+harness = false
+required-features = ["bench-internals"]
+
 [[example]]
 name = "websocket"
 required-features = ["ws", "rustls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,9 +27,15 @@ include = [
 default = []
 
 http = ["dashmap", "thiserror"]
-# Bench-only feature exposing internal-only subscription reconciliation hooks
-# to `benches/subscription.rs`; do not enable for normal builds. Not part of
-# the public API and not covered by semver.
+# Bench-only feature exposing internal-only handles to the benches that need
+# them, and to nothing else: the subscription reconciliation hooks
+# (`benches/subscription.rs`), the load observer (`benches/gauge.rs`), and the
+# kernel's driving and bookkeeping handles together with `Command::actions` —
+# the only route to a producer-originated quit — for the load-acceptance and
+# registry-scan harnesses (`benches/kernel_load.rs`, `benches/kernel_scan.rs`).
+# Do not enable for normal builds. Not part of the public API and not covered
+# by semver: every item it adds is `#[doc(hidden)]`, and a default build
+# compiles none of them.
 bench-internals = []
 # Test-only feature for the isolated loom cell core; do not enable for normal builds.
 loom-core = ["dep:loom"]
--- a/benches/kernel_load.rs
+++ b/benches/kernel_load.rs
@@ -1,0 +1,1097 @@
+//! Load harness for the reducer-first kernel — the re-derivation half of
+//! RFC 0014 §13.5, which RFC 0006 §5.2 names as INV-L4's prerequisite.
+//!
+//! `benches/runtime_load.rs` measures the superseded topology and is left
+//! untouched: it is the *before* column. This harness runs the same shapes of
+//! load against the kernel's own driving loop (`Kernel::drive`, reached
+//! through the bench-only `BenchKernel` handle), so the two columns differ in
+//! the topology under test and in nothing else the harness controls.
+//!
+//! # What the new topology changes about the measurement
+//!
+//! - **The quit route split.** RFC 0006's `quit_*` scenarios have `update`
+//!   return `Command::quit()`, which in the old topology became an effect
+//!   stream item travelling the dedicated quit channel — the channel INV-L4 is
+//!   about. On the kernel that spelling is the **synchronous** route, applied
+//!   at its own dispatch with no lane involved (RFC 0014 §3.3). The lane route
+//!   INV-L4's property carries to is the **producer-originated** quit, so
+//!   every quit row here is run twice: `…_sync` (update-returned) and
+//!   `…_control` (a spawned run emitting one quit on the control lane).
+//! - **Where the measured interval ends.** The old harness timestamps the
+//!   runtime's `quit signal received` tracing event, emitted from inside the
+//!   `select!` arm before any shutdown work. The kernel emits no counterpart —
+//!   its control drain applies the quit directly — so what this harness can
+//!   observe is `Kernel::drive` returning, which is the quit's application
+//!   **plus its immediate postcondition** (receivers dropped, buffers cleared,
+//!   every run revoked and aborted; RFC 0011 §4.4). The reported
+//!   `quit->applied` is therefore an **upper bound** on delivery, not the
+//!   delivery instant: it contains a term that scales with the data lane's
+//!   residual occupancy, which is why every row also reports the depth at the
+//!   quit instant. In the bounded rows that residue is capped at
+//!   `capacity + producers`, so the bound is tight there; in the unbounded
+//!   deep-backlog rows it is not, and the number must be read as a bound.
+//!   Whether the successor's acceptance instrument stays this bound or gains a
+//!   delivery-instant event is a contract question for RFC 0006, not a
+//!   harness choice — this file only reports what it can see.
+//! - **INV-L9 has no counterpart.** The `keyed_isolation` scenario quantifies
+//!   over the per-`CommandId` private channels the kernel removes (RFC 0006
+//!   §5.2 records the property loss), so it is absent here rather than
+//!   re-measured.
+//! - **No frame pacing.** The kernel's stage 4 renders once per pass with a
+//!   redraw pending and reads no clock (RFC 0014 §3.5, §6.3), so `FrameRate`
+//!   is not an input. `RuntimeConfig::new` still requires one to construct;
+//!   the kernel never reads it (`RuntimeConfig::kernel_controls`).
+//!
+//! # Rows
+//!
+//! - **`overload`, `burst_200k`, `steady_20k`** — INV-L1's depth bound,
+//!   INV-L2's losslessness, and INV-L3's update latency, unbounded and under
+//!   the RFC 0007 §5.1 bounded capacity.
+//! - **`quit_*`** — INV-L4's statistical rows, both routes and both modes,
+//!   `QUIT_TRIALS` trials each.
+//!
+//! Run all rows, or name a subset:
+//!
+//! ```bash
+//! cargo bench --bench kernel_load --features bench-internals
+//! cargo bench --bench kernel_load --features bench-internals -- overload
+//! ```
+
+// Metric reporting converts counters and nanosecond values to floating point
+// for human-readable output; precision loss there is irrelevant.
+#![expect(
+    clippy::cast_precision_loss,
+    clippy::cast_possible_truncation,
+    clippy::cast_sign_loss,
+    reason = "metric reporting casts counters and nanosecond values to floating point for human-readable output; precision loss there is irrelevant"
+)]
+
+use std::collections::BTreeMap;
+use std::collections::btree_map::Entry;
+use std::fmt::Debug;
+use std::num::{NonZeroU32, NonZeroUsize};
+use std::process::ExitCode;
+use std::sync::Arc;
+use std::sync::Mutex;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::time::{Duration, Instant};
+use std::{env, hint};
+
+use futures::stream::{self, StreamExt};
+use ratatui::Terminal;
+use ratatui::backend::TestBackend;
+use tears::prelude::*;
+use tears::{BenchKernel, BoxStream, FrameRate, RuntimeConfig, SubscriptionSource, producer_quit};
+use tokio::runtime::{Builder, Runtime as TokioRuntime};
+use tokio::task::yield_now;
+use tokio::time::{MissedTickBehavior, interval, timeout};
+use tokio_stream::wrappers::IntervalStream;
+use tracing::field::{Field, Visit};
+use tracing::level_filters::LevelFilter;
+use tracing::span::{Attributes, Id, Record};
+use tracing::subscriber::set_global_default;
+use tracing::{Event, Level, Metadata, Subscriber};
+
+/// Message rate for rows that emit their whole load in one burst.
+const BURST: u64 = 0;
+
+/// RFC 0007 §5.1's bounded capacity, carried over unchanged so the bounded
+/// column compares against the old harness's.
+const DATA_LANE_CAPACITY: usize = 1024;
+
+/// Trials per quit row — RFC 0006 INV-L4's "≥ 200 trials per scenario".
+const QUIT_TRIALS: u32 = 200;
+
+/// Attempts a quit row may spend collecting `QUIT_TRIALS` valid trials.
+const QUIT_ATTEMPT_CAP: u32 = 4 * QUIT_TRIALS;
+
+/// Terminal size every row renders into.
+const SCREEN: (u16, u16) = (80, 24);
+
+/// The frame rate `RuntimeConfig::new` requires and the kernel never reads.
+fn unread_frame_rate() -> FrameRate {
+    FrameRate::new(NonZeroU32::new(60).expect("non-zero fps")).expect("60 FPS is valid")
+}
+
+/// Which of RFC 0014 §3.3's two quit routes a row exercises.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum QuitRoute {
+    /// `update` returns `Command::quit()`: applied synchronously at that
+    /// dispatch, no lane.
+    Sync,
+    /// `update` returns a command whose spawned run emits one quit on the
+    /// control lane — the route INV-L4's backlog independence carries to.
+    Control,
+}
+
+/// The data lane's mode.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum Mode {
+    /// `data_lane_capacity` unset: unbounded lane.
+    Unbounded,
+    /// The RFC 0007 §5.1 capacity.
+    Bounded,
+}
+
+impl Mode {
+    fn config(self) -> RuntimeConfig {
+        let config = RuntimeConfig::new(unread_frame_rate());
+        match self {
+            Self::Unbounded => config,
+            Self::Bounded => config
+                .app_channel_capacity(NonZeroUsize::new(DATA_LANE_CAPACITY).expect("non-zero")),
+        }
+    }
+
+    const fn label(self) -> &'static str {
+        match self {
+            Self::Unbounded => "unbounded",
+            Self::Bounded => "bounded",
+        }
+    }
+}
+
+#[derive(Clone)]
+struct Cfg {
+    name: &'static str,
+    /// Messages per second per producer, or [`BURST`].
+    rate: u64,
+    /// Total flood messages across all producers.
+    total: u64,
+    update_cost: Duration,
+    render_cost: Duration,
+    /// Quit when this flood seq reaches `update`; `None` quits at `total`.
+    quit_at_seq: Option<u64>,
+    quit_route: QuitRoute,
+    producers: u32,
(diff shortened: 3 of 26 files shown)


akiomik:
Thanks — all four addressed (`c029766`..`b964484`).

**1. `kernel_load` smoke profile.** Same `--smoke` flag and same
construction as `runtime_load`: rows derived from the canonical
tables by name lookup plus struct update, with the lookup panicking
on an unknown name so a rename can't leave the profile pointing at a
stale literal. The gates are completion and sequence integrity — the
draining rows must deliver their exact `0..total`, since a total-only
check passes a drop paired with a duplicate — and no latency is
asserted anywhere.

Both quit routes are in the profile. The route split is what
distinguishes this harness from the old one, and a sync-only smoke
would leave the control lane — the route INV-L4's property carries
to — uncompiled and undriven.

`just bench-smoke` now runs both harnesses (~6s locally);
`kernel_scan` is criterion, so it joins `subscription` in criterion's
own test mode. Neither new bench was executed by CI before this,
which was the real gap. The header now separates the two run modes:
the full run is the only one whose numbers mean anything, because a
shortened row means a shallower backlog and the backlog is exactly
the term the quit-to-applied interval scales with — a smoke
percentile is a different measurement, not a noisier one.

**2. Exhaustion diagnostic.** `QuiescenceGate` carries an `exhausted`
flag set when the hold spends its budget instead of being released,
and the window row reads it at both points where the window must be
open. Previously an exhausted hold was indistinguishable from a run
that quiesced early, and the failure surfaced several lines later on
an admission count that named neither cause — two failures with
opposite fixes. Not panicking inside the drop is unchanged; that
would abort the process rather than fail the test.

**3. Forwarder-panic re-admission — measured first.**

| point | `admissions()` | `evaluations()` |
| --- | --- | --- |
| after boot | 1 | 1 |
| after the exit pass | 1 | 1 |
| next re-evaluation, declaration removed | 1 | 2 |
| next re-evaluation, declaration kept | 2 | 2 |

Both halves match the contract. *No re-admit at the exit pass*: a
panicked run's quiescence is a finish, not a stopped run's, so it
marks no dirt (RFC 0014 §5.2) and that pass runs no re-evaluation at
all — there is no admission site for a restart to happen at, which
is what RFC 0005 INV-13 says by placing the restart at the next
re-evaluation. *Re-admit at the next re-evaluation when still
declared*: confirmed at 2, the restarted run panics again and is
contained again (RFC 0011 INV-LC8), the loop never terminates, and
the surviving producer's output is delivered across both panics.

Both are pinned now, and the restart behaviour gets its own row
(`a_still_declared_source_restarts_after_its_forwarder_panicked`)
rather than staying prose. No defect: the kernel has no notion of a
run that panicked "too recently to retry", and inventing one would
be a policy this crate hasn't specified.

**4.** (a) `GAUGE_PARTITION_SEEN` is cleared at `await_quiescence()`,
which is also where clearing becomes sound — past that barrier the
observer that could emit for those partitions went with its kernel,
and the gauge sum reaching zero was that partition's last event.
`runtime_load.rs` is deliberately untouched as the before column.
(b) §9.1's partition sentence is split into a lead, the two groups
in parallel with their owners, and the conclusion; no contract text
changed. `leak_name` left as-is per your note.

Gates: full suite green (714 lib), clippy clean on both builds, fmt,
`cargo doc -D warnings`, `just bench-smoke` run locally, touched
rows repeated ×10.

akiomik:
Both addressed (`0b501af`, `ba7dc3e`).

**A — attempt cap.** Derived per row now, `10 × trials`, the form §6
states normatively; the constant is gone, since one number can no
longer describe a per-row bound. The cap still binds an `Always`
row, which is where this harness departs from the old one on
purpose: that predicate never misses, but an attempt here can yield
no sample for other reasons — a timed-out kernel, a trial whose quit
instant was never recorded — and those are exactly the failures that
would otherwise retry without end.

Verified both directions: the smoke rows collect 5/5 in 5 attempts
each, and with a predicate that can never hold the row stops at
**50** attempts and fails the run (`0 valid / 50 attempts` →
`smoke profile failed`, non-zero exit). Previously that would have
burned 800.

**B — ownership.** §6 now attributes the second profile:
`just bench-smoke` carries two, and the successor harness's is this
section's form applied rather than a second definition of it — same
`--smoke` argument, same split of assertion classes, no latency
comparison anywhere, same trial-scaled cap. It names the two
departures and where they come from: quit rows run once per quit
route because RFC 0014 §3.3 splits a quit that used to be one thing,
and the cap binds every row rather than the predicate rows alone.
The singular readings are widened to match. RFC 0006 §5.3 untouched
— it delegates the definition here, and that direction is right.

Checklist for the touched range: item 4 passes (each of the eight
claims cross-checked against the bench source, including the
`Always => u32::MAX` count differing between the two harnesses),
item 7 passes (`RFC 0014 §13.5` and `§3.3` both resolve; no
revision-history markers).

Gates: full suite green, clippy both builds, fmt,
`cargo doc -D warnings`, `just bench-smoke` run locally.

akiomik:
Fixed in `156d786` with your wording: "one row per quit route rather than each row twice, the two routes RFC 0014 §3.3 separates" — the smoke assigns its two rows across the two routes; only the full profile runs each row twice.

akiomik:
All four addressed. This round changed the acceptance's structure, so
the arc is worth laying out; the head is `e552730`.

**1. Quit snapshots at the request site.** `depth_at_quit`,
`blocked_at_quit`, and the instant are now taken as one act at each
route's own request point — inside the `producer_quit` callback for
the control route, and after the update's spin for the synchronous
one (whose old pair spanned the row's own 25 µs update cost).
`BlockedEq` validity selects on the value at the request instant.

**2. The third differential, everywhere.** The pre-pass turn is now
stated in RFC 0014 §7.2 (as item iv of the enumeration) and its
summary, RFC 0008 §1.3, and both source module docs; the sweep found
six stale "exhaustively two seams" sites and three deliberate
survivors (RFC 0014 §10 counts seams, and two comparisons where the
turn is common to both sides), recorded so the next sweep does not
reopen them.

**3 & 4. The acceptance is now calibration-then-validation, and the
validating run is fresh and isolated.** What happened in between is
the story:

- Sessions 1 and 2 are declared **calibration**. The producer gate
  became explicit arithmetic — `p99 ≤ B(row) = (remainder × 25 µs +
  0.50 ms render + 0.25 ms hop + residual(depth)) × 1.25` — with
  every constant pinned numerically before the validating run, and
  the sync gate kept its 2.5× multiplier over the p50 predictor.
- Session 2's re-run under the fixed snapshots **failed one row**:
  `quit_blocked_64_sync`, p99 0.111 vs gate 0.070. Per the
  discipline written into the section, the gate was not touched. A
  blocked-producer sweep (N = 1…128, now an opt-in probe series)
  confirmed a real O(N) term — 92.5 ns/producer to issue the aborts,
  1,070 ns/producer to drain the joins, R² ≥ 0.97 — but at N=64 that
  is ~6 µs, an order short of the 81 µs excess, so the hypothesis
  was rejected as the explanation. What the probe did expose is that
  this environment's p99 has a floor: identical configurations
  reproduce medians to microseconds and differ up to 40× at p99.
- The sync gate therefore gained a floor —
  `max(2.5 × predicted p50, 0.15 ms)` — derived from the quiet
  runs' largest observed tail (95 µs, at the probe's N=128, not
  from the failing row, whose quiet calibration figure of 54 µs set
  nothing) with a 1.5× one-sided margin. The floor binds only below
  depth ≈4,900; every depth-carrying row is gated exactly as
  before. This was committed **before** the validating run.
- **Session 3** ran from that commit under recorded isolation: a
  prebuilt binary invoked directly, no `cargo`/`rustc` at either
  end of the 5 min 50 s window, worktree clean, evidence in the
  record. **All sixteen rows pass**, producer margins +0.41 to
  +1.43 ms, sync margins +0.037 to +3.12 ms; the delivery half
  stays flat across a 293-fold depth range (3.49–3.75 ms, largest
  at the smallest depth).
- One honest wrinkle, stated in the RFC rather than glossed:
  `quit_blocked_64_sync`'s tail (0.113 ms) **reproduced under
  isolation**, so it is a property of the configuration, not of the
  environment noise — its `applied→exit` matches the probe's join
  term (64 × 1,070 ns ≈ 68 µs). It passes on the floor, which it
  did not set, and the text says what that means: below the floor
  this environment cannot measure a tail, and the gate admits the
  row without claiming to have measured it.

RFC 0014 §13.5 now records the satisfying run. The reproducibility
rule gained the isolation requirement (an acceptance run is the
machine's only substantial load; a run with company is calibration
at best), which session 2's overlap with concurrent commits
motivated and session 3 observed.

Gates across the round: full suite green (714 lib), clippy both
builds, fmt, `cargo doc -D warnings`, smoke run locally; the probe
rows are opt-in by name and absent from a bare run.

akiomik:
Follow-up on the three findings from the last round, all landed
(`d920b2f`, `ef0cd96`), plus the validation run they set up.

**P1-1(a) — depth estimator.** `residual(depth)` now names **p50** as
the estimator, which is the one the numeric B values were computed
from when the section pinned them; the verdict table gains a depth
column so the arithmetic is reproducible from the page. It changes
nothing: only `quit_overload_control` has p50 and max far enough
apart to show, and B moves 0.003 ms there against a +1.073 ms margin.

**P1-1(b) — H's derivation.** You were right that the arithmetic was
false — 0.216 doubled is 0.432, not 0.25. H is pre-registered so the
value stays; the description now says what it is, a round figure
about 1.16× above the largest measured hop. All eight B values were
recomputed to confirm only the prose was wrong.

**P1-2 — backlog independence.** Also correct: "flat" was a judgement
made after seeing three runs. It now has two numbers over the six
non-idle rows — least-squares slope of delivery-half p50 against
depth ≤ +0.5 ns/envelope (under a sixteenth of the 8.4 ns residual
slope the model already carries) and span ≤ 0.6 ms (≈2.3× the widest
the calibration produced). Both one-sided, both pinned before the
validating run. That run measures **−0.0101 ns/envelope** and
**0.110 ms**, so both hold with wide margin, and the condition is
established on a criterion fixed beforehand.

**P2 — dates** split: calibration 2026-08-19, acceptance 2026-08-20.

Two things worth surfacing from the validating run rather than
leaving in the record. It began at load average 8.23, so it is not a
second acceptance run under §5.1's environment rule; it is usable
for this oracle because the oracle reads p50, and p50 held while the
tails did not — including a `quit_overload_control` p99 that would
miss its bound. That row's acceptance is not reopened, since a run
failing the environment rule cannot retract one any more than it
could grant one, and the section says so. Separately,
`quit_blocked_64_sync` now has three samples (0.111 / 0.113 /
0.092 ms) across rising load: the value recurs but does not track
load, so the earlier wording calling it "reproduced under isolation"
overstated what two points showed. Corrected; the floor argument
does not depend on it.

RFC 0014 §13.5's pending clause is gone — INV-L4's acceptance holds
on both its per-row gates and its backlog-independence oracle.

akiomik:
Taking the No-Go in order, because the sequence is the substance.

**The finding was right.** `d920b2f` pre-registered the validating
run as one made under §5.1's environment rule. The run that followed
began at load average 8.23 and did not meet it, and `ef0cd96`
admitted it anyway on the ground that the oracle reads p50 and p50
had held — a justification drawn from the very run the rule
excluded, applied after the results were visible. That is a post-hoc
evidence-class change, which is the one thing pre-registration
exists to prevent.

**`df354db` reverted it.** Backlog independence went back to
pending, that run was demoted to calibration, and the section names
the reasoning as the mistake it was rather than quietly dropping it.
The two constants were explicitly *not* re-fitted to include it —
they were pinned before the acceptance run and widening their basis
afterwards would be the same error pointing the other way.

**`4829664` validates it on an admissible run.** The new run's
eligibility is settled before any of its numbers are read, and from
facts external to its measurements: no `cargo`/`rustc` at either end
of the window, every concurrent working session halted before launch
with none polling inside it, and a closing load average of 1.73 —
over 8 min 22 s, a closing one-minute average describes the window
where the opening figure describes the minute before it. None of
those would read differently had every row come out otherwise, which
is exactly what the demoted run's argument could not say for itself.

Both conditions then hold: slope **−0.0806 ns/envelope** against
+0.5, span **0.113 ms** against 0.6, over the same six rows, oracle
and constants unchanged throughout. INV-L4's backlog independence is
established, and §13.5 now names the run carrying each of the two
conditions rather than one blanket claim.

One correction rides along: `quit_blocked_64_sync` has four samples
now, and ordered by how quiet the window was they read
0.111 / 0.113 / 0.092 / 0.084 ms. The tail depends on the
environment and shrinks on the quiet side, so the acceptance figure
is the largest of the four rather than a representative one — two
samples that agree establish nothing about independence. The floor
argument is unaffected; that row passes because F sits above it.

akiomik:
The No-Go was right, and the fix is procedural rather than
rhetorical.

**Two runs had been argued eligible after the fact and neither
was.** One's case came from its own output; the other's came from a
closing load average, which describes the last minute and cannot
reach back to a window that opened with one process holding 68.2% of
the machine. Both are now calibration, and backlog independence went
back to pending each time.

**`fca096f` stopped relying on the argument.** An acceptance run now
confirms quiet *before* it records anything, and the harness
performs the check: no `cargo`/`rustc`, one-minute load ≤ 2.5, no
single working process above 20% CPU, at most ten minutes waiting
for the three together. Thresholds come off the calibration runs —
2.5 has been observed at 1.73 and 2.14 with the bench alone and
separates those from the 7.87 and 8.23 that opened the failures; 20%
sits just under the quietest observed maximum of 19.5%; ten minutes
bounds a decay that took under nine. A barrier that does not clear
means the run does not exist: non-zero exit, nothing recorded.

**It then refused a run, which is the part worth reporting.**
Reaching a clean session took three attempts — one that met the
barrier but could not start recording, fixed at the source
(`67ca8ba`: the harness now flushes each finished row itself rather
than depending on how it was launched); one that did not meet it and
exited having measured nothing; one that met it at its ninth probe
and completed.

**`7077321` records the verdict.** Slope +0.1013 ns/envelope against
+0.5, span 0.102 ms against 0.6, oracle and constants exactly as
pinned before the acceptance run. INV-L4's backlog independence is
established. The slope is positive for the first time and that is
recorded rather than explained — the bound is signed, and the
magnitude is about an eighty-third of the residual term the model
already carries. Per-row figures ride along as confirmation with the
acceptance source unchanged; fourteen of sixteen come in at or below
it, and both that do not are well inside their own bounds.

On P2: the `quit_blocked_64_sync` samples are now stated as an
observation, not a mechanism. Five uncontrolled runs show the value
varying by about 2× with the acceptance figure the largest of them;
calling the environment the cause would take a controlled experiment
none of the five was. The floor argument is separate and unaffected.

akiomik:
The No-Go was right, and closing it took a rule change plus five more
attempts.

**The requirement moved from the window's start to the whole
window.** A pre-flight check says nothing about the minutes after it,
and the barrier's own logs showed a window climbing from 2.37 to 6.95
inside ten minutes. `891d682` made isolation two-stage and put both
stages in the harness rather than a launcher, since a launcher can
only observe around the measurement.

**The first in-window rule was unattainable, and nine runs proved
it.** Voiding on a single sample above 20% CPU, nine attempts between
22:15 and 00:52 all ended without data — six voided, three refused at
pre-flight. Every void was the 20% condition, never the load ceiling,
and no `cargo` or `rustc` process appeared in any sample of any
attempt; the load averages at the violating samples ran 1.42–2.25, an
idle machine touched by a daemon burst.

**`00ad9e8` revised it, before any validating run existed.** The 20%
condition now needs three consecutive samples, with anything above
100% voiding on sight, and every sample is recorded whether or not it
voids. The timing is the point: every attempt had ended in a void or
a refusal, so there was no result the change could have been steered
toward — unlike the two revisions this section already records as
post-hoc. The evidence's limit is recorded too: the harness stops at
the first violation, so those logs show the threshold was reached,
not that it was reached briefly, which is also why all samples must
now be kept.

**The revision reached only the stage it was meant to.** The next two
attempts were still refused at pre-flight, which the persistence
requirement does not touch — a genuinely occupied machine was turned
away while a daemon's two-sample burst no longer discards a good
window.

**`2642e74` records the verdict.** The validating run cleared
pre-flight at its first probe and held all 63 in-window samples, with
two non-voiding bursts logged. Slope −0.1263 ns/envelope against
+0.5, span 0.096 ms against 0.6, oracle and constants exactly as
pinned before the acceptance run. INV-L4's backlog independence is
established, and §13.5 now records both conditions as met.

Two things worth flagging from the checking rather than the running.
The slope's sign is not stable — −0.426, −0.081, +0.101, −0.126
across four sessions — and that is recorded as an observation, with
the note that a directional criterion would have failed twice where a
signed bound survives. And the calibration table had one slope as
−0.421 where the delivery halves give −0.426, from a transcribed row
value; corrected, with the constants unaffected since the bound came
from the residual slope rather than from those three numbers.

akiomik:
The No-Go was right on all three counts, and the third had decided
the previous verdict.

**`018689c` closes the three paths in one decision.** The monitor's
verdict now comes from a pure function that settles the immediate
conditions first — a build process in any sample, load above the hard
ceiling, a single reading above 100% — and consults the persistence
streak only when none of those hold. The old order returned a warning
from the streak branch with the void conditions already established,
so `cargo` plus a short burst read as elevated; reintroducing that
order now fails a self-test rather than a review. An unreadable load
average returns infinity instead of NaN, which fails the comparison
closed instead of open. And the window's end is synchronized: first
sample at t=0, a final sample taken after the stop signal and before
the verdict, and zero samples is an error rather than a window held
over an empty range. Every sample line now carries every condition's
input — load, build-process presence, top process — so the raw record
answers "was a compiler running at that burst?" by itself. The
negative tests sit behind a self-test flag wired into the smoke
target, so they gate CI.

**`d8c6078` demotes the run those paths had vouched for.**
Symmetrically with the earlier demotions: its record cannot establish
that no build process was present at the two bursts it does report,
and it declared its window held without reading the window's end.
Both conditions went back to pending. The same commit fixes the four
stale restatements — the slope list carries all seven calibration
values, the applied-to-exit count reads sixteen series, the sample
list for the worst row is complete with its unsupported quietness
ordering dropped, and the 20% threshold's direction is stated
correctly against the observed 19.5%.

**The validating run is the sixth attempt, and the five before it
are the fix visible in the field.** Each revised clause fired: one
void line lists two conditions together, which the swallowed path
would have collapsed to a warning; one void required three
consecutive samples where a single burst no longer suffices; one was
a single reading above 100%, voided on sight; one never cleared
pre-flight. The sixth cleared pre-flight at its first probe and held
all 61 samples with none voided — every one showing no build process
— with two sub-threshold bursts logged, and the held declaration
following a final sample taken after the last row.

**`90292c8` records the verdict.** Slope −0.1263 ns/envelope against
+0.5, span 0.095 ms against 0.6, oracle and constants exactly as
pinned before any acceptance run. INV-L4's backlog independence is
established on a window whose monitor could not have failed open in
the three ways this round named. Per-row figures ride along as
confirmation with their source unchanged; fourteen of sixteen come in
at or below it, and the two increases are one to three units of
reporting resolution.

One audit note from the checking rather than the running: printed at
four decimals, this slope collides with one calibration run's. The
underlying vectors differ in five of six components; the depth term
the slope divides by is about 7×10¹⁰ and identical across runs, so a
0.063 difference in the numerator compresses to four orders below the
printed precision. The bound is signed and absolute, so nothing turns
on the collision — but a criterion that ever compared slopes across
runs would need more digits than the display carries.

akiomik:
The No-Go was right: the window's opening was being assumed, not
observed.

**`9f10899` closes all three.** The start call now takes its first
sample synchronously on the calling thread, so by the time it
returns the window is open and observed — a delayed sampler can no
longer produce its only sample after the measurement it was meant to
guard. The sampler is injectable, so the guarantee is tested rather
than trusted to a quiet machine. Every guarded invocation now exits
through a single close-window path — the combined-flags form
included — so no return can reach exit 0 without the final sample,
the join, and the held check; finishing was chosen over rejecting
the combination because the module header promises the flags are
independent, and a rejection would have contradicted it. And the
process listing is parsed by a pure function for which a non-zero
exit, empty output, or no parsable row is an observation failure,
not a quiet machine. Each fix was confirmed by reverting it: all
three mutations panic the self-test, which runs inside the smoke
target and therefore gates CI.

**`930f8b7` demotes the run the open boundary had vouched for.** Its
record shows the first sample preceding the first row report, which
looks like the ordering the condition needs — but a row reports only
after its two hundred trials complete, so "before the first report"
is a far weaker fact than "before measurement began", and the second
does not follow from it. Same standard as the demotions before it:
not a claim that the window was dirty, but that the record cannot
say. Both conditions went back to pending in the same commit.

**The validating run took eight attempts, and the seven refusals are
the instrument's liveness.** Four voided on sight at a single
reading above 100%; three voided on three consecutive elevated
samples — one of those catching the development tooling driving this
campaign, so the guard is not tuned to daemons only. The eighth
cleared pre-flight at its first probe and held all 61 samples with
none voided: load between 1.22 and 2.40, no build process in any
sample, nothing above 100%, and the longest elevated run two samples
against a bound of three — the persistence rule holding right at its
threshold. Both ends bracket the measurement: the first sample
completed before measuring began, by construction, and the held
declaration followed a final sample taken after the last row.

**`da5c70a` records the verdict.** Slope −0.1200 ns/envelope against
+0.5, span 0.102 ms against 0.6, oracle and constants exactly as
pinned before any acceptance run. Per-row figures ride along as
confirmation: every row inside its pre-registered bound on both
routes at 200 valid trials with no predicate misses; thirteen of
sixteen at or below the source run's p99, the three above it higher
by 5 to 16 µs.

The rounding-collision check ran pre-emptively this time. No new
four-decimal collision: this slope is distinct from every earlier
run's, the nearest pair sitting 0.006 ns/envelope apart. One span
value does coincide with an earlier calibration run's — checked, the
underlying six-component vectors differ, and a max-minus-min
statistic coinciding across distinct vectors is unremarkable. Both
checks are in the record so the earlier alarm does not repeat.

PR #301, opened 2026-08-26
ci(codecov): stop the project status failing on measurement noise

### LINKED ISSUE #299: codecov/project fails on commits that change no Rust: default threshold 0% meets a coverage measurement that oscillates
## Summary

There is no `codecov.yml` in the repo, so Codecov applies its default `project` status: `target: auto`, `threshold: 0%`. Any decrease at all fails the check.

The measurement is not stable to that resolution. On `docs/migration-guide-0-11` it lands on exactly two values for the same executable code, and `codecov/project` passes or fails accordingly.

## Evidence

Every commit below is on one branch against one base (`f64e44b`). The `.rs` column counts Rust files changed by that commit; the three that touch `.rs` change only a `cfg(doctest)` item and rustdoc comments, so no executable line differs anywhere in the range.

| commit | .rs changed | codecov/project |
| --- | --- | --- |
| 89c8a3b | 0 | **failure** 95.59% (-0.03%) |
| f878f57 | 0 | success 95.61% (+0.00%) |
| acbb974 | 0 | success 95.61% (+0.00%) |
| 7e5df44 | 1 (rustdoc only) | success 95.61% (+0.00%) |
| 80e9a7d | 0 | success 95.61% (+0.00%) |
| 8980d3c | 0 | **failure** 95.59% (-0.03%) |

`codecov/patch` reported "Coverage not affected" on every one of them, including both failures — the diffs introduce no uncovered lines, because they introduce no lines.

Re-running the `Code Coverage` job on a failing commit has flipped it to 95.61% (+0.00%) each time it was tried, with nothing else changed. PR #295, a `CHANGELOG.md`-only change, showed the same pattern once.

So a docs-only PR fails a check about half the time, and the only remedy is to re-run a job until the measurement lands on the higher of two values.

## Why

`threshold: 0%` means "fail on any decrease". Combined with a measurement that varies run to run — plausibly the async/kernel scheduling-dependent paths, where which branches execute depends on thread interleaving — false failures are structural rather than occasional.

## Options

Current operating policy is that this check is not required and its failures are judged by content. Both options below preserve that; they differ in what the check reports.

1. **Make it informational.** Matches the policy exactly: Codecov still posts the number and the PR comment, and never reports failure.

   ```yaml
   # codecov.yml
   coverage:
     status:
       project:
         default:
           informational: true
   ```

2. **Keep it a gate, with a threshold above the noise.** Observed spread is 0.02–0.03 points, so anything at or above 0.1% removes these while still catching a real drop.

   ```yaml
   coverage:
     status:
       project:
         default:
           threshold: 0.1%
   ```

Either way `codecov/patch` is unaffected and keeps flagging genuinely uncovered new lines, which is the part that has been accurate throughout.

## Notes

Worth confirming the oscillation's cause before picking a threshold, in case it points at a test whose coverage is interleaving-dependent — that would be worth knowing independently of this check. The two-value pattern above is the whole of what has been measured so far.

akiomik:
## What

Adds `codecov.yml`, setting `coverage.status.project.default.threshold: 0.5%`.

## Why

There is no `codecov.yml` in the repo, so Codecov applies its default `project` status — `target: auto`, `threshold: 0%` — which fails on any decrease at all. The measurement is not stable to that resolution.

On `docs/migration-guide-0-11`, six commits against one base landed on exactly two values for the same executable code:

| commit | .rs changed | codecov/project |
| --- | --- | --- |
| 89c8a3b | 0 | **failure** 95.59% (-0.03%) |
| f878f57 | 0 | success 95.61% (+0.00%) |
| acbb974 | 0 | success 95.61% (+0.00%) |
| 7e5df44 | 1 (rustdoc only) | success 95.61% (+0.00%) |
| 80e9a7d | 0 | success 95.61% (+0.00%) |
| 8980d3c | 0 | **failure** 95.59% (-0.03%) |

Re-running the `Code Coverage` job on a failing commit has flipped it to the higher value each time it was tried, with nothing else changed. So a docs-only pull request fails this check about half the time, and the only remedy is to re-run a job.

## Why a threshold and not `informational`

`informational: true` was the first thing this PR did, and it silenced more than the noise. `project` is the only signal watching code a diff does not touch: `patch` reports on the lines a diff **adds**, so a pull request that deletes a test covering something else adds none and passes. The `Code Coverage` job cannot cover for it either — it generates a report and uploads it, and nothing in it reads the number back or fails on it. With `project` informational, nothing anywhere would notice that change.

0.5% is set against the observed spread rather than against a target regression size: the oscillation is 0.02–0.03 points, so the tolerance sits more than an order of magnitude above the noise it exists to absorb, while a decrease worth acting on is not that small. Worth re-measuring if the oscillation's cause is ever found and fixed, since the reason for any tolerance at all goes with it.

## What this is and is not

Neither codecov status is required by the repository's ruleset — the required contexts are `Format Check`, `Clippy`, `Documentation`, `Code Coverage` and the three `Test (…, stable)` legs. So this is a red mark a reviewer judges by content, not a merge gate. That is the operating policy, and the policy needs something to judge, which is the argument against turning the signal off.

`patch` is left at its default for the same reason. On every commit in the range above it read "Coverage not affected" — which says only that those diffs added no lines for it to measure.

## Verification

```
$ curl -X POST --data-binary @codecov.yml https://codecov.io/validate
Valid!
{"coverage":{"status":{"project":{"default":{"threshold":0.5}}}}}
```

Whether the setting takes effect on this PR itself or only once it is on the default branch is not something I have verified; the `codecov/project` check on this PR is the observation.

Closes #299

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -8,20 +8,30 @@
 # the time, and the only remedy is to re-run a job until the measurement lands
 # on the higher value.
 #
-# `informational` keeps the number and the pull request comment and never
-# reports failure. That is what the operating policy already is — this check is
-# not required and its failures are judged by content — so the setting states
-# it rather than changing it.
+# A threshold rather than `informational`, which was the first thing tried
+# here. `informational` never reports failure, and that is more than this
+# needs: `project` is the only signal that watches *unchanged* code, because
+# `patch` reads the lines a diff adds and a pull request that deletes a test
+# covering something else adds none. Silencing it leaves nothing to notice
+# that, the `Code Coverage` job included — it uploads with
+# `fail_ci_if_error: false` and cannot fail on content.
 #
-# `patch` is left at its default, which makes it a report and not a gate:
-# the repository's ruleset requires neither codecov status, and the
-# `Code Coverage` job that feeds them uploads with `fail_ci_if_error: false`,
-# so it cannot fail on their content either. What `patch` keeps doing is
-# reporting on the lines a diff adds. On every commit in the range above it
-# read "Coverage not affected", which says only that those diffs added no
-# lines for it to measure.
+# 0.5% is chosen against the observed spread rather than against a target
+# regression size: the oscillation is 0.02–0.03 points, so this is more than
+# an order of magnitude above the noise it exists to absorb, while a real
+# decrease large enough to matter is not that small. Re-measure it if the
+# oscillation's cause is ever found and fixed, since the reason for any
+# tolerance at all goes with it.
+#
+# Neither codecov status is a merge gate, and that is deliberate rather than
+# an oversight: the repository's ruleset requires neither, so what a threshold
+# buys is a red mark a reviewer can judge by content — which is the operating
+# policy, and which needs something to judge. `patch` is left at its default
+# for the same reason, reporting on the lines a diff adds. On every commit in
+# the range above it read "Coverage not affected", which says only that those
+# diffs added no lines for it to measure.
 coverage:
   status:
     project:
       default:
-        informational: true
+        threshold: 0.5%
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,25 @@
+# Without this file Codecov applies its default `project` status, `target:
+# auto` with `threshold: 0%` — fail on any decrease at all. The measurement is
+# not stable to that resolution. On one branch, against one base, six commits
+# that changed no executable Rust landed on exactly two values, 95.61% and
+# 95.59%, and the check passed or failed accordingly; re-running the Code
+# Coverage job with nothing else changed has flipped a failure to a pass every
+# time it was tried. So a docs-only pull request fails this check about half
+# the time, and the only remedy is to re-run a job until the measurement lands
+# on the higher value.
+#
+# `informational` keeps the number and the pull request comment and never
+# reports failure. That is what the operating policy already is — this check is
+# not required and its failures are judged by content — so the setting states
+# it rather than changing it.
+#
+# `patch` is deliberately left at its default. It reports on the lines a diff
+# adds, so it has nothing to say about a diff that adds none: it read "Coverage
+# not affected" on every commit in the range above, including both failures.
+# Uncovered new lines are the part that has been accurate throughout, and it
+# stays a gate.
+coverage:
+  status:
+    project:
+      default:
+        informational: true
--- a/codecov.yml
+++ b/codecov.yml
@@ -13,11 +13,13 @@
 # not required and its failures are judged by content — so the setting states
 # it rather than changing it.
 #
-# `patch` is deliberately left at its default. It reports on the lines a diff
-# adds, so it has nothing to say about a diff that adds none: it read "Coverage
-# not affected" on every commit in the range above, including both failures.
-# Uncovered new lines are the part that has been accurate throughout, and it
-# stays a gate.
+# `patch` is left at its default, which makes it a report and not a gate:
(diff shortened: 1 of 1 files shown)


akiomik:
Both addressed. Head is 34b743e, 21/21 green, `mergeStateStatus=CLEAN`.

## Finding 1 — `informational` silenced more than the noise

Taken, and the config changed to `threshold: 0.5%`.

Your reasoning is the part I had missed when I proposed this: `project` is the only signal watching code a diff does not touch, because `patch` reads the lines a diff *adds*. A pull request that deletes a test covering something else adds none, so `patch` passes; `project` informational never fails; and the `Code Coverage` job uploads with `fail_ci_if_error: false`, so it cannot fail on content either. Nothing anywhere would have noticed.

0.5% rather than the 0.1% I originally offered as the alternative: the value is set against the observed spread (0.02–0.03 points), so it sits more than an order of magnitude above the noise it exists to absorb, while a decrease worth acting on is not that small. The comment says to re-measure it if the oscillation's cause is ever found, since the reason for any tolerance goes with it.

## Finding 2 — two sentences claimed more than their evidence

Both confirmed against the repository before changing anything:

```
required contexts: Format Check, Clippy, Documentation, Code Coverage,
                   Test (ubuntu|macos|windows-latest, stable)
coverage job:      codecov/codecov-action@v7 with fail_ci_if_error: false
```

So "it stays a gate" was false, and — as you say — sitting beside "this check is not required" it read as though `patch` were the one that is. The file now says plainly that neither status is a gate and why that is deliberate.

"the part that has been accurate throughout" is gone. The observation is stated as what it is: `patch` read "Coverage not affected" across the range, which says only that those diffs added no lines for it to measure.

The same two sentences were in the PR body; it has been rewritten around the threshold, with the required-context list quoted rather than characterised.

## Note on how this one arrived

The `informational` choice came from a question I put to the maintainer, and I did not surface the consequence you identified — that it removes the only signal for unchanged code. The alternatives I offered were `informational` and a 0.1% threshold, described only in terms of noise. That framing is why the weaker option was picked, and it was mine.

akiomik:
Both fixed in e9299fb. Head is green, 21/21, `mergeStateStatus=CLEAN`.

## 1 — the grounds did not reach the claim

You are right, and the flag does not do what I used it for. `fail_ci_if_error` decides whether an upload or processing *error* fails the job; it has nothing to say about the number. Setting it `true` would not make a coverage decrease fail the job either, since the action does not wait on the statuses at all — so the sentence was not merely under-supported, it was pointing at an unrelated mechanism.

Replaced with the actual reason: the job generates a report and uploads it, and nothing in it reads the number back. Same correction in the PR body, which no longer mentions the flag at all.

## 2 — "the first thing tried here"

Also right, and this one is a repeat for me rather than a one-off: a file that lives on the default branch cannot resolve "here", and revision history belongs in Git rather than in the artefact. Clause dropped; the paragraph after it already carried the argument on its own merits.

## On the branch name

Worth doing something about, since `Merge pull request #301 from akiomik/ci/codecov-project-informational` would sit on `main` describing the opposite of the change. `gh pr merge` takes `--subject`, so the merge commit can read accurately without renaming anything or losing this thread. I have raised it with the maintainer as a merge-time choice rather than deciding it here, since it deviates from the default subject every other merge on `main` uses.

## Both of these were the same class as the previous round

Two rounds running, the findings have been "the claim is fine, the grounds under it are not". That is a pattern in what I write rather than three separate slips, and I have recorded it as one.

PR #302, opened 2026-08-27
ci: run the toolchain each job names

### LINKED ISSUE #297: CI: rust-toolchain.toml overrides every dtolnay/rust-toolchain step, so the MSRV and beta jobs do not test what they name
## Summary

`rust-toolchain.toml` pins `channel = "1.97.0"`. `dtolnay/rust-toolchain` selects a toolchain with `rustup default <toolchain>` only — it sets neither `RUSTUP_TOOLCHAIN` nor a directory override. rustup's precedence is `RUSTUP_TOOLCHAIN` > directory override > `rust-toolchain.toml` > `rustup default`, so the file wins in every CI job.

## Evidence

```
$ rustup default
stable-aarch64-apple-darwin (default)

$ rustup show active-toolchain
1.97.0-aarch64-apple-darwin (overridden by '/…/tears/rust-toolchain.toml')
```

## Consequences

- **`msrv` (`.github/workflows/ci.yml`, "MSRV (1.88.0)")** runs `cargo check` under 1.97.0, not 1.88.0. `rust-version = "1.88.0"` in `Cargo.toml` is currently unverified — a 1.89+ feature could land without CI noticing.
- **`test` matrix `rust: [stable, beta]`** runs 1.97.0 for both legs. `beta` is never exercised, and 3 of the 6 matrix jobs duplicate the other 3.

## Possible fixes

Either set `RUSTUP_TOOLCHAIN` in the jobs that need a specific toolchain:

```yaml
- uses: dtolnay/rust-toolchain@master
  with:
    toolchain: 1.88.0
- run: cargo check --all-targets --all-features
  env:
    RUSTUP_TOOLCHAIN: 1.88.0
```

…or have those jobs remove/override `rust-toolchain.toml` before running, or add `rustup override set` steps. Whichever is chosen, it is worth asserting the active toolchain in the job so a future precedence change is caught:

```yaml
- run: rustc --version | grep -q '1\.88\.0'
```

## Notes

Pre-existing; found while reviewing #296, whose own premise is that a CI job should run what its name claims. Not fixed there because it changes what every job tests and deserves separate review.

akiomik:
## What

Every job in `ci.yml` (and the publish job in `release.yml`) now runs the toolchain it names.

- A `versions` job reads the pinned channel from `rust-toolchain.toml` and the MSRV from `Cargo.toml`'s `rust-version`, and every other job takes its toolchain from there instead of restating a number.
- `msrv` sets `RUSTUP_TOOLCHAIN` to `rust-version` and derives its own display name from it.
- The `test` matrix's first leg is renamed `stable` → `pinned` — which is what it has always run — and the `beta` leg sets `RUSTUP_TOOLCHAIN` so that it is beta.
- Both of those jobs assert the active toolchain.

## Why

`dtolnay/rust-toolchain` selects a toolchain with `rustup default` only. rustup's precedence is `RUSTUP_TOOLCHAIN` > directory override > `rust-toolchain.toml` > `rustup default`, so the pin won in every job whatever the job asked that action for:

```
$ rustup default
stable-aarch64-apple-darwin (default)

$ rustup show active-toolchain
1.97.0-aarch64-apple-darwin (overridden by '…/rust-toolchain.toml')

$ RUSTUP_TOOLCHAIN=1.88.0 rustup show active-toolchain
1.88.0-aarch64-apple-darwin (overridden by environment variable RUSTUP_TOOLCHAIN)
```

So `MSRV (1.88.0)` ran `cargo check` under 1.97.0 and `rust-version` was asserted by nothing; the `test` matrix ran the pin on both legs, so `beta` was never exercised and three of its six jobs repeated the other three. Two quieter consequences, both fixed by the same change: a job's `components` were added to the toolchain the action installed rather than the one it built with (the `coverage` job's `llvm-tools-preview`, the `fmt` job's `rustfmt`), and `Swatinem/rust-cache` keyed off that same unused toolchain.

Keeping the pin as what CI runs is deliberate. It is the toolchain a contributor's checkout selects, and `clippy`'s `-D warnings` is the one place where a lint added after the pin would fail a pull request that `just clippy` reports as clean.

## Verification

The two newly-real toolchains build:

```
$ RUSTUP_TOOLCHAIN=1.88.0 cargo check --all-targets --all-features
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 42s

$ RUSTUP_TOOLCHAIN=beta cargo check --all-targets --all-features     # 1.99.0-beta.1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10m 36s
```

beta reports three deprecation warnings (`Atomic::<u64>::fetch_update` is renamed `try_update`), which is the beta leg doing what it is for. Not acted on here: `try_update` does not exist on the MSRV, so it is a separate decision.

`actionlint` (with shellcheck) reports no errors on the changed workflows.

Not verified locally: that the full test suite *passes* on beta, only that it compiles. If the beta leg goes red on this PR that is a real finding rather than a defect in this change.

## Required status checks need updating with this

The repository ruleset lists `Test (ubuntu-latest, stable)`, `Test (macos-latest, stable)` and `Test (windows-latest, stable)` as required. Renaming the leg replaces those with `… , pinned)`, so the ruleset has to be updated or nothing merges — including this PR. The other open PRs should be merged or rebased around that change, since a PR built from the old workflow produces the old names.

Closes #297

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,15 @@ jobs:
   # the wrong one installed its components somewhere it never built and cached
   # under a key that did not describe the build.
   #
+  # One consequence worth recording, because it stays silent until it bites:
+  # the action installs with `--profile minimal`, and it is now installing the
+  # pin itself. Before, the pin arrived by rustup auto-installing it from
+  # `rust-toolchain.toml`, which carries `profile = "default"` and so brought
+  # rustfmt, clippy and rust-docs along unasked. A job gets exactly what its
+  # `components:` names now. Every job here asks for what it uses, but one
+  # added later that runs, say, `just check` would fail on a component that
+  # used to be present by accident.
+  #
   # Reading both values out of the files that declare them is what keeps a job
   # honest: the pin has one home (`rust-toolchain.toml`) and the MSRV has one
   # home (`Cargo.toml`'s `rust-version`), and bumping either moves CI with it
@@ -106,12 +115,21 @@ jobs:
           toolchain: ${{ env.RUSTUP_TOOLCHAIN }}
       - name: Verify the active toolchain
         shell: bash
-        # A leg is only what it claims while that precedence holds. Asserting it
-        # is what turns a future change in it into a failure here rather than
-        # into a leg that quietly tests the same thing as its neighbour.
+        # A leg is only what it claims while that precedence holds, so this
+        # turns a future change in it into a failure rather than into a leg
+        # that quietly tests the same thing as its neighbour. Note what it can
+        # and cannot catch: on the `beta` leg the two candidate answers differ,
+        # so the check is the precedence claim itself; on the `pinned` leg
+        # `RUSTUP_TOOLCHAIN` and `rust-toolchain.toml` carry the same string by
+        # construction, so it passes whichever wins and is a consistency check
+        # rather than evidence.
+        #
+        # A prefix comparison rather than `grep`, whose pattern would read the
+        # dots in a version as any-character.
         run: |
-          rustup show active-toolchain
-          rustup show active-toolchain | grep -q "^${RUSTUP_TOOLCHAIN}-"
+          active=$(rustup show active-toolchain)
+          echo "$active"
+          [[ $active == "${RUSTUP_TOOLCHAIN}-"* ]]
       - uses: Swatinem/rust-cache@v2
       - name: Run tests
         # Benches are excluded from this matrix (and run once, on Linux
@@ -232,11 +250,15 @@ jobs:
       - name: Run doc tests against the packaged crate
         run: just test-doc-packaged
 
-  # The version is `Cargo.toml`'s `rust-version` rather than a number written
-  # here, so the job's name, the toolchain it installs and the claim it is
-  # checking cannot come apart.
+  # The version installed is `Cargo.toml`'s `rust-version` rather than a number
+  # written here, so the toolchain this checks and the claim it is checking
+  # cannot come apart. The *name* is deliberately static and carries no
+  # version: a check-run name is what a required-status-check rule matches on,
+  # so deriving it from a file would let a `rust-version` bump rename the check
+  # with no diff here — which is the same trap the `pinned` leg above avoids by
+  # not being called `1.97.0`. The version is printed in the job instead.
   msrv:
-    name: MSRV (${{ needs.versions.outputs.msrv }})
+    name: MSRV
     runs-on: ubuntu-latest
     needs: versions
     # Without this the pin outranks the toolchain installed below and this job
@@ -250,8 +272,9 @@ jobs:
           toolchain: ${{ env.RUSTUP_TOOLCHAIN }}
       - name: Verify the active toolchain
         run: |
-          rustup show active-toolchain
-          rustup show active-toolchain | grep -q "^${RUSTUP_TOOLCHAIN}-"
+          active=$(rustup show active-toolchain)
+          echo "$active"
+          [[ $active == "${RUSTUP_TOOLCHAIN}-"* ]]
       - uses: Swatinem/rust-cache@v2
       - name: Check MSRV
         run: cargo check --all-targets --all-features
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,13 +14,50 @@ env:
   RUST_BACKTRACE: 1
 
 jobs:
+  # Every job below names the toolchain it runs, and gets it from here rather
+  # than restating it.
+  #
+  # Why the indirection: `dtolnay/rust-toolchain` selects a toolchain with
+  # `rustup default` only, and rustup's precedence is `RUSTUP_TOOLCHAIN` >
+  # directory override > `rust-toolchain.toml` > `rustup default`. So the
+  # pinned channel outranks whatever a job asks that action for, and asking
+  # for `stable` while running the pin is the state this indirection replaces:
+  # the toolchain the action installs is the one the job's components are
+  # added to and the one `Swatinem/rust-cache` keys off, so a job that named
+  # the wrong one installed its components somewhere it never built and cached
+  # under a key that did not describe the build.
+  #
+  # Reading both values out of the files that declare them is what keeps a job
+  # honest: the pin has one home (`rust-toolchain.toml`) and the MSRV has one
+  # home (`Cargo.toml`'s `rust-version`), and bumping either moves CI with it
+  # instead of leaving a job checking a version nothing claims any more.
+  versions:
+    name: Declared Versions
+    runs-on: ubuntu-latest
+    outputs:
+      pinned: ${{ steps.read.outputs.pinned }}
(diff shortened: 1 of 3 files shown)


akiomik:
Findings 2–5 and 7 are fixed in e33f7a8; 6 is filed as #308; 1 is the merge-window action and is unchanged. 22/22 green, and the check list now reads `MSRV` rather than `MSRV (1.88.0)`.

## 2 — the MSRV job name was this branch's own trap

Taken, and the inconsistency you point at is the substance of it: fifteen lines earlier this branch picks `pinned` over `1.97.0` *precisely* so a bump cannot rename a check, and then derives the MSRV check name from `Cargo.toml`. Two judgements pointing opposite ways in one file. `name: MSRV` is static now and the version is printed in the job.

I did not take the second half — collapsing `versions` back into per-job steps. It is true that the dynamic name was the only thing forcing a job rather than a step, but inlining puts the `sed` in ten jobs, and #308 (filed from your finding 6) is about that snippet already existing in two places and drifting. Ten seconds of serialisation looks like the cheaper side of that trade. Happy to be told otherwise.

## 3 — the assertion is evidence on two legs, not six

Correct, and it is the "grounds do not reach the claim" shape again. On the `pinned` leg the two candidate answers are the same string by construction, so the check passes whichever wins. The comment now says which legs carry the precedence claim (`beta`, `msrv`) and which is a consistency check.

## 4 — dots as wildcards

Real, and confirmed before changing it:

```
$ echo "1x97y0-aarch64" | grep -q "^1.97.0-" && echo match
match
$ [[ "1x97y0-aarch64" == "1.97.0-"* ]] && echo match || echo "no match"
no match
```

Prefix comparison at both sites, and it still rejects a genuinely wrong toolchain (checked with `RUSTUP_TOOLCHAIN=beta` against `1.97.0-`).

## 5 — the pin now arrives `--profile minimal`

Recorded in the `versions` comment block, framed as you put it: every job currently asks for what it uses, but that stopped being free, and a job added later that runs `just check` would fail on a component that used to be present by accident.

## 6 — declined here, filed as #308

Under the stopping rule this branch is now operating by, it is latent — `test -n` catches the likely extraction failure and `--dry-run` runs first in the same job — and fixing it either adds an assertion to a workflow this branch only touches in passing or introduces a composite action. Both are a different shape from what this PR claims. #308 has the reasoning and three options.

## 7 — stale doc line

Fixed; `docs/testing.md` now says "separate from the test matrix".

## 1 — the merge window

Unchanged and still the blocker: `mergeStateStatus: BLOCKED`, because the ruleset requires `Test (…, stable)` and this PR posts `Test (…, pinned)`. I have the maintainer's agreement to make that edit and have asked for the go-ahead, flagging that it also blocks #293 until that PR is rebased onto a `main` containing this one — #293's base is `f64e44b`, so it needs a rebase regardless.

akiomik:
The high finding is correct, it is mine, and it is closed. `Declared Versions` is now a required context.

## Verified before acting

```
needs: versions in this PR      10   (fmt clippy test bench loom api-surface doc doctest msrv coverage)
needs: anywhere on main          0
Declared Versions required?      no
```

So all seven required contexts sit downstream of one job that nothing required. `versions` fails → ten jobs skipped → skipped counts as success for a required check → a pull request merges with nothing having compiled. Fail-open, introduced by this branch, and my own reasoning for the `versions` job never asked what happens when it fails.

## Fixed in the ruleset, in the same window

```
Format Check
Clippy
Documentation
Code Coverage
Test (ubuntu-latest, pinned)
Test (macos-latest, pinned)
Test (windows-latest, pinned)
Declared Versions          <- added
```

Both edits verified by diffing the whole ruleset before and after: `enforcement`, `conditions`, `bypass_actors` and all four rule types identical, only the contexts changed. Backups of both states kept.

Now `versions` failing is a failed required check rather than seven skipped ones, so the merge is blocked with the cause named.

I agree with not collapsing `versions` into per-job steps, and for the reason you give rather than the one I gave earlier: `env: RUSTUP_TOOLCHAIN` at job level cannot read a step output, so it would mean ten `$GITHUB_ENV` writes — more duplication than the `sed` it removes, and #308 is already about that snippet existing twice.

## nit

Taken: `docs/testing.md` now says "in its own job, `api-surface`". You are right that it matters more than usual here, since a check-run name *is* a job name and that is what this whole branch is about.

## What this one was

Not a missed edge case — a structure I added without asking what happens when the new dependency fails. The blast radius of `needs:` was the whole question and I did not ask it. Recorded that way: a gating job must be added to the required contexts in the same change that other jobs start depending on it.

22/22 green, `mergeStateStatus: CLEAN`.

akiomik:
Fixed here rather than deferred, in d643dee — because this branch is what opens the path, and because it has already happened on this branch once.

## It is not hypothetical

Run 33041812710, attempt 1, before I re-ran the failed jobs:

```
failure    Test (macos-latest, beta)      <- actions/checkout failed to load on that runner
cancelled  Test (macos-latest, pinned)
cancelled  Test (ubuntu-latest, beta)
cancelled  Test (ubuntu-latest, pinned)
cancelled  Test (windows-latest, beta)
cancelled  Test (windows-latest, pinned)
```

One leg failed for a reason that had nothing to do with Rust and took the other five with it, all three now-required `pinned` legs included. (The current API view of that run shows six successes — it reports the latest attempt, and I re-ran the failed jobs at the time. `…/attempts/1/jobs` is where the above comes from.)

That failure was infrastructure rather than an upstream `beta` regression, so it is not an instance of your scenario — but it exercises the same mechanism, and it is the mechanism that decides the blast radius.

## Why in this PR

Your framing is right that the default is pre-existing, and it is also right that it was harmless until now: while both legs ran the pin, the only failures possible were ones every leg shared, so nothing could take out a leg its neighbours would have survived. This branch makes `beta` able to fail alone. The exposure is created here, so the one line belongs here.

`cancelled` blocks rather than passes, so this was never fail-open the way the unrequired gating job was — the cost is operational, not correctness. Worth saying in the comment beside it, which now does.

22/22 green, `mergeStateStatus: CLEAN`.

akiomik:
Correction: two commits I reported as landing on this PR were never on it, and it merged without them.

This PR merged as `e33f7a8`. It does **not** contain:

- `fail-fast: false` on the test matrix — my comment said "fixed here rather than deferred, in d643dee"
- the `docs/testing.md` line naming the `api-surface` job — my comment said "taken"

Both are real commits; they were made on a detached `HEAD`. `git push origin ci/pin-toolchain-per-job` then pushed the branch ref, which had not moved, so it was a no-op and exited zero. I read the local `HEAD` back and took that as confirmation the remote had it. The CI run I then quoted as "22/22 green" was real, but it was green for `e33f7a8` — the content actually on the PR, not the content I said was on it.

Both are now on #309, unchanged from what was reviewed here.

What landed in this merge is everything through `e33f7a8`: the toolchain-per-job change, the MSRV check-run name, the prefix comparisons, the profile note, the ruleset edits (`pinned` contexts plus `Declared Versions`). The two ruleset edits were applied directly and are unaffected.

PR #303, opened 2026-08-27
ci: name the feature set instead of asking for all of them

### LINKED ISSUE #298: CI: --all-features hides subscription::signal, so its unit tests run on no OS and no toolchain
## Summary

`src/subscription.rs` gates the module:

```rust
#[cfg(not(all(feature = "loom-core", test)))]
pub mod signal;
```

`--all-features` enables `loom-core`. Under `cargo test --lib --all-features` the condition holds and the whole module is compiled out before its tests are collected, so `subscription::signal`'s unit tests execute on **no** OS and **no** toolchain in CI — the `test` job runs `cargo test --lib --bins --tests --examples --all-features`.

## Evidence

```
$ cargo test --lib                | grep -c "signal::"
3
$ cargo test --lib --all-features | grep -c "signal::"
0
```

## Doctests are affected too, by a mechanism that is *not* established

Enabling `loom-core` also removes `subscription::signal`'s two doctests:

```
$ cargo test --doc                        # 61 tests, 2 signal doctests
$ cargo test --doc --features loom-core   # 60 tests, 0 signal doctests
$ cargo test --doc --all-features         # 71 tests, 0 signal doctests
```

Diffing the two test-name lists shows exactly the two `signal` doctests dropped and nothing else.

**The obvious explanation does not hold.** For `--lib` the `test` in the cfg is genuinely set, which explains that case. For doctests it is not: dumping every `--cfg` token passed to every rustdoc invocation in the run yields only

```
'feature="default"'
'feature="loom-core"'
```

with no `--cfg test` in any of the three invocations. So the effect is reproducible but its cause is unidentified. Worth pinning down before relying on any model of it.

## Why the gate exists

`loom-core` pulls in loom, and `--cfg loom` reconfigures tokio, dropping modules like `signal` — the `loom` job's own comment records this. The gate is correct; using `--all-features` to mean "everything a user can enable" is what is wrong, since `loom-core` and `bench-internals` are both documented as build-only and not part of the public API.

## Possible fix

Name the user-facing feature set instead of `--all-features`:

```
cargo test --lib --bins --tests --examples --features http,ws,native-tls
```

That covers strictly more than either `--all-features` or default features. `loom-core` already has its own dedicated `loom` job, and `bench-internals` is covered by the `bench` job, so nothing loses coverage. (#296 applies exactly this to `just test-doc`, where it takes the doctest count from 61/71 to 73.)

Worth checking the other `--all-features` call sites at the same time (`clippy`, `msrv`, `coverage`, `doc`): for the lint/check jobs the gate's effect is only that `signal` is not linted, but that is the same blind spot.

## Notes

Pre-existing; found while reviewing #296. Not fixed there because it changes what the main test matrix compiles.

akiomik:
> Stacked on #302 — the base is `ci/pin-toolchain-per-job`, since both change the same jobs. Retarget to `main` once #302 merges.

## What

CI and the `justfile` stop passing `--all-features` and name the feature set instead.

- `justfile`: `doc_features` is renamed `user_features` (every feature a dependant can name), and `build_features := user_features + ",bench-internals"` is added — everything except `loom-core`. `clippy`, `clippy-fix`, `test`, `coverage`, `coverage-lcov` use it.
- `ci.yml`: the `versions` job reads the value with `just --evaluate build_features` and `clippy`, `test`, `bench`, `msrv`, `doc` and `coverage` use that instead of `--all-features`.

## Why

`--all-features` enables `loom-core`, and that feature *removes* code from the build rather than adding to it. `subscription::signal` is gated `#[cfg(not(all(feature = "loom-core", test)))]`, so under `cargo test --lib --all-features` the module is compiled out before its tests are collected — and the `test` job runs exactly that, on every OS and both toolchains:

```
$ cargo test --lib                | grep -c "signal::"
3
$ cargo test --lib --all-features | grep -c "signal::"
0
```

The `test` in that gate bounds what was lost, and it is narrower than it first looks. Where `cfg(test)` is off the `all(..)` is false and the module is **kept**, so `doc` documented it and the plain `--lib` unit of `--all-targets` linted and MSRV-checked it throughout. What `--all-features` removed is the module from *test-target* builds: `test` never ran its three unit tests, and `coverage` reported the module at 0% — it was in the report all along, but the tests that would have covered it were the ones removed.

The gate itself is correct — `--cfg loom` reconfigures tokio and drops modules like `signal`, which the `loom` job's own comment records. What is wrong is using `--all-features` to mean "everything a user can enable", when `loom-core` and `bench-internals` are both documented as build-only.

`bench-internals` stays enabled, and that is why the set is "everything except `loom-core`" rather than `user_features`: it only adds. It is what the benches' `required-features` name, so without it `--all-targets` does not reach a bench at all, and the `#[doc(hidden)]` handles it exposes would go unlinted — coverage `--all-features` did have.

The `justfile` is the single source and CI reads it rather than restating it, which puts CI behind the staleness check `just test-doc-packaged` already runs against the crate's own feature table. A new feature that nobody adds to the list fails that check instead of silently going uncompiled.

## What changes in the numbers

Measured on rustc 1.97.0:

```
cargo test --lib                       540 total,  3 signal::
cargo test --lib --all-features        590 total,  0 signal::
cargo test --lib --features <new set>  585 total,  3 signal::
```

590 → 585 is eight rows lost and three gained. The eight are exactly:

```
kernel::accounting_core::tests::*                (4)
subscription::http::cell_core::tests::*          (4)
```

which is precisely what the `loom` job runs (`cargo test --features loom-core --lib -- cell_core accounting_core`, `RUSTFLAGS=--cfg loom`) — model-checked there rather than merely executed once. Nothing stops being tested.

Doctests are untouched at 73: `just test-doc` already used this list, and this only renames the variable.

## Coverage moves in both directions, once

The `coverage` job measures a different set of code after this, so Codecov compares against a base computed under the old one and reports a one-off change on this PR and on its merge commit. It is not a regression:

- **gains coverage**: `subscription::signal`, which the base measured at 0% (27 lines, 0 hits) and this head measures at 48.78% (41 lines, 20 hits), as its three unit tests start running
- **loses coverage**: the `cell_core` and `accounting_core` mirrors, which are well covered and leave the measured set — the instrumented file count drops 65 to 63

`codecov/project` carries a 0.5% threshold, so the net may or may not show. Neither codecov status is a required check.

## Verification

Run locally, all green:

```
just clippy                                      # signal now linted
just test-doc                                    # 73 (59 + 14 ignored)
just test-doc-packaged                           # staleness guard passes after the rename
RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --features <new set>
cargo test --lib --bins --tests --examples --features <new set>   # 585 + integration, 0 failed
actionlint                                       # no errors
```

`cargo doc` under `-D warnings` is clean. It is worth being precise that this was *not* the risky one: rustdoc never sets `--cfg test`, so that job had `subscription::signal` all along.

Closes #298

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,27 +36,43 @@ jobs:
   # added later that runs, say, `just check` would fail on a component that
   # used to be present by accident.
   #
-  # Reading both values out of the files that declare them is what keeps a job
-  # honest: the pin has one home (`rust-toolchain.toml`) and the MSRV has one
-  # home (`Cargo.toml`'s `rust-version`), and bumping either moves CI with it
-  # instead of leaving a job checking a version nothing claims any more.
+  # Reading these out of the files that declare them is what keeps a job
+  # honest: the pin has one home (`rust-toolchain.toml`), the MSRV has one
+  # home (`Cargo.toml`'s `rust-version`), and the feature set has one home
+  # (the `justfile`), so bumping any of them moves CI with it instead of
+  # leaving a job checking something nothing claims any more.
+  #
+  # `features` is what the jobs below enable instead of `--all-features`.
+  # `--all-features` turns on `loom-core`, which *removes* code from the
+  # build — `subscription::signal` is gated `cfg(not(all(feature =
+  # "loom-core", test)))`, so it and its three unit tests were compiled out of
+  # every job here. `build_features` is everything else, and the `justfile` is
+  # the source because `just test-doc-packaged` already fails when that list
+  # drifts from the crate's own feature table.
   versions:
     name: Declared Versions
     runs-on: ubuntu-latest
     outputs:
       pinned: ${{ steps.read.outputs.pinned }}
       msrv: ${{ steps.read.outputs.msrv }}
+      features: ${{ steps.read.outputs.features }}
     steps:
       - uses: actions/checkout@v7
+      - uses: taiki-e/install-action@just
       - id: read
-        name: Read the pinned channel and the declared MSRV
+        name: Read the pinned channel, the declared MSRV and the feature set
         run: |
           pinned=$(sed -ne 's/^channel = "\(.*\)"$/\1/p' rust-toolchain.toml)
           msrv=$(sed -ne 's/^rust-version = "\(.*\)"$/\1/p' Cargo.toml)
+          features=$(just --evaluate build_features)
           test -n "$pinned" || { echo "no channel in rust-toolchain.toml" >&2; exit 1; }
           test -n "$msrv" || { echo "no rust-version in Cargo.toml" >&2; exit 1; }
-          echo "pinned=$pinned" >> "$GITHUB_OUTPUT"
-          echo "msrv=$msrv" >> "$GITHUB_OUTPUT"
+          test -n "$features" || { echo "no build_features in the justfile" >&2; exit 1; }
+          {
+            echo "pinned=$pinned"
+            echo "msrv=$msrv"
+            echo "features=$features"
+          } >> "$GITHUB_OUTPUT"
 
   fmt:
     name: Format Check
@@ -87,7 +103,7 @@ jobs:
           components: clippy
       - uses: Swatinem/rust-cache@v2
       - name: Run clippy
-        run: cargo clippy --all-targets --all-features -- -D warnings
+        run: cargo clippy --all-targets --features ${{ needs.versions.outputs.features }} -- -D warnings
 
   test:
     name: Test
@@ -138,7 +154,7 @@ jobs:
         # `--test` flag and always runs its full scenarios; matrixing that
         # across every OS/toolchain makes Windows CI disproportionately slow
         # for no additional coverage.
-        run: cargo test --lib --bins --tests --examples --all-features
+        run: cargo test --lib --bins --tests --examples --features ${{ needs.versions.outputs.features }}
 
   bench:
     name: Benchmarks
@@ -159,7 +175,7 @@ jobs:
         # iteration, no sample, so what it gates is that it still compiles
         # and runs.
         run: |
-          cargo test --bench kernel_scan --all-features
+          cargo test --bench kernel_scan --features ${{ needs.versions.outputs.features }}
       - name: Run the load harness's smoke profiles
         # The full scenarios of the load harness leave CI: they are
         # deliberate acceptance/regression runs (RFC 0006 §5.1, RFC 0007
@@ -219,7 +235,7 @@ jobs:
           toolchain: ${{ needs.versions.outputs.pinned }}
       - uses: Swatinem/rust-cache@v2
       - name: Build documentation
-        run: cargo doc --no-deps --all-features
+        run: cargo doc --no-deps --features ${{ needs.versions.outputs.features }}
         env:
           RUSTDOCFLAGS: -D warnings
 
@@ -277,7 +293,7 @@ jobs:
           [[ $active == "${RUSTUP_TOOLCHAIN}-"* ]]
       - uses: Swatinem/rust-cache@v2
       - name: Check MSRV
-        run: cargo check --all-targets --all-features
+        run: cargo check --all-targets --features ${{ needs.versions.outputs.features }}
 
   coverage:
     name: Code Coverage
@@ -293,7 +309,7 @@ jobs:
       - name: Install cargo-llvm-cov
         uses: taiki-e/install-action@cargo-llvm-cov
       - name: Generate coverage
-        run: cargo llvm-cov --all-features --lcov --output-path lcov.info
+        run: cargo llvm-cov --features ${{ needs.versions.outputs.features }} --lcov --output-path lcov.info
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v7
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,6 +173,7 @@ jobs:
           echo "$active"
           [[ $active == "${RUSTUP_TOOLCHAIN}-"* ]]
       - uses: Swatinem/rust-cache@v2
+      - uses: taiki-e/install-action@just
       - name: Run tests
         # Benches are excluded from this matrix (and run once, on Linux
         # only, in the `bench` job below) since the load harness's
@@ -181,6 +182,15 @@ jobs:
         # across every OS/toolchain makes Windows CI disproportionately slow
         # for no additional coverage.
         run: cargo test --lib --bins --tests --examples --features ${{ needs.versions.outputs.features }}
+      # The mirrors' eight rows, which the set above cannot compile. They used
+      # to run on every leg of this matrix under `--all-features`, and losing
+      # that would move them from three required contexts to the `loom` job,
+      # which is advisory and ubuntu-only — so a regression in them would stop
+      # blocking a merge, and stop being seen off Linux at all. `loom` still
+      # model-checks them; this is the leg that says they compile and pass on
+      # three platforms and on beta.
+      - name: Run the loom mirrors' rows as ordinary tests
+        run: just test-mirrors
 
   bench:
     name: Benchmarks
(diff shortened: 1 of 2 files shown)


akiomik:
Head is c870cac, 22/22 green. Three fixed, one filed, one answered with evidence, one noted in the body.

## 1 — 479 lines left clippy (medium): fixed

Correct, and the distinction I missed is the one you draw: the `loom` job owns those tests' *execution*, and I wrote "nothing stops being tested" as though that covered their *linting*. It does not — that job is a plain `cargo test` with no `-D warnings` and no MSRV toolchain.

No single feature set can see the whole crate, because `loom-core` is the one feature that removes code. So there are two clippy passes now. Verified by putting a violation in `accounting_core`:

```
--features <build_features>   silent      (the file is not compiled at all)
--features loom-core          error: …    (caught)
```

I left the MSRV side alone on purpose: `rust-version` is a promise about what a *dependant* compiles, and a dependant never compiles `cfg(test)` mirrors. Say the word if you would rather have that leg too.

## 2 — the guard forced classification, not enablement (low): fixed

Exactly right, and it is this recipe's own failure mode one list over. `build_features` is now held to the feature table minus `loom-core`, alongside the existing `user_features` check. Verified against your scenario — add a feature, add it to the `user_features` exclusions, leave `build_features` alone:

```
jq: error: build_features is stale: expected …,probe-internals,… got …
```

## 3 — `api_surface.rs` (low): filed as #310

Latent for the reason you give — `rustdoc-json` builds without `cfg(test)`, so the `test` in that gate keeps `subscription::signal` in the surface. Fixing it is a Rust test with its own feature-naming and its own recorded-surface fixture, which is a different shape from this branch. #310 has the reasoning and notes that the failure message at line 109 would become the wrong instruction.

## 4 — pwsh and the comma list (low, needed verification): verified, no change

It works, and here is the run rather than an argument. `Test (windows-latest, pinned)` on this head:

```
Run cargo test --lib --bins --tests --examples --features dashmap,http,native-tls,rustls,…,bench-internals
…
test result: ok. 585 passed
```

No unexpected-argument error, and 585 is far above what a default-feature build produces, so the list arrived as one argument with the features in effect. Both Windows legs green. Leaving it as-is under the stopping rule — it is verified rather than assumed now, which was the open half of your finding.

## 5 — coverage moves both ways (low): noted in the PR body

Added a section: `subscription::signal` (495 lines, `stream()` unreached by any test) comes in, the two well-covered mirrors and their eight tests go out, so a one-off change lands on this PR and its merge commit and is not a regression. `codecov/project` carries a 0.5% threshold and neither codecov status is required.

## 6 — unreachable guard (nit): fixed

Confirmed independently: under `bash -eo pipefail` the script stops at the assignment, so `test -n "$features"` never runs. Removed, and replaced with a note on why the two `sed` reads *do* need theirs — `sed` prints nothing and exits zero when it matches nothing, which is the difference.

akiomik:
### Code review

Ran a five-agent review of this branch (repo conventions, shallow bug scan, git history, prior PR comments, in-file comment compliance) and fixed everything that scored at or above the bar. Head is 82b23db, 22/22 green.

Three findings were raised independently by four or five of the five agents, which is why they are listed first.

**1. `just clippy` no longer reproduced the `Clippy` job** (4 of 5 agents)

The loom lint pass I added was a raw `cargo` invocation with no recipe behind it, so a violation in `cell_core` or `accounting_core` failed CI with no local command that reproduced it. That is the failure the Clippy job's own comment says pinning its channel exists to avoid, and every other check here routes through `just` for exactly that reason.

There is a `clippy-loom` recipe now; `check`, `pre-commit` and `quick` run it, and the job calls `just clippy-loom`.

**2. "compiled out of every job here" was false** (3 of 5, one with a planted-violation experiment)

The gate is `cfg(not(all(feature = "loom-core", test)))`. Where `cfg(test)` is off the `all(..)` is false and the module is **kept** — so `doc` documented `subscription::signal` throughout, and the plain `--lib` unit of `--all-targets` linted and MSRV-checked it. An agent planted a `str_to_string` violation in the non-test code and the old `--all-features` command caught it.

What was actually lost is the module from test-target builds: three unit tests, and coverage. Corrected in `ci.yml`, the `justfile` and the PR body — which had claimed `doc` was "the one with real risk", the opposite of the truth.

**3. A comment named a shell option the step does not run under** (5 of 5)

No `shell:` key means `bash -e {0}`, not `-eo pipefail`. The conclusion held on `-e` alone, so this was a wrong reason rather than wrong behaviour — but the same comment argued `features` needed no `test -n` guard, and that argument only covers an unknown name. `just --evaluate` on a variable that exists and is empty exits zero; verified. The guard is restored for that case and each of the three now says which failure it is for.

**4. The mirrors' eight tests left the required checks** (scored 100, introduced by this branch)

`build_features` excludes `loom-core`, so those rows stopped running in the `test` matrix and ran only in `loom` — not a required context, and ubuntu-only. A regression in them went from failing three required checks to failing one advisory one, and stopped being seen off Linux.

`just test-mirrors` runs them as ordinary threaded tests and the `test` job calls it. Confirmed on this run that the step ran and passed on ubuntu, macOS and Windows. `loom` is unchanged and still model-checks them.

Fixed this way rather than by making `loom` required: that would have restored the gate but not the platforms.

**5. Also corrected**: the `justfile` claimed `test-loom` was "the only place `loom-core` is meant to be on", which this branch's own second pass contradicts; `user_features` claimed to be what CI reads, which is `build_features`; and the "building, linting, testing and covering" list named recipes that do not use it. The deliberate gaps — no MSRV and no cross-platform build of the mirrors' *compilation* under the MSRV toolchain — are now recorded as deliberate, with the reason: `rust-version` is a promise about what a dependant compiles, and a dependant never compiles a `cfg(test)` mirror.

**Not fixed, below the bar**: the staleness guard runs in `Doc Tests`, which is not a required context (scored 75). Real, but the drift it guards would still break the `test` job directly; it prevents silent mismatches only. Raising it needs a ruleset edit, which I have left as a maintainer decision.

🤖 Generated with [Claude Code](https://claude.ai/code)

akiomik:
### Code review

Five rounds run. Rounds 1-4 each found regressions introduced by the previous round's fix; round 5 found the two files clean and only the PR description stale, which is now corrected. Head is 535c7c6, 22/22 green.

**What the rounds found, in order**

1. The loom lint pass was a raw `cargo` invocation with no recipe, so a red `Clippy` reproduced nowhere locally — against the parity rationale this branch relies on. Now `just clippy-loom`.
2. "compiled out of every job here" was false: the `test` in `cfg(not(all(feature = "loom-core", test)))` means only test-target builds lost the module. Proven by planting a `str_to_string` violation that the old `--all-features` command caught.
3. A comment cited `bash -eo pipefail` on a step with no `shell:` key. Behaviour was right, the mechanism named was not.
4. The mirrors' eight rows had left the three required `Test` legs for the advisory ubuntu-only `Loom` job. Now `just test-mirrors`, back on all three platforms.
5. That recipe turned out to run **unbounded loom model checking**, not the "ordinary threaded tests" its comment claimed — loom's runtime is active whenever the mirrors compile, confirmed by `LOOM_MAX_BRANCHES=1` tripping loom's own branch-limit panic with `--cfg loom` nowhere in sight. Bounded to the same preemption count `test-loom` uses.
6. The recipe then swallowed every byte of failure output: `out=$(… 2>&1)` under `set -e` aborts at the assignment, so a failing mirror produced a non-zero exit and nothing else. Fixed and verified against a deliberately failing test.
7. Its row-count guard fired only at zero, so renaming one mirror still passed at `4 passed`. Now asserts exactly eight.
8. Four comment claims that measurement did not support: "six required legs" (three are), "0.05s either way" (bounded 0.05-0.11s vs unbounded 0.16-0.29s), compile times given as a stable range when they were one run, and "exercised" for items that are compiled but filtered out.

**Also**: `clippy`, `test-loom` and `coverage-lcov` were restated in the workflow rather than invoked — the loom job duplicated `RUSTFLAGS` and the preemption bound across two files. All three go through `just` now.

**Not fixed, deliberately**: the staleness guard runs in `Doc Tests`, which is not a required context (scored 75, below the bar). Raising it needs a ruleset edit and is a maintainer decision. `tests/api_surface.rs` still uses `--all-features`; latent, tracked in #310.

**The honest summary of this exercise**: one finding was a defect in the original change. Everything after it was a defect in a fix, and most were my prose claiming more than the evidence under it. The code converged after round 3; the comments took until round 5.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #304, opened 2026-08-27
test(kernel): release the quiescence hold structurally, not on a budget

### LINKED ISSUE #300: Conformance test premise is OS-scheduling dependent: the_mandated_supersession_window_never_invokes_the_superseded_spawner flakes on loaded runners
## Summary

`kernel::conformance::admission::the_mandated_supersession_window_never_invokes_the_superseded_spawner` failed on `Test (macos-latest, stable)` on a commit that changed no Rust at all, and passed on re-run with nothing changed. The failure is its own premise assertion, not the property it tests.

```
---- kernel::conformance::admission::the_mandated_supersession_window_never_invokes_the_superseded_spawner stdout ----
thread '...' panicked at src/kernel/conformance/admission.rs:334:5:
the hold ran out its budget instead of being released, so the window below was
never open — this row's premise failed, not the kernel

test result: FAILED. 583 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

The two Windows legs were cancelled by fail-fast, so one flake takes three matrix jobs with it.

## Why it is a premise failure, not a defect

The assertion is on `settle`'s budget:

```rust
driver.settle(THREADED_TURNS, || gate.entered());
assert!(
    !gate.exhausted(),
    "the hold ran out its budget instead of being released, so the window below was never \
     open — this row's premise failed, not the kernel"
);
```

`THREADED_TURNS` is documented as a hang guard, and its own rustdoc says the quantity it bounds is not under the test's control:

> Beside worker threads a turn is a rendezvous with another thread instead, and how many of them pass before a worker picks a granted producer up is **the operating system's business** — so the budget here is a hang guard whose value is mechanism, and nothing these series claim rests on it.

So the budget being exhausted means the runner did not schedule the worker within 4096 rendezvous turns. Nothing about the kernel's behaviour is implicated, and the test says so in its own message — but it still reports as a test failure, indistinguishable at the check level from a real one.

## Evidence it is not the commit

The failing commit changed `docs/migrations/0.10-to-0.11.md` and `justfile` only — zero `.rs` files, so the compiled library was byte-identical to the previous head, which passed. Re-running the failed jobs with nothing changed turned the run green.

## Scope

Two tests assert on `gate.exhausted()` as a premise, across `src/kernel/conformance/admission.rs` and `src/kernel/conformance/quit.rs`; three call sites use `settle(THREADED_TURNS, ...)`. Only the one above has been observed failing, but the others rest on the same OS-scheduling assumption.

## Options

The premise is a setup step, so failing the test is arguably the wrong response to it:

1. **Retry the premise.** Re-run the `settle` (or the whole test body) when `gate.exhausted()` is what failed, distinguishing "the OS did not cooperate" from "the kernel misbehaved". Keeps the budget tight, which is what makes it useful as a hang guard.
2. **Raise `THREADED_TURNS` on CI.** Cheapest, but it only moves the threshold — a loaded enough runner still trips it, and a larger value weakens the hang guard for everyone.
3. **Fail the premise differently.** If a premise failure produced a skip or a distinct diagnostic rather than a test failure, a reviewer would not have to read the log to tell it apart from a real regression.

Option 1 looks closest to what the rustdoc already asserts — that nothing the series claims rests on the number.

## Notes

Found while reviewing #296. Not fixed there because it is a test-design decision, and the PR changed no Rust.

akiomik:
## What

`the_mandated_supersession_window_never_invokes_the_superseded_spawner` held a stopped run open for a counted number of `thread::yield_now()` calls, and failed its own premise assertion when that budget ran out. The hold now blocks on a `Condvar`, the driver that runs the script owns the release, and the premise the whole thing rests on is checked rather than assumed.

- `GatedQuiescence::drop` blocks on `QuiescenceGate::wait` instead of spinning a budget, inside `tokio::task::block_in_place` so the executor keeps turning while the worker is held.
- `threaded_driver_with` returns a `GatedDriver`, which takes the gates off the script's own feeds and opens them in its `Drop`. There is no second constructor, no guard to bind, no gate argument, and no field order to preserve.
- `QuiescenceGate::wait` declines to wait if it is reached on the thread that has to release it, so a violated premise is a named assertion failure rather than a hang.
- `QuiescenceGate::exhausted` and both assertions on it are gone: the premise can no longer fail, and one of the two could no longer fail either way.

## Why the budget had to go rather than grow

`exhausted` had one writer, `GatedQuiescence::drop`, and the budget it spent was `HANDSHAKE_YIELDS` — 1,000,000 yields on a worker thread — not the `THREADED_TURNS` the issue named. That budget is spent *concurrently with* the script's progress through the window, and neither bounds the other, so which finishes first is the scheduler's business. Measured on an M1 Max (10 cores, rustc 1.97.0, `-O`): 1,000,000 yields is 152 ms idle and 2.14 s beside eight spinners, while the hold actually spends 122 yields idle and at most 5144 over 60 runs at 4x oversubscription. Wide margin, unbounded tail. The full correction to the issue's diagnosis is [here](https://github.com/akiomik/tears/issues/300#issuecomment-5433993163).

Every deadline-shaped alternative is closed by contract: `clippy.toml`'s `disallowed-methods` (RFC 0009 §3.1, INV-C1) bans `Instant::now`, `thread::sleep`, `thread::park_timeout`, `Condvar::wait_timeout` and `Receiver::recv_timeout`. Untimed blocking stays allowed, which is what this uses.

## What the review changed

The first version of this branch was wrong in a way worth recording, since the shape recurs.

**It deadlocked.** A worker blocked *inside a task* is indistinguishable from a worker running a long one, so the scheduler counts it as active and need not wake its parked neighbour for the no-op task `TestDriver::turn` injects. That task stranded, `turn` never returned, and the `settle` whose turn budget was meant to be the bound never got to spend it. Reproduced at 2 hangs in 800 runs at 16x parallelism; the counted version stalls and completes under the same load, so it was a regression from rare *failure* to rare *hang*. `block_in_place` is the missing announcement — it hands the worker's core to another thread — and is guarded on `RuntimeFlavor::MultiThread`, the flavour check mattering because `block_in_place` panics on a current-thread runtime and a panic in a drop during unwinding aborts the process.

**Its safety rested on obligations the types did not hold to.** The release was a guard the script had to bind, after the driver, and every way of getting that wrong was a hang. It is now `GatedDriver`'s `Drop`, which runs before any field is dropped — so field order, binding position, and remembering the guard at all stopped being things to get right. The gates come off the script rather than from an argument, so the gate released is the gate the source holds.

**Its unbounded wait rested on an unstated premise.** That the abort drops a cancelled run's stream on a worker and never on the script's thread is the kernel's behaviour, not this fixture's, and no `block_in_place` check would catch a change to it — a multi-thread `block_on` thread reports `MultiThread` and runs the closure inline. So `wait` compares against the gate's `releaser`, `threaded_driver_with` asserts that is the thread building the driver, and `GatedDriver` is `!Send` so it is also the thread dropping it.

## Verification

- window row: 800 runs at 16x parallelism — 0 hung, 0 failed
- whole `--lib` suite at `--test-threads=24`: 24 runs at 8x parallelism — 0 hung, 0 failed
- forced `panic!` inside the window: fails in 0.02 s
- fields of `GatedDriver` swapped **and** a panic injected: still 0.02 s, where the field-order design hung until killed at 90 s
- `open` regressed to a bare `notify_all`: the fixture row fails in 5 s with its bound's message
- `GatedDriver: !Send`: confirmed by compile error
- `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --lib` 537 passed; doctests 73, unchanged

Three fixture rows carry the gate's contract: a hold begun after its release waits for nothing, a hold reached on the releasing thread declines to wait, and dropping a driver opens the gates its script declared.

## Not fixed here

`MidBatchHandshake::open_and_await_commit` still spends `HANDSHAKE_YIELDS`. Its counted bound is right — the party that must signal is the producer, inside a run the script does not own, and a body that parks before its send keeps its sender alive so not even `Disconnected` arrives — but a yield costs about 154 µs in a debug build with a blocked peer, so a producer that never commits takes ~154 s to be reported. Sizing, on a path nothing currently reaches; worth its own issue.

Closes #300

### DIFF
--- a/src/kernel/conformance/admission.rs
+++ b/src/kernel/conformance/admission.rs
@@ -53,7 +53,7 @@ use crate::kernel::arbiter::WakeSource;
 
 use super::support::{
     Feed, ProbeSource, QuiescenceGate, Script, TEST_TURNS, THREADED_TURNS, accept, accept_within,
-    cap, config, driver, driver_with, parking_effect, step_when_ready, threaded_driver_with,
+    cap, config, driver, driver_with, gated_threaded_driver_with, parking_effect, step_when_ready,
 };
 
 // INV-SE1: one spawner invocation per admitted run, made at the admission
@@ -299,7 +299,11 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
     let first = ProbeSource::gated("a", gate.clone());
     let superseded = ProbeSource::silent("b");
     let newest = ProbeSource::silent("c");
-    let (mut driver, journal) = threaded_driver_with(
+    // Driven through the gated constructor, not `threaded_driver_with`: the
+    // driver it returns owns the gate's release and drops it before shutting
+    // the executor down, so a failure inside the window below unwinds into a
+    // test failure rather than a join on the held worker (`GatedDriver`).
+    let (mut driver, journal) = gated_threaded_driver_with(
         Script::new(parking_effect([1, 2]))
             .feeding([
                 Feed::new(first.clone()),
@@ -310,12 +314,8 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
             .redeclaring(1, ["b"])
             .redeclaring(2, ["c"]),
         config().batch_max_messages(cap(1)),
+        &gate,
     );
-    // After the driver, so it drops before it: an assertion below that fails
-    // unwinds through this guard, which opens the gate and lets the held
-    // worker finish its drop, and only then does the driver shut the executor
-    // down and join that worker. See `QuiescenceGate::release_on_drop`.
-    let _release = gate.release_on_drop();
     let trigger = driver.boot().started[0].clone();
     assert_eq!(first.admissions(), 1, "boot admitted A");
 
--- a/src/kernel/conformance/admission.rs
+++ b/src/kernel/conformance/admission.rs
@@ -53,7 +53,7 @@ use crate::kernel::arbiter::WakeSource;
 
 use super::support::{
     Feed, ProbeSource, QuiescenceGate, Script, TEST_TURNS, THREADED_TURNS, accept, accept_within,
-    cap, config, driver, driver_with, gated_threaded_driver_with, parking_effect, step_when_ready,
+    cap, config, driver, driver_with, parking_effect, step_when_ready, threaded_driver_with,
 };
 
 // INV-SE1: one spawner invocation per admitted run, made at the admission
@@ -299,11 +299,7 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
     let first = ProbeSource::gated("a", gate.clone());
     let superseded = ProbeSource::silent("b");
     let newest = ProbeSource::silent("c");
-    // Driven through the gated constructor, not `threaded_driver_with`: the
-    // driver it returns owns the gate's release and drops it before shutting
-    // the executor down, so a failure inside the window below unwinds into a
-    // test failure rather than a join on the held worker (`GatedDriver`).
-    let (mut driver, journal) = gated_threaded_driver_with(
+    let (mut driver, journal) = threaded_driver_with(
         Script::new(parking_effect([1, 2]))
             .feeding([
                 Feed::new(first.clone()),
@@ -314,7 +310,6 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
             .redeclaring(1, ["b"])
             .redeclaring(2, ["c"]),
         config().batch_max_messages(cap(1)),
-        &gate,
     );
     let trigger = driver.boot().started[0].clone();
     assert_eq!(first.admissions(), 1, "boot admitted A");
@@ -345,6 +340,13 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
     // `{B}` → `{C}` inside that window. The barrier defers this
     // re-evaluation's admissions too, so the generation that was declared
     // when the stop was issued passes without ever being admitted.
+    //
+    // That the hold is still holding across this pass is not re-asserted
+    // after it. Nothing between here and `gate.open()` below can release it —
+    // the gate has exactly two openers, that call and `GatedDriver`'s drop —
+    // so a repeat of the quiescence check above could not fail, and an
+    // assertion that cannot fail reads as a guarantee it is not making. The
+    // check that carries the claim is the one before this pass.
     let evaluations = journal.evaluations();
     driver
         .step_pass(WakeSource::Data)
@@ -363,12 +365,6 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
         newest.admissions(),
         0,
         "and C waits behind the outstanding stop rather than admitting beside it"
-    );
-    assert_eq!(
-        first.quiescences(),
-        0,
-        "the hold was still holding across that pass, so the two readings above are the \
-         barrier's answer and not a released run's"
     );
 
     // The window closes: A quiesces, and the pass its notification begins
--- a/src/kernel/conformance/admission.rs
+++ b/src/kernel/conformance/admission.rs
@@ -311,6 +311,11 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
             .redeclaring(2, ["c"]),
         config().batch_max_messages(cap(1)),
     );
+    // After the driver, so it drops before it: an assertion below that fails
+    // unwinds through this guard, which opens the gate and lets the held
+    // worker finish its drop, and only then does the driver shut the executor
+    // down and join that worker. See `QuiescenceGate::release_on_drop`.
+    let _release = gate.release_on_drop();
     let trigger = driver.boot().started[0].clone();
     assert_eq!(first.admissions(), 1, "boot admitted A");
 
@@ -331,11 +336,6 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
     // The window opens: A's dismantling has begun on a worker and is held
     // there, so the stop is outstanding and the quiescence has not happened.
     driver.settle(THREADED_TURNS, || gate.entered());
-    assert!(
-        !gate.exhausted(),
-        "the hold ran out its budget instead of being released, so the window below was never \
-         open — this row's premise failed, not the kernel"
-    );
     assert_eq!(
         first.quiescences(),
         0,
@@ -364,8 +364,9 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
         0,
         "and C waits behind the outstanding stop rather than admitting beside it"
     );
-    assert!(
-        !gate.exhausted(),
+    assert_eq!(
+        first.quiescences(),
+        0,
         "the hold was still holding across that pass, so the two readings above are the \
          barrier's answer and not a released run's"
     );
--- a/src/kernel/conformance/admission.rs
+++ b/src/kernel/conformance/admission.rs
@@ -53,7 +53,7 @@ use crate::kernel::arbiter::WakeSource;
 
 use super::support::{
     Feed, NEGATIVE_TURNS, ProbeSource, QuiescenceGate, Script, TEST_TURNS, THREADED_TURNS, accept,
-    accept_within, cap, config, driver, driver_with, parking_effect, step_when_ready,
+    accept_within, cap, config, driver, driver_with, parking_effect, spend_turns, step_when_ready,
     threaded_driver_with,
 };
 
@@ -340,11 +340,7 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
     // times in 20, reporting a supersession window that never opened; with
     // them, 0 in 20.
     driver.settle(THREADED_TURNS, || gate.entered());
-    let mut spent = 0_usize;
-    driver.settle(TEST_TURNS, || {
-        spent += 1;
-        spent > NEGATIVE_TURNS
-    });
+    spend_turns(&mut driver, NEGATIVE_TURNS);
     assert_eq!(
         first.quiescences(),
         0,
--- a/src/kernel/conformance/admission.rs
+++ b/src/kernel/conformance/admission.rs
@@ -52,8 +52,9 @@
 use crate::kernel::arbiter::WakeSource;
 
 use super::support::{
-    Feed, ProbeSource, QuiescenceGate, Script, TEST_TURNS, THREADED_TURNS, accept, accept_within,
-    cap, config, driver, driver_with, parking_effect, step_when_ready, threaded_driver_with,
+    Feed, NEGATIVE_TURNS, ProbeSource, QuiescenceGate, Script, TEST_TURNS, THREADED_TURNS, accept,
+    accept_within, cap, config, driver, driver_with, parking_effect, step_when_ready,
+    threaded_driver_with,
 };
 
 // INV-SE1: one spawner invocation per admitted run, made at the admission
@@ -330,7 +331,20 @@ fn the_mandated_supersession_window_never_invokes_the_superseded_spawner() {
 
     // The window opens: A's dismantling has begun on a worker and is held
     // there, so the stop is outstanding and the quiescence has not happened.
+    //
+    // The turns between the two are what make the second line a check. The
+    // hold marks `entered` on its way in and `quiesced` on its way out, so
+    // `settle` returning says only that the dismantling started — a hold that
+    // did *not* block marks both back to back, and this thread can read
+    // between them. Without the turns, forcing that case passed this row 10
+    // times in 20, reporting a supersession window that never opened; with
+    // them, 0 in 20.
     driver.settle(THREADED_TURNS, || gate.entered());
+    let mut spent = 0_usize;
+    driver.settle(TEST_TURNS, || {
+        spent += 1;
+        spent > NEGATIVE_TURNS
+    });
     assert_eq!(
         first.quiescences(),
         0,
--- a/src/kernel/conformance/support.rs
+++ b/src/kernel/conformance/support.rs
@@ -20,13 +20,12 @@
 //!   all. A blocking wait on a worker also has to announce itself with
 //!   [`block_in_place`], or the executor it is blocking stops being able to
 //!   turn.
-//!
-//! [`block_in_place`]: tokio::task::block_in_place
 //! - **One turn, one construction.** A turn is a spawn of a fresh no-op task
 //!   onto the executor plus a join on it (RFC 0008 §9.6), which is what
 //!   [`turn`] does and what [`TestDriver::settle`] does inside the driver.
 //!
 //! [`TestDriver::settle`]: crate::testing::driver::TestDriver::settle
+//! [`block_in_place`]: tokio::task::block_in_place
 
 use std::collections::{HashMap, VecDeque};
 use std::future::{Future, pending, ready};
@@ -1541,4 +1540,61 @@ pub async fn turn() {
     tokio::spawn(ready(()))
         .await
         .expect("a no-op task neither panics nor is aborted");
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // The gate stores its release rather than signalling it, which is what
+    // makes the order between a script's `open` and the dismantling it holds
+    // not load-bearing: a hold that begins after the release reads the flag
+    // and returns without waiting for anything.
+    //
+    // Run on a plain test thread on purpose. That is the arm the hold takes
+    // when there is no multi-thread worker whose core it could hand off —
+    // the case a runtime shutdown reaches, where blocking announces nothing
+    // because there is nothing left to announce it to.
+    #[test]
+    fn a_hold_begun_after_its_release_waits_for_nothing() {
+        let gate = QuiescenceGate::default();
+        let quiesced = Beacon::default();
+        gate.open();
+
+        drop(GatedQuiescence {
+            gate: gate.clone(),
+            quiesced: quiesced.clone(),
+        });
+
+        assert!(
+            gate.entered(),
+            "the dismantling recorded itself on the way in"
+        );
+        assert!(
+            quiesced.marked(),
+            "and reached its end rather than holding, with no release left to wait for"
+        );
+    }
+
+    // The release a `GatedDriver` owns opens the gate, and it does so before
+    // the driver it wraps is dropped. The first half is what this asserts
+    // directly; the second is the field order above it, which is why a
+    // failing window unwinds into a test failure rather than into a join on
+    // the thread the hold is running on.
+    #[test]
+    fn a_release_guard_opens_its_gate_when_dropped() {
+        let gate = QuiescenceGate::default();
+        let quiesced = Beacon::default();
+
+        drop(gate.release_on_drop());
+
+        drop(GatedQuiescence {
+            gate,
+            quiesced: quiesced.clone(),
+        });
+        assert!(
+            quiesced.marked(),
+            "the guard's drop released the hold that began after it"
+        );
+    }
 }
--- a/src/kernel/conformance/support.rs
+++ b/src/kernel/conformance/support.rs
@@ -1742,18 +1742,29 @@ mod tests {
     /// A hang guard, so the two costs it trades between are not symmetric and
     /// the value is not a midpoint.
     ///
-    /// Passing costs nothing whatever the value is: the loop exits at the
+    /// Passing costs nothing whatever the value is: the loops exit at the
     /// first mark, so nothing here is spent on a bound that is never reached.
-    /// Its two callers wait for different things and were measured
-    /// separately, 30 runs each, worst of each:
+    /// What has to fit under it is the longest *stall* — the gap between two
+    /// progress marks — not a whole script, which is what keeps the quantity
+    /// small enough to survive the conversion below. Worst observed, by
+    /// caller and platform:
     ///
-    /// |                              | idle | beside 16 busy-loop processes |
-    /// | ---                          | ---  | ---                           |
-    /// | a bare `drop` on a thread    |  174 |                             2 |
-    /// | a whole driver script's drop |  611 |                             3 |
+    /// |                              | macOS, 10 cores | Linux CI, 4 cores |
+    /// | ---                          | ---             | ---               |
+    /// | a bare `drop` on a thread    | (not sampled)   |             1913 |
+    /// | a whole driver script's drop |           6664 |             3511 |
     ///
-    /// The loaded column is the cheaper one because a yield under load is a
-    /// longer time, so the passing path's worst case is an *idle* machine.
+    /// The Linux column is 100 suite runs at 4x oversubscription and no
+    /// failures; the macOS column is 11 runs at default parallelism, ten of
+    /// them between 68 and 373 and one at 6664 with a build running beside
+    /// them. That single outlier is the number the margin is taken against,
+    /// because a thirtyfold tail is the thing a bound has to survive.
+    ///
+    /// A count is not a portable unit, which is why both columns exist: a
+    /// `thread::yield_now()` yields the scheduling quantum on Darwin and
+    /// returns almost immediately on Linux when the peer is already running.
+    /// Sampling one platform and calling the ratio a margin is how this
+    /// constant was first set, and it was wrong by an order of magnitude.
     ///
     /// Failing is the only thing a larger value delays, and it is measurably
     /// expensive: a yield costs about 154 µs in a debug build *with a blocked
@@ -1761,14 +1772,12 @@ mod tests {
     /// [`HANDSHAKE_YIELDS`] would report a regression 154 seconds later — a
     /// bound that slow reads as the hang it replaces.
     ///
-    /// So the value wants to be far above the worst of the column that
-    /// matters and far below that: `65_536` is roughly 107x over 611, for a
-    /// failure reported in about 10 seconds. The margin matters more than the
-    /// arithmetic suggests, since scheduler latencies are heavy-tailed and
-    /// this suite runs thousands of times on shared machines — a bound that
-    /// merely clears the observed maximum is how the budget this change
-    /// removed came to fail in the first place.
-    const HOLD_YIELDS: usize = 65_536;
+    /// So the value sits far above the worst stall and far below that:
+    /// `262_144` is about 39x over 6664 and 75x over 3511, for a failure
+    /// reported in around 45 seconds. A bound that merely clears the observed
+    /// maximum is how the budget this change removed came to fail in the
+    /// first place.
+    const HOLD_YIELDS: usize = 262_144;
 
     // The gate stores its release rather than signalling it, which is what
     // makes the order between a script's `open` and the dismantling it holds
--- a/src/kernel/conformance/support.rs
+++ b/src/kernel/conformance/support.rs
@@ -226,10 +226,11 @@ impl QuiescenceGate {
     /// that states it. A named failure, where the alternative is a job that
     /// times out with no test named at all.
     ///
-    /// The explicit `drop` is what `clippy::significant_drop_tightening`
-    /// asks for: the guard is dead once the loop has read `true` out of it,
-    /// and releasing it there rather than at the end of the scope keeps the
-    /// next holder from waiting on a lock nobody is using.
+    /// The trailing `drop` buys no contention back — it is the last statement
+    /// in the function, so the guard would fall at the same point without it.
+    /// It is there because `clippy::significant_drop_tightening` asks for it
+    /// and the lint job runs with `-D warnings`; removing it fails that job
+    /// rather than changing when anything is unlocked.
     fn wait(&self) {
         if thread::current().id() == self.releaser {
             return;
@@ -270,15 +271,18 @@ impl Drop for GatedQuiescence {
     /// Held with a blocking wait rather than a waker or a deadline: a drop
     /// cannot await, and a deadline would be a clock read.
     ///
-    /// What ends the wait is structural rather than counted. The script owns
-    /// a [`GateRelease`] declared after its driver, so the gate is opened on
-    /// every path out — the ordinary one at the end of the window, and an
-    /// unwinding one, where the guard drops first and this hold finishes
-    /// before the driver's shutdown reaches the join. A counted bound was the
-    /// alternative and it is the wrong shape here: the count would be spent
-    /// concurrently with the script's own progress through the window, so the
-    /// two race, and the hold releasing early fails a premise the kernel had
-    /// nothing to do with (issue #300).
+    /// What ends the wait is structural rather than counted, and the script
+    /// is not asked to remember any of it: the driver that runs the script
+    /// holds the gates the script declared and opens them in its own
+    /// [`Drop`], which runs before the field holding the executor does
+    /// ([`GatedDriver`]). So the gate is opened on every path out — the
+    /// ordinary one at the end of the window, and an unwinding one, where the
+    /// release still lands before the shutdown that would join this thread.
+    ///
+    /// A counted bound was the alternative and it is the wrong shape here:
+    /// the count would be spent concurrently with the script's own progress
+    /// through the window, so the two race, and the hold releasing early
+    /// fails a premise the kernel had nothing to do with (issue #300).
     ///
     /// Blocking a worker thread is what makes the window constructible at all:
     /// the abort drops the run's stream on a worker, so the driving thread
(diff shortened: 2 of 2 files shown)


akiomik:
All four findings addressed in a3f49c6. Finding 1 reproduced first — 2 hangs in 800 runs at 16x parallelism, matching your numbers — so this was a real regression and the review was right to block it.

## 1 (HIGH, deadlock) — fixed with `block_in_place`

Reproduced on the previous head before changing anything:

```
800 runs at 16x parallel -> 2 hung, 0 failed
```

Your mechanism is right, with one refinement to the wording: the injected task does not land in the blocked worker's *local* queue — `TestDriver::turn` is `executor.block_on(...)` from the test thread, which is not a worker, so `tokio::spawn` goes to the injection queue. The stranding is the same and for the reason you give: a worker blocked *inside a task* is indistinguishable from a worker running a long one, so the scheduler counts it as active and need not wake the parked neighbour to drain inject. `turn` never returns, and `settle`'s turn budget never gets the chance to be the bound.

`block_in_place` as you suggested, guarded as you suggested:

```rust
match Handle::try_current() {
    Ok(handle) if matches!(handle.runtime_flavor(), RuntimeFlavor::MultiThread) => {
        task::block_in_place(|| self.gate.wait());
    }
    _ => self.gate.wait(),
}
```

Same stress after the fix:

```
800 runs at 16x parallel -> 0 hung, 0 failed
```

## 2 — doc corrected

The "second worker keeps the executor able to turn" paragraph is gone. It now says the opposite explicitly — that a second worker does *not* save it, and why — and names `block_in_place` as the announcement that makes the hold safe. The reproduction numbers are in the rustdoc so the claim carries its evidence.

## 3 — enforced instead of documented

You are right that a tuple return does not fix it (elements drop left to right, so a `(Self, GateRelease)` from `ProbeSource::gated` declared above the driver still drops the driver first). What does fix it is field order in a type that owns both:

```rust
pub struct GatedDriver {
    release: GateRelease,   // fields drop in declaration order …
    driver: TestDriver<Scripted, TestBackend>,   // … so the gate opens before this shuts down
}
```

`release_on_drop` is private now, and `gated_threaded_driver_with(script, config, &gate)` is the only way to get a driver for a gated source. Omitted, immediately dropped, and bound-above-the-driver are all unreachable. `GatedDriver` derefs to the driver, so the row reads as before.

The one hole left is calling `ProbeSource::gated` with plain `threaded_driver_with`; that is now a named constructor to bypass rather than a binding to misplace, and both rustdocs say so.

## 4 — rationale reconciled

`MidBatchHandshake`'s doc no longer claims its wait is bounded "like every other wait here", nor that a deadline is the only way to bound a block. It now states why *its* bound stays counted while the gate's does not: there the script itself must release, so the release can be made structural; here the *producer* must signal, from inside a run the script does not own, and a body that parks before its send keeps its sender alive so not even `Disconnected` arrives. The module header is updated in the same commit and mentions the `block_in_place` requirement.

## Verification

- 800 runs of the row at 16x parallelism: 0 hung, 0 failed
- 40 runs of the whole `--lib` suite at 8x parallelism: 0 hung, 0 failed
- forced `panic!` inside the window: fails in 0.02 s, no deadlock
- `cargo clippy --all-targets --features <user+bench-internals> -- -D warnings`: clean
- `cargo test --lib`: 534 passed

Thanks — the reproduction was the part I got wrong, and the stress shape (16x parallel rather than CPU burners beside a serial run) is what my original verification was missing.

akiomik:
Follow-up on the two codecov checks, and on a red leg that was not this change.

**`codecov/patch` — one arm, now covered.** It was pointing at the branch `GatedQuiescence::drop` takes when there is no multi-thread worker whose core it could hand off. Nothing in the suite reached it, because in every series the dismantling lands on a worker — but it is the arm a runtime shutdown reaches, and the one that must *not* call `block_in_place`. b1d75a8 adds two rows on a plain test thread, which is what puts them on that arm:

- a hold that begins after its release waits for nothing — the "stores rather than signals" property, and the reason the order between a script's `open` and the dismantling is not load-bearing;
- a `GateRelease` opens its gate when dropped — the half of `GatedDriver`'s guarantee that is a behaviour rather than a field order.

What remains uncovered is `Deref::deref` for `GatedDriver`, three lines. `DerefMut` requires `Deref`, and the row only ever borrows mutably, so exercising it would mean adding a call for the sake of the counter. Left alone.

Both codecov checks are still red at 95.58% patch (target 95.61%) and 95.59% project (-0.02%). That is 0.03 and 0.02 points, inside the oscillation band documented in #299 — the same band that fails docs-only pull requests about half the time. Neither is a required check, and #301 makes `project` informational.

**`Test (macos-latest, beta)` went red and took the matrix with it.** Not this change: the job died in `Set up job`, before any Rust ran.

```
##[error]actions/checkout/v7/action.yml: (Line: 1, Col: 1, Idx: 0) … find character that cannot start any token.
##[error]Failed to load actions/checkout/v7/action.yml
```

A corrupt action download on that runner. Re-run with nothing changed: all six legs green, and `MSRV (1.88.0)`, `Loom`, `Clippy`, `Documentation`, `Doc Tests` and `Benchmarks` with them.

Head is now b1d75a8 with everything green except the two codecov numbers above.

akiomik:
All four closed in 1a1c70c, by mechanism in each case rather than by another paragraph — the common thread in your findings was that the last commit stated invariants the types did not hold to.

## medium: constructor and gate argument both optional

Both were, and the gate argument was the worse half: it was *named beside* the script rather than taken *from* it, so a different `QuiescenceGate::default()` compiled and hung.

There is one multi-worker constructor again. `threaded_driver_with` reads the gates off the script's own feeds:

```rust
let gates = script.gates();   // Script -> Feed -> ProbeSource::gate
```

So the gate released is the gate the source will hold at, there is no second constructor to bypass, and no argument that can disagree. A script with no gated source carries an empty vector. `gated_threaded_driver_with` is gone; `quit.rs` and `delivery.rs` are unchanged, since `GatedDriver` derefs.

## medium: unbounded on an unstated premise

You are right that the premise was doing all the work and that no `block_in_place` check would catch its violation — a multi-thread `block_on` thread reports `MultiThread` and runs the closure inline, so the failure would be silent.

Rather than document it, the gate records the thread it was constructed on, and a hold reaching that thread declines to wait:

```rust
fn wait(&self) {
    if thread::current().id() == self.releaser { return; }
    …
}
```

Then the run quiesces at once and the window's own assertion — "A is stop-requested and has not quiesced, which is the state the sequence needs" — fails in the file that states it, instead of the job timing out with no test named. It closes the same hole for `driver` / `driver_with`, where a gated source would otherwise hold the current-thread runtime's only thread.

## low: vacuous assertion

Confirmed, and your reasoning is exactly right — nothing between the window's halves can release the hold, so it could not fail. Removed, with what it was trying to say moved into the comment above the pass. The check before the pass is the one that carries the claim, and the premise check above now gives it something to catch: if the hold declines, that assertion is what fires.

## low: no test for the field order

The better answer was to stop depending on it. `GatedDriver` now has an explicit `Drop` that opens the gates, and `Drop::drop` runs before any field is dropped — so reordering, renaming or inserting a field cannot put the shutdown in front of the release.

Verified adversarially, which is the check the old design could not pass: fields swapped **and** a panic injected mid-window still fails in 0.02 s, where the field-order version hung until killed at 90 s.

## Fixture rows

Three, all reachable without a runtime except the last:

- a hold begun after its release waits for nothing (on a spawned thread, so it exercises the flag rather than the thread guard)
- a hold reached on the releasing thread declines to wait — the premise check, asserting the gate was never opened
- dropping a driver opens the gates its script declared — the wiring and the release together

## Verification

- 800 runs of the row at 16x parallelism: 0 hung, 0 failed
- 40 runs of the whole `--lib` suite at 8x parallelism: 0 hung, 0 failed
- fields swapped + panic mid-window: fails in 0.02 s
- `cargo clippy --all-targets --features <user+bench-internals> -- -D warnings`: clean
- `cargo test --lib`: 537 passed; doctests 73, unchanged

akiomik:
All three in 2a7560f.

## 1 — `GatedQuiescence::drop` doc described a type that no longer exists

Confirmed: the only `GateRelease` left in the tree was the unrelated `FetchGateReleases` in `test_support/async_utils.rs`. The paragraph described the discarded design — a guard the script declares after its driver — in the very doc that should state the current one, contradicting both the module header and `GatedDriver` two hundred lines below, with a dangling intra-doc link that only `#[cfg(test)]` keeps out of the `doc` job.

Rewritten to describe what is there: the driver holds the gates its script declared and opens them in its own `Drop`, which runs before the field holding the executor, so no position and no binding is load-bearing. `grep -rn GateRelease src/kernel/` is now empty.

## 2 — `drop(open)` rationale was false

Right — it is the last statement in the function, so it shortens nothing for anyone. It now says what it is: a `clippy::significant_drop_tightening` requirement, kept because the lint job runs with `-D warnings`, not a contention win. Your check that removing it fails that job matches mine.

## 3 — the unbounded `join()`

The sharpest of the three, and correct: that was the one wait in the file with nothing behind it to end it, in a test whose subject is that such waits are ruled out. The thread is detached now and the result read from the beacon on a counted bound, in the file's existing style.

**The bound is measured, because the obvious constant turned out to be badly wrong here.** `HANDSHAKE_YIELDS` was the natural choice, so I regressed `open` to a bare `notify_all` — your exact scenario — to see the failure report:

```
test result: FAILED … finished in 154.46s
the hold did not finish in 1000000 scheduler yields …
```

A yield costs about **154 µs** in a debug build with a blocked peer, not the 152 ns I measured earlier under `-O` on an idle machine. So a million of them report a regression two and a half minutes later, which reads as the hang it is meant to replace.

What the row actually waits for is one thread scheduled once to run a drop that returns immediately. Measured over 30 runs each:

| | yields spent (worst of 30) |
| --- | --- |
| idle | 174 |
| beside 16 busy-loop processes | 2 |

Loaded is cheaper because a yield under load is a longer time. `HOLD_YIELDS = 4096` keeps a ~24x margin over the worst of that, and the same regressed `open` now reports in **0.39 s**.

Worth flagging separately, since it is not this PR's to fix: the same 154 µs applies to `MidBatchHandshake::open_and_await_commit`, which still spends `HANDSHAKE_YIELDS`. If a producer ever fails to commit, that row will take ~154 s in a debug build to say so. Its bound is sound — nothing structural can release that one, as its rustdoc now explains — but the value is sized for a cost that does not hold on the failure path. Happy to open an issue.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings` (the CI invocation): clean
- `cargo test --lib`: 537 passed
- window row, 800 runs at 16x parallelism: 0 hung, 0 failed
- whole suite at `--test-threads=24`, 24 runs at 8x parallelism: 0 hung, 0 failed
- regressed `open` to a bare `notify_all`: test fails in 0.39 s with the bound's message

akiomik:
All three in 0b25147, and the description is rewritten.

## 1 — `releaser` not tied to the thread that releases

Right, and the direction of the residual failure is what makes it worth closing: the check would guard a thread that releases nothing, while the *real* driving thread blocked on the gate short of the drop that would release it. That is the same hang, reached from the other side.

Taken both of your options rather than one, because they close different halves:

```rust
for gate in &gates {
    assert_eq!(thread::current().id(), gate.releaser, "…");
}
```

closes construction — a gate built elsewhere fails at the constructor, named, instead of hanging later. And `GatedDriver` is now `!Send`, which closes the gap between construction and drop, so the asserted thread is also the one that runs the release. Verified the marker actually bites:

```
error[E0277]: `*const ()` cannot be sent between threads safely
   = help: within `support::GatedDriver`, the trait `Send` is not implemented for `*const ()`
```

I preferred this to re-stamping `releaser` at driver construction, because `QuiescenceGate` is `Clone` and the source's clone — the one whose `wait` actually runs — already exists by then, so a re-stamp would have to be shared state to reach it.

## 2 — the bound's asymmetry

This is the one I got wrong, and the asymmetry is exactly as you state: the 154 µs per yield is a property of the *failing* path, where a blocked peer is the condition being reported. The passing path exits at the first mark whatever the bound is, so a larger value costs a passing suite nothing and delays only a test that was going to fail. Choosing between the two costs was the error; there is no trade to make.

`HOLD_YIELDS` is 65_536 — about 380x over the idle worst case of 174, against 4096's ~23x. Failing still reports promptly, measured with `open` regressed to a bare `notify_all` again:

| bound | failure reported in |
| --- | --- |
| `HANDSHAKE_YIELDS` (1,000,000) | 154 s |
| 65_536 | 5 s |
| 4096 | 0.39 s |

The doc now says why the value sits far above one number and far below the other rather than between them, and names the thin-margin failure mode as the one this change removed.

## 3 — description described a design that is gone

Correct, and the worst part is what you name: it asked for a decision about the guard's position, which no longer exists, so anyone reviewing from it would weigh a risk that is gone and miss the one that was real — finding 1 above. Rewritten around what is in the diff: `GatedDriver` owning the gates, the `block_in_place` announcement, the releaser check. `GateRelease`, `release_on_drop`, `#[must_use]` and the "trade-off, stated plainly" section are all out. The first version's mistakes are kept as a short section, since the deadlock shape is worth recording.

## Verification

- window row, 800 runs at 16x parallelism: 0 hung, 0 failed
- whole suite at `--test-threads=24`, 24 runs at 8x parallelism: 0 hung, 0 failed
- `GatedDriver: !Send`: compile error as above
- regressed `open`: fixture row fails in 5 s
- `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --lib` 537 passed

akiomik:
Both in 53d6a88.

## 1 — nothing detected the ordering guarantee

Correct, and the diagnosis is exact: without `boot()` no gated run existed, no hold ever blocked, and the row asserted that a drop flips a boolean. The guarantee it was named for was the one thing it could not see.

It now boots, stops the run, settles until `gate.entered()`, and drops the driver **with the hold still blocking**. The check that it actually carries the guarantee is to break the guarantee:

```
# `GatedDriver::drop` gutted to a no-op
dropping_a_driver_releases_a_hold_that_is_still_blocking ... FAILED   (7.6 s)
  the driver's drop did not return in 65536 scheduler yields: its shutdown
  joined the worker running the hold without the release having happened first
a_hold_begun_after_its_release_waits_for_nothing ... ok
a_hold_reached_on_the_releasing_thread_declines_to_wait ... ok
a_hold_reached_inside_the_runtime_declines_on_the_releasing_thread ... ok
```

So the manual field-swap check is permanent now, and it is the only row that sees it.

One wrinkle worth naming, since it came from an earlier round: the row cannot hand the driver to a watcher thread, because `GatedDriver` is `!Send` — that was the fix for the previous `releaser` finding. So the whole script runs on the spawned thread and only a `Beacon` comes back, watched here on a counted bound. That keeps a regression a failure rather than a hang, which is the constraint from the round before that.

## 2 — premise check covered on the wrong arm

Right. Added a row that drops the hold inside a multi-thread `block_on`, on the releasing thread, asserting it declines with the gate still shut. Confirmed the two rows genuinely take the two arms by making each unreachable in turn:

| probe | bare-thread row | in-runtime row |
| --- | --- | --- |
| `unreachable!` after `block_in_place(…)` | ok | **FAILED** |

## Unrelated flake found while stressing this

Not this branch's, and I checked rather than assumed. At `--test-threads=32` across 8 concurrent processes:

| | failures |
| --- | --- |
| `main` | **4 / 96** — three `panic::tests::test_compose_hook_restores_then_delegates_once`, one `quit::a_discarded_quit_s_dequeue_retires_its_origin_with_the_kernel_still_running` |
| this branch | 1 / 48 — the panic-hook row |

The panic-hook row does `take_hook` / `set_hook` / panic / `set_hook` against the **process-global** hook while other tests panic beside it, so its `order` picks up their panics and the assertion reads `["restore", "next", …]`. The `quit` one is a different row asserting `keyed_commands` last value, `Some(1)` where it wants `Some(0)`.

Neither is reachable from CI's default thread count, and both reproduce on `main` at a rate this branch does not exceed. Happy to open issues for them, along with the `MidBatchHandshake` bound sizing from the previous round — say the word and I will, rather than growing this PR.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test --lib`: 538 passed
- window row, 800 runs at 16x parallelism: 0 hung, 0 failed
- gutted `GatedDriver::drop`: the new row fails in 7.6 s, the others pass
- arm probes as tabled above

akiomik:
Both in 87fe4c1. The second one turned out worse than reported when I went to check it.

## 1 — `released.mark()` only on the success path

Right, and the diagnosis it produced was actively misleading: any unwind sailed past the mark, the bound expired, and the row reported the one thing that had not happened.

Two beacons rather than one, because "the thread ended" and "the thread got there" are different questions:

- `exited`, marked from a `DropMark`, so an unwind marks it too — the bound now means only "the drop never returned";
- `arrived`, marked on the last line, so a script that panicked fails an assertion here rather than passing.

`DropMark` alone would have made a panicking script *pass*, which is why it is a pair.

## 2 — the row did not assert its own premise

Correct, and adding the window row's check at that point is not sufficient — I checked before claiming it. `entered` and `quiesced` are marked either side of the hold, so observing the first says nothing yet about the second, and a hold that declined marks both back to back with the test thread free to look in between. Forcing `wait` to decline:

| | caught |
| --- | --- |
| bare `assert_eq!(first.quiescences(), 0, …)` | **2 / 5** |
| same check after spending turns first | **20 / 20** |

The turns use the `settle`-with-a-counter idiom already in this file, so a declined hold has every chance to finish before the check reads. Costs microseconds on the passing path, where the hold is blocked and the turns just cycle the executor.

## The three failures are now three different reports

Each verified by causing it:

| broken | reported as |
| --- | --- |
| `GatedDriver::drop` gutted | `the driver's drop did not return in 65536 scheduler yields: its shutdown joined the worker running the hold without the release having happened first` |
| script panics early | `the script thread ended before its last line: it panicked, and its own message says where` |
| `wait` always declines | `the hold is still blocking, which is what makes the drop below a release rather than a formality` |

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test --lib`: 538 passed
- support rows, 480 runs at 16x parallelism: 0 hung, 0 failed
- window row, 480 runs at 16x parallelism: 0 hung, 0 failed
- the three probes above

Still outstanding from earlier rounds, all agreed as out of scope for this branch: the `MidBatchHandshake` bound sizing, and the two pre-existing flakes (`panic::tests::test_compose_hook_restores_then_delegates_once` and `quit::a_discarded_quit_s_dequeue_retires_its_origin_with_the_kernel_still_running`, 4/96 on `main` at `--test-threads=32`). Say the word and I will file them.

akiomik:
Both in 4112b3c. The first one is the most serious finding of this review so far, and it was mine to catch: I identified the race, hardened the fixture row against it, wrote the measurement into that row's comment, and then did not carry it to the row the invariant actually rests on.

## 1 — the window row was a coin toss

Reproduced at exactly your number before changing anything:

```
window row with wait() forced to decline: 10 / 20 PASSED
```

And your point about why nothing downstream covers for it is the part that makes it serious: because the row steps pass-by-pass, an early quiescence does not admit C either, so every later assertion still holds and the row reports success on a sequence whose middle state was never constructed. A green INV-SE4 that tested the collapsed case twice.

Same three lines as the fixture row, now in `admission.rs`:

```rust
driver.settle(THREADED_TURNS, || gate.entered());
let mut spent = 0_usize;
driver.settle(TEST_TURNS, || { spent += 1; spent > NEGATIVE_TURNS });
assert_eq!(first.quiescences(), 0, "A is stop-requested and has not quiesced, …");
```

```
after: 0 / 20 PASSED, reporting at admission.rs:348 with the intended message
```

`PARK_TURNS` is now `NEGATIVE_TURNS` and public. With a second caller it stopped being about parks and is about the shape both share — turns spent before asserting that something has *not* happened — and its doc says the thing that went wrong here: read too early, "it has not happened yet" and "it does not happen" are the same reading.

## 2 — gated source with the current-thread constructors

Agreed, and agreed on the direction it fails in: a silent no-op window is worse than a named error, and it is the same obligation-the-type-does-not-hold shape this branch has been removing. `driver_with` now rejects a script carrying gates:

```
a gated source needs `threaded_driver_with`: on the current-thread executor its hold runs on the
driving thread and declines, so the window it exists to open is never open
```

Verified by building exactly that combination in a temporary row — it fails at the constructor, before any driving.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test --lib`: 538 passed
- window row, 480 runs at 16x parallelism: 0 hung, 0 failed
- support rows, 480 runs at 16x parallelism: 0 hung, 0 failed
- both guards checked by causing what they guard against (forced decline; gated script on `driver_with`)

Thanks for pushing on this one — the fixture row's own comment carried the measurement that should have told me to look here.

akiomik:
All three in cee1aa7.

## 1 — the `_` arm fell the wrong way

Agreed on the direction argument, which is the whole of it: misidentify a future flavour into the plain wait and you get the unannounced block that hung 2 in 800 before `block_in_place`, silent and unnamed; misidentify it into the announcement and the worst case is a panic where it is misused, in a drop that is not unwinding. So the two cases that must *not* announce are named and everything else announces:

```rust
match Handle::try_current().map(|handle| handle.runtime_flavor()) {
    Err(_) | Ok(RuntimeFlavor::CurrentThread) => self.gate.wait(),
    Ok(RuntimeFlavor::MultiThread | _) => task::block_in_place(|| self.gate.wait()),
}
```

Re-checked the arms still split as intended by making the announcing one unreachable: the in-runtime row fails, the bare-thread row passes.

The `LocalSet` hazard is recorded in the rustdoc as not covered, since there is no `LocalSet` in the crate and the runtime exposes no way to ask.

## 2 — two constants coupled

Right, and `NEGATIVE_TURNS`'s own doc invites the move that breaks it. `spend_turns(&mut driver, turns)` derives the budget from the count:

```rust
pub fn spend_turns<P: Program, B: Backend>(driver: &mut TestDriver<P, B>, turns: usize) {
    let mut spent = 0_usize;
    driver.settle(turns + 1, || { spent += 1; spent > turns });
}
```

Both call sites use it. Verified by raising `NEGATIVE_TURNS` to 32, past `TEST_TURNS` — both rows still pass, where the inline form would have failed on `bounded settle exhausted … after 16 executor turns`.

## 3 — the promised message could be nowhere

This is the one I would not have found, and the reasoning is exactly right: on a spawned thread the hook is the only route, and `silently` replaces it process-wide with this row holding no guard.

Taken the payload option rather than the `hook_guard()` one, so the row does not serialize against the panic-class series just to keep its diagnostics: the script is wrapped in `catch_unwind` and the payload reported from the watcher. Verified with the script forced to panic, in both hook states:

| hook | reported |
| --- | --- |
| normal | `the script thread panicked: assertion 'left == right' failed: TEMP: unrelated premise failure` |
| silenced process-wide | *same* |

Worth noting my first attempt at that probe passed and I nearly believed it — the patch had missed its anchor after `cargo fmt` re-indented the closure, so a stale binary ran. The results above are from a probe whose compile output I checked.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test --lib`: 538 passed
- window row, 480 runs at 16x parallelism: 0 hung, 0 failed
- support rows, 480 runs at 16x parallelism: 0 hung, 0 failed
- all three fixes checked by causing what they guard against

akiomik:
Both in 73a46a6.

## 1 — the guard was at one of three constructors

"The strength of the rule is whichever sibling forgets it" is exactly right, and `failing_driver` and `park_kernel` both forgot it. The rule is a method on `Script` now and all three current-thread constructors call it:

```rust
fn assert_ungated(&self) {
    assert!(self.gates().is_empty(), "a gated source needs `threaded_driver_with`: …");
}
```

Verified by building that script at each of the three:

```
temp_a (driver_with)     ... FAILED
temp_b (failing_driver)  ... FAILED
temp_c (park_kernel)     ... FAILED
```

## 2 — the bound's message did not hold its cause

Also right, and it is the file's own standard, so it should not have needed pointing out twice. A third beacon marks reaching the drop, and the verdict splits:

```rust
let cause = if watch_at_drop.marked() {
    "the driver's drop did not return: its shutdown joined the worker running the hold …"
} else {
    "the script did not reach its drop: something before it outran this bound, and the
     release is not implicated"
};
```

Verified by stalling the script before the drop — it now reports the second, where before it would have blamed the release.

### And the margin was justified from the wrong measurement

You are right that 174 is the bare-`drop` rows' number and this row was never measured. It is worse, as you would expect from a two-worker script:

| worst of 30 | idle | beside 16 busy-loop processes |
| --- | --- | --- |
| bare `drop` on a thread | 174 | 2 |
| whole driver script's drop | **611** | 3 |

So the margin is ~107x, not ~380x. Left the value at `65_536` — 107x is still ample against a failure path that reports in ~10 s — but the doc tabulates both callers and takes its margin from the one that matters, rather than quoting a number measured elsewhere.

## Verification

- `cargo clippy --all-targets --all-features -- -D warnings`: clean
- `cargo test --lib`: 539 passed
- window row, 480 runs at 16x parallelism: 0 hung, 0 failed
- support rows, 480 runs at 16x parallelism: 0 hung, 0 failed
- both fixes checked by causing what they guard against

akiomik:
Filed the three follow-ups that came out of this review, so they stop living in PR comments:

- **#305** — `HANDSHAKE_YIELDS` is sized against the passing path's per-yield cost, not the failing path's, so an unanswered handshake takes ~154 s to report. The bound itself is right and the issue says why, so it is not read as an argument for removing it.
- **#306** — `test_compose_hook_restores_then_delegates_once` takes `hook_guard()`, but that only serializes it against tests that also take it, and the 31 `#[should_panic]` rows do not. Its own failure payload proves the mechanism: the recorded vector is the expected one repeated.
- **#307** — the `quit` row reads `keyed_commands` immediately after its last `step_pass`, on the driving thread, for a value a worker writes.

One correction to what I said in an earlier comment on this PR: I described the panic-hook row as running without a guard. It does take `hook_guard()` — the gap is that the guard has no counterparty in the `#[should_panic]` rows. #306 has it right.

#307 is the same class as the premise race this PR fixed, and #305 came out of the per-yield measurement this PR needed for a different bound, so all three are this review's by-products rather than new ground.

akiomik:
Fixed in 52ad027.

Confirmed exactly as described: inserting `assert_ungated` above `gates` ran the two blocks together, so the assertion helper's rustdoc summary became "Every `QuiescenceGate` a declared source of this script holds" — a description of what it does not do — and `gates` was left with none. Split, each function keeping the block written for it.

Also swept the file for the same mistake in the other insertions this session (`panic_text`, `bound_verdict`, `spend_turns`), since it is an artefact of how I added them rather than of any one of them. Each owns its summary:

```
spend_turns    Hands the driver exactly `turns` turns, and asserts nothing.
panic_text     What a caught panic's payload says, for a failure that happened on a …
bound_verdict  Which of two things a watcher's exhausted bound caught, given whether the …
assert_ungated Rejects this script at a current-thread constructor.
gates          Every `QuiescenceGate` a declared source of this script holds.
```

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --lib` 540 passed. No behaviour change.

The three follow-ups from earlier rounds are filed: #305 (`HANDSHAKE_YIELDS` sizing), #306 (the hook guard's missing counterparty), #307 (the `quit` row's gauge read).

akiomik:
Fixed in 2b88df5.

"asserts nothing" was wrong, and the scenario you name is the right way to find it: a row that released a grant and reached for this helper to show nothing had arrived would panic on `settle`'s misuse message, naming neither the helper nor what the row was trying to establish.

**Documented rather than routed around**, for two reasons. `TestDriver::turn` is private, and `tears::testing` is public API guarded by `tests/api_surface.rs` — spending turns without `settle` would mean widening that surface for a fixture helper. And it would be the wrong direction anyway: the grant precondition is this helper's own rule, not an inherited inconvenience. A released send still in flight is exactly what makes an append to the guaranteed sequence possible during these turns (RFC 0008 §9.6), so with a grant outstanding "it has not happened" is not a claim any number of turns can establish. A series that wants to spend them there is asking a question the turns cannot answer.

So the `# Panics` section states both conditions and says that the second is load-bearing rather than incidental.

`cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings` clean; `cargo test --lib` 540 passed. Doc only.

## Standing state

All four PRs are green apart from the codecov project number, which is inside the band #299 records. The three follow-ups are filed: #305, #306, #307.

Two things left, both needing a decision rather than work:

1. **#302 needs the ruleset updated as part of merging it.** It renames the `Test` matrix leg `stable` → `pinned`, and the ruleset requires `Test (ubuntu-latest, stable)` and its two siblings. Order: merge #301 and #304, update the three contexts, then merge #302 — updating first blocks everything, merging first blocks #302 itself.
2. **#303 has never run CI**, because its base is #302's branch and the workflow triggers on pull requests into `main` only. It needs retargeting once #302 lands. It is verified locally (clippy, test, doc, doctest, packaged doctest) but that is not the same as a run.

akiomik:
Round 12 closed, and the measurement changed the answer. Head is db0b2e8, 21/21 green, `mergeStateStatus=CLEAN`.

## The Linux measurement

Ran as a temporary CI job on `ubuntu-latest` (added, read, reverted — the branch's `.github/` diff against `main` is empty). Not treated as a benchmark: a shared runner cannot measure absolute cost, so the only quantity collected was the **ratio** between the bound and the worst stall, taken as a maximum over many runs under oversubscription. Runner noise pushes stalls up, which is the conservative direction, so it works in the measurement's favour.

**100 suite runs, 4 cores, 4x oversubscription, 0 failures:**

| | worst stall | at the old 65_536 |
| --- | --- | --- |
| bare `drop` on a thread | 1913 | 34x |
| whole driver script's drop | 3511 | 18.7x |

So Linux is the *safer* side — but only because the previous commit made the bound cover a stall rather than a whole script. Your diagnosis of the direction was right; the fix had already removed the exposure.

## What it changed: macOS was the binding side, and my earlier figure was wrong

Eleven macOS runs: ten between 68 and 373, and **one at 6664** with a build running beside them. Against that outlier the old value was a **9.8x** margin — not the 107x I reported two rounds ago. That figure came from a different measurement (total spins, `--test-threads=1`, one platform) that I carried into a table it did not belong to. A thirtyfold tail is exactly what a bound has to survive, so it is the number the margin is now taken against.

`HOLD_YIELDS` is `262_144`: ~39x over the macOS outlier, ~75x over the Linux worst, failure reported in ~45 s rather than ~11 s. The doc carries both columns, their sample sizes and conditions, and states why one platform's ratio is not a margin.

## Stopping rule — adopted

Taking your three criteria as the merge-blocking test from here:

1. affects shipped code
2. turns CI red or hangs on a healthy tree
3. a test passes without checking its invariant

Applied to this round, the `HOLD_YIELDS` finding was (2) and is fixed. **Any further latent-only finding I will decline and file rather than fix in this PR** — as you say, for `cfg(test)` fixtures the cost of fixing after merge is the same as before it.

Your correction to my self-assessment is accepted: the error was not defining a stopping condition at round 2, not the size of the change. I had proposed shrinking to a one-line constant bump; you are right that it would have drawn the same density of findings, does not close #300, and was unbuilt. I have dropped that option.

Standing follow-ups: #305, #306, #307. This PR is ready to merge from my side.

akiomik:
History cleaned. Head is now f22e834, 21/21 green, `mergeStateStatus=CLEAN`.

## The TEMP pair is gone

`git rebase --onto a908bab 4a06a9b` — dropped rather than squashed, as you suggested, since every other commit stands on its own and squashing would lose that.

Verification that only those two went:

```
tree before  89f21a1b7143ad4988c9e133e231eb162320c3ad
tree after   89f21a1b7143ad4988c9e133e231eb162320c3ad   (identical)
commits      19 -> 17
```

All 17 remaining subjects match `type(scope): summary`, and no commit message contains "TEMP" or "Not for merge". `.github/` still has a zero-line diff against `main`.

## On how that got there

You are right that it was mine to prevent, and the fix is not "remember to revert". A temporary CI instrument does not belong on the branch that is going to be merged — with merge-commit integration, a reverted pair still lands. The measurement should have gone on a throwaway branch with a throwaway pull request, which gets the same CI and touches nothing that merges. Recorded that way rather than as a reminder.

Two other things from this round I have written down rather than left in the thread:

- on a shared runner, measure **ratios and maxima**, never absolute cost — which is what made the Linux result usable despite the runner varying
- the stopping rule, as the merge-blocking test, with latent-only findings declined and filed

## State

Ready to merge from my side. Recommended order, since #302 renames the `Test` matrix leg and the ruleset pins the old context names:

1. merge #301 and #304
2. update the three required contexts, `Test (…, stable)` → `Test (…, pinned)`
3. merge #302
4. retarget #303 to `main` — it has never had a CI run, because its base is #302's branch and the workflow only triggers on pull requests into `main`

Follow-ups filed: #305, #306, #307.

PR #309, opened 2026-08-27
ci: stop a failing matrix leg cancelling its neighbours

akiomik:
## What

Two things that were written for #302 and did not reach it. Both are corrected from what was reviewed there — see "Corrections" below.

- `fail-fast: false` on the `test` matrix
- `docs/testing.md` naming the check the API-surface job posts

## Why `fail-fast: false`

Under the default, one failing leg cancels the rest of the matrix. Three of these six legs are required contexts, so any single failure leaves a merge blocked on checks that never ran and can therefore only ever report `cancelled`. `fail-fast: false` lets each leg report its own result.

Whether a cancellation clears depends on why the leg failed, and both cases are real:

- **transient** — clears on a re-run. Run 33041812710 is the demonstration: at attempt 1 a `Set up job` failure (`Failed to load actions/checkout/v7/action.yml` on one runner) cancelled the other five legs; attempt 2 turned all six green at the same head sha with nothing fixed anywhere.
- **deterministic** — does not clear. This is where the `beta` leg matters, now that #302 made it real: it fails the same way every attempt and cancels the same neighbours each time. The `beta` legs are not required contexts, so they would be holding up a merge they have no standing to block.

The run is quoted from `gh api repos/akiomik/tears/actions/runs/33041812710/attempts/1/jobs`; in the web UI it is [attempt 1](https://github.com/akiomik/tears/actions/runs/33041812710/attempts/1). The bare run id resolves to attempt 2, which shows six successes. Its legs are named `stable`, because it predates #302's rename:

```
failure    Test (macos-latest, beta)
cancelled  Test (macos-latest, stable)
cancelled  Test (ubuntu-latest, beta)
cancelled  Test (ubuntu-latest, stable)
cancelled  Test (windows-latest, beta)
cancelled  Test (windows-latest, stable)
```

## What it costs

Not cancelling spends the survivors' runtime. The six legs total a median of **580s** across the last 25 runs (range 428–1047s; this branch's own run is 562s), so five of them running on rather than being cancelled is most of that.

It is runner *time*, not money: this repository is public, so GitHub-hosted runners are free and every run reports zero billable milliseconds. What is spent is occupancy and queue latency, against a merge that would otherwise be blocked on a healthy tree.

`cancelled` blocks rather than passes, so the default was never fail-open. This is about which failure costs less, not about safety.

## Why the doc line

`api-surface` is the YAML key; the job's `name:` is `Public API Surface`, and that is the check-run a reader sees. The original line called it a step, which was wrong in a different way.

## Corrections to what #302 reviewed

A code review of this branch found the justification, not the change, to be substantially wrong. Recorded because the errors are the same class this repository keeps catching:

1. The run transcript was **re-labelled**: the three cancelled legs were quoted as `pinned`, and the run has them as `stable`. Not an omitted condition — an altered one.
2. The premise was backwards. It argued the hazard opened when `beta` became real; the run cited three lines later is from before the split and shows a single leg failing alone, so unshared failures were always possible.
3. "until upstream is fixed" was wrong for the cited case, where a plain re-run recovered it.
4. The cost was wrong twice — first "a few runner-minutes" unmeasured, then "roughly 70 billable minutes", which is a bill never issued: the repo is public, so runners are free. The 950s behind it was also a cold-cache outlier against a 580s median.
5. "this branch" and "observed here" referred to #302 and to #304's branch respectively, and resolve to nothing once merged.
6. This body previously cited commit `d643dee` as evidence; it returns 422, having only ever existed on a detached `HEAD`.

## Why they missed #302

They were committed on a detached `HEAD`, so `git push origin ci/pin-toolchain-per-job` pushed a branch ref that had not moved — a no-op that exited zero. #302 merged as `e33f7a8` without them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,18 +137,34 @@ jobs:
     name: Test
     needs: versions
     strategy:
-      # A leg that fails must not cancel its neighbours, now that `beta` can
-      # fail on its own. Until this branch both legs ran the pin, so the only
-      # failures possible were ones every leg shared; a real `beta` picks up
-      # an upstream regression that none of the `pinned` legs have, and with
-      # the default `fail-fast` that cancels them — including all three
-      # required contexts, on a tree with nothing wrong in it, until upstream
-      # is fixed. Observed here already: on run 33041812710 one leg failed for
-      # an unrelated infrastructure reason and took the other five with it.
+      # A leg that fails must not cancel its neighbours. Under the default
+      # `fail-fast`, one failing leg cancels the rest of the matrix, and three
+      # of these six are required contexts — so any single failure leaves a
+      # merge blocked on checks that never ran and can therefore only ever
+      # report `cancelled`.
       #
-      # `cancelled` at least blocks rather than passes, so the default is not
-      # unsafe — it is just far heavier than what it buys, which is a few
-      # runner-minutes.
+      # Whether that clears depends on why the leg failed, and both cases are
+      # real. A transient failure clears on a re-run: that has happened here,
+      # an `actions/checkout` fault on one runner cancelling the other five,
+      # then a re-run turning all six green with nothing else changed. A
+      # deterministic one does not — a `beta` leg tracking an upstream
+      # regression fails the same way and cancels the same neighbours every
+      # time, and the `beta` legs are not required, so they would be holding
+      # up a merge they have no standing to block.
+      #
+      # The cost of not cancelling is the survivors' runtime: the six legs
+      # total a median of 580s over the last 25 runs, so five of them running
+      # on rather than being cancelled is most of that. It buys back a merge
+      # that would otherwise be blocked on a tree with nothing wrong in it,
+      # which is the better side of the trade.
+      #
+      # Runner *time*, not money — this repository is public, so GitHub-hosted
+      # runners are free and its runs report zero billable milliseconds. What
+      # is actually spent is occupancy and queue latency.
+      #
+      # `cancelled` blocks rather than passes, so the default was never
+      # fail-open. This is a question of which failure costs less, not of
+      # safety.
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,19 @@ jobs:
     name: Test
     needs: versions
     strategy:
+      # A leg that fails must not cancel its neighbours, now that `beta` can
+      # fail on its own. Until this branch both legs ran the pin, so the only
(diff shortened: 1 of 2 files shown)


akiomik:
### Code review

Three rounds. The `fail-fast: false` change itself was correct throughout; every finding was in the prose around it, and every one was mine. Round 3 reports no issues. Head is 5044aba, 22/22 green.

**Round 1** (five agents, unanimous on most of it)

1. **The quoted run transcript was re-labelled.** The three cancelled legs were given as `pinned`; run 33041812710 attempt 1 has them as `stable`. Not an omitted condition — an altered one. I wrote the leg names as they are *now* rather than as the record has them.
2. **The premise was backwards.** It argued the hazard opened when `beta` became real. The run cited three lines later predates the split and shows one leg failing alone, so unshared failures were always possible — the evidence refuted the claim it was placed under.
3. **"until upstream is fixed" was wrong for the cited case.** Attempt 2 turned all six legs green at the same head sha with nothing fixed. The real distinction is transient (clears on re-run) versus deterministic (does not), and only the second is the `beta` scenario.
4. **"this branch" and "observed here"** referred to #302 and to #304's branch. Neither resolves from `main` after merge.
5. **`docs/testing.md` named the YAML key.** `api-surface` is the key; `Public API Surface` is the check-run. The commit justified the edit with "a check-run name is what a required-status rule matches on" and then wrote the string that matches nothing.
6. **The PR body cited `d643dee`**, which returns 422 — it only ever existed on a detached `HEAD` — while resting the PR's reviewability on it.

**Round 2**

7. **"roughly 70 billable minutes" is a bill that is never issued.** This repository is public, so GitHub-hosted runners are free; `…/timing` reports zero billable milliseconds on every run sampled, macOS legs included. The 950s behind the figure was also the second-slowest of thirty runs, measured in the cold-cache window right after #302 landed. Median six-leg total is 580s; this branch's own run is 562s. That makes three attempts at stating the cost, wrong three different ways: unmeasured, then a fabricated currency, then an outlier.
8. **"can never report" contradicted paragraph four of the same comment.** Cancelled legs do report `cancelled` — which is precisely why they block.
9. The body's `/attempts/1/jobs` URL 404s in a browser; it is a REST path.

**Round 3**: every correction re-derived independently — the 580s median and 562s reproduced from a 25-run window, zero-billable confirmed across four runs, the attempt-1 transcript matched character-for-character, both URLs resolve, `d643dee` still 422s. No issues found.

**What is left deliberately**: the run id stays in this pull request and out of the workflow. Attempt data expires, the bare id resolves to attempt 2 which shows six successes, and no other tracked file in the repository cites one.

**Worth naming**: the change is two lines. The review found nine defects in the sentences justifying them, and the recurring shape was a claim whose stated grounds did not support it — the same class the reviews of #301, #302 and #303 kept landing on.

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #337, opened 2026-09-03
refactor(kernel): build a fresh run entry through one constructor

### LINKED ISSUE #336: Add a shared RunEntry test constructor to absorb field churn
## Background

`RunEntry` (`src/kernel/registry.rs:206`) is a `pub` struct with nine fields, no `#[non_exhaustive]`, and no constructor. It is built as a full struct literal at four sites:

- `src/kernel/producer.rs:259` and `:331` (production spawn paths)
- `src/kernel/bench_support.rs:202` (`#[cfg(feature = "bench-internals")]`)
- `src/kernel/registry.rs` — the registry unit row added by #319 (`an_exit_observed_after_the_drain_retires_the_entry`)

Adding a field to `RunEntry` now breaks all four literals, including the unrelated registry unit test, which must hand-mint an `AbortHandle` via `JoinSet::spawn(pending())` purely to fill the `abort` field. `bench_support`'s `RegistryScan::with_runs` already solves the `JoinSet`/`AbortHandle` fixture problem but is feature-gated out of `#[cfg(test)]`, so the unit test cannot reuse it.

## Proposal

Add an inherent constructor on the existing `impl RunEntry` block for the common "fresh `Running`, unrevoked, unexited" shape — e.g. `RunEntry::running(token, kind, scope, counter, gate, abort)` — and route the four sites through it (production sites included where they build that shape). A field addition then touches one place, and the registry unit test stops hand-rolling the nine-field literal.

## Why deferred from #319

#319 was a test-only PR pinning the removal condition's exit-observation half; adding a constructor touches the production `producer.rs` build sites too, which is a different change from that row's scope. Surfaced as a marginal reuse note across #319's review rounds.

akiomik:
Closes #336.

## What

`RunEntry` (`src/kernel/registry.rs`) is a nine-field record with no constructor, built as a full struct literal at four sites:

- `src/kernel/producer.rs` — `ProducerHarness::start` and `CleanupHarness::start`
- `src/kernel/bench_support.rs` — `RegistryScan::with_runs`
- `src/kernel/registry.rs` — the registry unit row `an_exit_observed_after_the_drain_retires_the_entry`

Adding a field to the record breaks all four, including a test that has no interest in the new field.

## Why this shape

The four literals are identical, and not by coincidence: a run is only ever *recorded* at its start, and every other phase is reached by transition. `phase`, `revoked` and `exited` are therefore not per-site choices — they are what "fresh" means, and nothing a spawn site holds decides them — so `RunEntry::running(token, kind, scope, counter, gate, abort)` takes only the six the spawn site does hold and fills the three state fields itself.

It is a convention its call sites keep, not a barrier: the fields stay `pub` because `pass` filters on `phase` and `testing::driver` reads `exited`, so the literal remains writable crate-wide and a site that skips the constructor still compiles. What it buys is the one place a field addition has to be taught the fresh shape.

## Not in scope

The issue also notes the registry unit row hand-mints an `AbortHandle` via `JoinSet::spawn(pending())` because `bench_support`'s fixture is feature-gated out of `cfg(test)`. The constructor does not change that, and un-gating the fixture is a separate decision.

## Verification

- `just clippy` (`--all-targets`, `build_features`) clean — every target, including the `bench-internals` fixture, type-checks against the new signature.
- `just fmt-check` clean.
- Behaviour is unchanged by construction: the constructor's body is the same field-for-field initializer the four literals spelled.

### On `codecov/patch`

Red, and judged by content per `codecov.yml` (neither codecov status is a merge gate). All 5 missed diff lines are in `src/kernel/bench_support.rs`, which is 0% covered on both base and head: it is `bench-internals`-gated and reachable only from the benches, which the coverage job does not run. `registry.rs` reports every one of its 15 new lines hit and its miss count unchanged at 6; `producer.rs`'s is unchanged at 1. Absolute misses fall 614 → 612 and `codecov/project` reports +0.01%. The patch percentage drops because the diff *rewrites* lines in an already-uncoverable file, not because it adds untested code.

### DIFF
--- a/src/kernel/bench_support.rs
+++ b/src/kernel/bench_support.rs
@@ -47,7 +47,7 @@ use super::Kernel;
 use super::accounting::PendingCounter;
 use super::cleanup::CleanupLedger;
 use super::lane::{GateMode, RunToken, SendGate};
-use super::registry::{Phase, RunEntry, RunKind, ScopeRegistry};
+use super::registry::{RunEntry, RunKind, ScopeRegistry};
 
 /// A command whose spawned run emits one **producer-originated** quit on the
 /// control lane and then ends (RFC 0014 §3.3).
@@ -199,17 +199,14 @@ impl RegistryScan {
                     RunKind::Sub(id)
                 }
             };
-            registry.insert(RunEntry {
-                token: index as RunToken + 1,
+            registry.insert(RunEntry::running(
+                index as RunToken + 1,
                 kind,
-                scope: run_scope(index),
-                phase: Phase::Running,
-                revoked: false,
-                exited: false,
-                counter: Arc::new(PendingCounter::default()),
-                gate: Arc::clone(&gate),
+                run_scope(index),
+                Arc::new(PendingCounter::default()),
+                Arc::clone(&gate),
                 abort,
-            });
+            ));
         }
         Self {
             registry,
--- a/src/kernel/producer.rs
+++ b/src/kernel/producer.rs
@@ -45,7 +45,7 @@ use crate::testing::driver::{AcceptanceRecorder, IntentRecorder};
 
 use super::accounting::PendingCounter;
 use super::lane::{ControlSender, DataSender, IngressHandle, RunToken, SendGate};
-use super::registry::{Phase, RunEntry, RunKind};
+use super::registry::{RunEntry, RunKind};
 
 /// What a producer body receives.
 pub struct EffectCtx<Msg> {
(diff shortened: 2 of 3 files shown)


akiomik:
Backfilling a disposition that was decided but never recorded here. Review round 2 on this PR returned no correctness findings and two non-blocking observations; one was fixed before merge, the other was declined in conversation only, which left it invisible to anyone reading the PR.

**1. "this module's own rows" (plural) where there is one row** — fixed before merge, in the final commit.

**2. `RunEntry::running` is `const fn` but unusable in const context** (no caller can const-construct an `Arc` or an `AbortHandle`), so the marker is decorative, and a future field needing a non-const initializer will force dropping it.

Declined, and not by preference: `clippy.toml`'s lint set runs `missing_const_for_fn`, and `just clippy` passes `-D warnings`, so omitting `const` here fails CI's Clippy job. It was written without `const` first and rejected for exactly that. The observation's forward-looking half stands — a field like an `Instant::now()` default would make the function non-const and the marker would come off then, which is a one-word edit at that point and not a cost carried in the meantime.

No issue filed, because there is nothing to do until such a field exists and the lint would then require the change itself.

Issue #306, opened 2026-08-27
test_compose_hook_restores_then_delegates_once is serialized only against tests that take the hook guard, and 31 #[should_panic] rows do not


akiomik:
## Summary

`panic::tests::test_compose_hook_restores_then_delegates_once` (`src/panic.rs`) installs a recording hook process-wide, panics once on purpose, and asserts the hook ran exactly once. It takes `hook_guard()` to prevent exactly the interference it then suffers:

```rust
// Serialize against other tests that touch the global panic hook or
// panic, so a concurrent panic cannot invoke our recording hook.
let _hook_guard = hook_guard();
```

The guard is a mutex, so it only serializes against tests that also take it. Every `#[should_panic]` row panics on its passing path and takes nothing — there are **31** of them, in `src/testing/driver.rs`, `src/testing.rs`, `src/kernel/accounting.rs` and `src/kernel/conformance/bounded.rs`, and none of those files mention `hook_guard`. Each such panic, if it lands while the recording hook is installed, runs it and appends another `restore`/`next` pair.

## Evidence

The failure says so in its own payload — the recorded vector is the expected one repeated:

```
thread 'panic::tests::test_compose_hook_restores_then_delegates_once' panicked at src/panic.rs:195:9:
assertion `left == right` failed: restore must run exactly once, before the delegated hook
  left: ["restore", "next", "restore", "next"]
 right: ["restore", "next"]
```

and once with three pairs, i.e. two foreign panics inside the window rather than one.

Reproduced on `main` (cac8fa4), whole `--lib` suite at `--test-threads=32` across 8 concurrent processes:

| | failures in 96 runs |
| --- | --- |
| this row | 3 |
| `kernel::conformance::quit::a_discarded_quit_s_dequeue_retires_its_origin_with_the_kernel_still_running` | 1 (unrelated, filed separately) |

Not reachable at CI's default thread count, which is why it has not been seen there. It is reachable by anyone running the suite with a raised `--test-threads`, and the rate rises with it.

## Why the guard cannot fix this by itself

Adding `hook_guard()` to 31 `#[should_panic]` rows would serialize them against each other as well, which is a real cost for rows that have nothing to do with hooks and would grow with every new one. It also leaves the obligation exactly where this class of bug lives — on whoever writes the next panicking test.

Directions that do not have that shape:

[...]

PR #338, opened 2026-09-03
test(observability): share the census's rise half and its settle bound

### LINKED ISSUE #335: Consolidate the INV-LC7 census's rise half and settle-loop bound into tests/common/gauges.rs
## Background

#318 extracted the INV-LC7 producer-gauge census's *settle* half into `tests/common/gauges.rs` (`PRODUCER_GAUGES`, `producer_gauges_are_zero`, `producer_gauge_report`), shared by `tests/lifecycle.rs` and `tests/observability.rs`. The module doc promises "a producer gauge added here reaches every census at once." Two adjacent pieces were left un-shared:

1. **The rise half is still hand-unrolled.** `tests/observability.rs` asserts each gauge rose with three literal `contains(&1)` calls; `tests/lifecycle.rs` runs the same claim as a loop over `expected_active` with `any(|v| v > 0)`. A fourth gauge added to `PRODUCER_GAUGES` is settle-checked automatically but **never rise-checked in `observability.rs`** — and since every snapshot carries all counts, the new never-rising gauge reads `Some(0)` and clears even the strict branch: the silenced-instrument vacuity the census exists to prevent, confined to that file. The two copies also disagree on what "rose" means (`contains(&1)` fails a gauge that jumped straight to 2; `any(>0)` does not).
2. **The settle-loop bound is duplicated.** `tests/observability.rs`'s bare `0..1_000` duplicates `tests/lifecycle.rs`'s `SETTLE_STEPS = 1_000`, whose ~20-line comment carries the load-bearing rationale (the bound counts executor drains, not time; a `sleep`/`timeout` bound would advance the paused clock and hand an uncancelled producer a second way to stop, masking the exact teardown bug these rows catch). The bare literal carries none of that.

## Proposal

- Add a `producer_gauges_rose(recorder, expected_active)` beside `producer_gauges_are_zero` and use it at both targets, so the rise half follows `PRODUCER_GAUGES` too.
- Move the settle-loop bound constant and its rationale into `tests/common/gauges.rs` (both targets already include it).
- Optionally reject an empty `expected_active` in `producer_gauges_are_zero` (documented today as "a no-producers reading, not an INV-LC7 check", but not enforced).

## Why deferred from #318

#318 shared the settle predicate; sharing the rise half and the bound is a follow-on cleanup, latent today (no fourth producer gauge exists).

akiomik:
Closes #335.

## What

The INV-LC7 producer-gauge census (`tests/common/gauges.rs`) shared its field set (`PRODUCER_GAUGES`) and its settle predicate (`producer_gauges_are_zero`), but not the two pieces on either side of them. This adds `producer_gauges_rose` and moves `SETTLE_STEPS` into the census, so both including targets — `tests/lifecycle.rs` and `tests/observability.rs` — read one definition of each.

## The rise half was the live gap

`observability.rs` asserted three literal `contains(&1)` calls; `lifecycle.rs` looped over `expected_active` with `any(> 0)`. A gauge added to `PRODUCER_GAUGES` was therefore settle-checked automatically and rise-checked **nowhere** in `observability.rs` — and since every gauge event carries all the counts at once, a never-rising gauge reads `Some(0)` on some other producer's event and clears even the strict settle branch. That is exactly the silenced-instrument vacuity the census exists to prevent.

Reproduced by adding RFC 0006 §4.4's fourth count `blocked` to `PRODUCER_GAUGES` (it rides the same snapshot and stays at zero over this run):

- **Before:** `producer_gauges_rise_and_fall_over_a_run` **passes**, with `blocked` reading `[0, 0, 0, 0, 0, 0]`.
- **After:** the row **fails** — `the blocked gauge must have risen while this row's producers ran, or its fall to zero witnesses nothing: [0, 0, 0, 0, 0, 0]` — and `#[track_caller]` names `tests/observability.rs:104`, the caller, not the helper.

`producer_gauges_rose` takes the set as an argument, so both halves follow `PRODUCER_GAUGES` together.

## The membership guard

`producer_gauges_rose` carries the same census-membership assert as `producer_gauges_are_zero`, and the rise direction needs it more. `u64_values` reads a field name off the whole `tears::runtime::load` target, where every gauge event also carries `seq` and `runtime_id` — both above zero by construction. So a name that misses the census does *not* read empty and fail; it can read one of those and **pass**, asserting that the ordering counter rose. Verified: without the guard `producer_gauges_rose(&recorder, &["seq"])` passes on an empty recorder's target; with it, it panics naming the census.

## On "rose"

The two copies disagreed: `contains(&1)` fails a gauge that jumped straight to 2, `any(> 0)` does not. Settled toward `any(> 0)`. The forms agree for a counter that moves by one per producer and emits on every move, so they can differ only where an emission was lost — and there the exact-value form fails for the wrong reason, reporting a gauge that plainly ran as one that never started. How many producers ran is a different claim, made by the unit-layer rows that pin exact gauge traces (`an_anonymous_run_raises_and_lowers_the_unkeyed_command_gauge` and its siblings in `src/kernel/producer.rs`).

## The settle bound

`observability.rs` spelled a bare `0..1_000` against `lifecycle.rs`'s `SETTLE_STEPS`, whose comment carries the load-bearing rationale: the bound counts executor drains rather than time, because a `sleep`/`timeout` bound would advance the paused clock and hand an uncancelled producer a second way to stop — masking the exact teardown bug these rows catch. The bare literal carried none of that. The constant moves to the census beside its rationale.

## Not taken from the issue

Rejecting an empty `expected_active` in `producer_gauges_are_zero`. The census's own unit row (`an_active_gauge_with_no_events_is_not_settled`) passes `&[]` deliberately, to pin the tolerant branch a no-producers reading needs, so an assert there would fail a row that documents the behaviour.

## Verification

- `just clippy` (`--all-targets`, `build_features`) clean, `just fmt-check` clean.
- `cargo test --test lifecycle --test observability`: 15 + 3 passed, 0 failed.
- The `blocked` mutation above, run in both directions.
- The `seq` guard probe above, run in both directions.

### DIFF
--- a/tests/common/gauges.rs
+++ b/tests/common/gauges.rs
@@ -1,8 +1,9 @@
 //! The INV-LC7 producer-gauge census, shared via `#[path]` so every
 //! integration target that reads the gauges reads one definition of the
-//! field set and one meaning of "settled" — a producer gauge added here
-//! reaches every census at once instead of leaving a copy waiting on the
-//! old set.
+//! field set, one meaning of "rose", one meaning of "settled", and one
+//! bound on how long settling may take — a producer gauge added here
+//! reaches every census at once, on both halves, instead of leaving a copy
+//! waiting on the old set.
 //!
 //! An including target must also declare the recorder at its crate root,
 //! spelled exactly `#[path = "common/trace_recorder.rs"] mod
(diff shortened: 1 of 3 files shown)


akiomik:
Review round 2 closed with no correctness findings. Two low-severity observations were raised, both latent-only and neither reachable from a call site that exists — `producer_gauges_rose`'s vacuous pass on an empty `expected_active` (the one caller that builds the slice dynamically seeds it with `"keyed_commands"`), and the membership guard having no row of its own. Both are scoped out of this PR and filed as #339, where the `#[should_panic]`-per-including-binary cost can be weighed against #306 rather than settled in passing here.

PR #341, opened 2026-09-03
test(conformance): size the handshake bound against the failing path

### LINKED ISSUE #305: Conformance fixtures: HANDSHAKE_YIELDS is sized for a cost that does not hold on the path it guards, so an unanswered handshake takes ~154 s to report
## Summary

`HANDSHAKE_YIELDS` (`src/kernel/conformance/support.rs`) is the hang guard on `MidBatchHandshake::open_and_await_commit`, and it is 1,000,000 `thread::yield_now()` calls. The bound is sound — nothing structural can release that wait, see below — but the value was chosen against a per-yield cost that does not hold on the path where the bound is actually spent.

```rust
const HANDSHAKE_YIELDS: usize = 1_000_000;
```

## The two costs differ by three orders of magnitude

A yield is cheap when nothing else wants the core and expensive when a peer is blocked. Measured on an Apple M1 Max (10 cores), rustc 1.97.0:

| condition | 1,000,000 `thread::yield_now()` |
| --- | --- |
| release build, idle | 152 ms |
| release build, 8 competing spinners | 2.14 s |
| **debug build, with a blocked peer** | **~154 s** |

The last row is the one that matters, because a blocked peer is the failing path's own condition: the bound is only ever spent when the gated producer never commits, which is exactly when a thread is sitting there not making progress. The 154 s is not extrapolated — it is the observed runtime of a fixture row that spent this constant under those conditions:

```
test result: FAILED. 0 passed; 1 failed; … finished in 154.46s
```

## Why it matters

A producer that never reaches its send is a real defect, and the row that would catch it takes two and a half minutes of a CI job to say so. That reads as a hang rather than a failure, which is the thing a hang guard exists to avoid being mistaken for.

Four rows spend this constant: `src/kernel/conformance/delivery.rs` (one) and `src/kernel/conformance/quit.rs` (three), all through `Script::handshaking_on`.

## Why the bound itself is right

Worth stating so this is not read as an argument for removing it. The quiescence gate in the same file could replace its own counted bound with a structural release, because the party that must release it is the script itself. This one cannot: the party that must signal is the *producer*, running inside a run the script does not own, and a body that parks before its send holds its sender alive — so not even `Disconnected` arrives. A count is what is left, and `clippy.toml`'s `disallowed-methods` (RFC 0009 §3.1, INV-C1) rules out every deadline-shaped alternative.

## What a fix looks like

Pick the value against the failing path's cost rather than the passing path's, since the two are asymmetric: the passing path exits at the commit whatever the bound is, so a smaller value costs a passing suite nothing, and the only thing the value decides is how long a real failure takes to surface. A bound in the tens of thousands would report in single-digit seconds while still being orders of magnitude above what a commit actually takes.

That number wants measuring rather than guessing — the same measurement this issue's table is, but taken at the handshake's own call sites.

## Notes

Found while reviewing #304, which measured the per-yield cost for a different bound in the same file and sized that one against it. Not changed there because it is a different guard with different callers, and no path currently reaches it.

akiomik:
Closes #305.

## The problem

`HANDSHAKE_YIELDS` (`src/kernel/conformance/support.rs`) guards `MidBatchHandshake::open_and_await_commit`. The bound itself is right — the party that must signal is the *producer*, running inside a run the script does not own, and a body that parks before its send holds its sender alive, so not even `Disconnected` arrives. A count is what is left.

The **value** is spent when the gated producer never reaches its send: a blocked peer, which is when a yield costs ~154 µs rather than nothing. That rate is the one this loop's exhaustion was measured at rather than extrapolated to — #305 records a row that spent the whole 1,000,000 taking **154.46 s** to report. A hang guard that slow reads as the hang it exists to tell apart.

`250_000` is a quarter of it: **~38 s**.

## The margin on the other side

A bound set too low fails a correct tree. The margin against that is the yields a *successful* commit spends. A count is not a portable unit, so they come from a table rather than a host — the lesson `HOLD_YIELDS` records in this file's `mod tests`. A temporary probe reported them for the four rows that reach the bound (`delivery`'s one, `quit`'s three) on **two CI runs of every leg** — including the `coverage` job, which runs the same rows under `llvm-cov` instrumentation and was missing from the first table — plus 20 local suite runs:

| leg | first run | second run |
| --- | --- | --- |
| macOS CI, pinned | 1, 1, 1, 1 | 1, 1, 1, 1 |
| macOS CI, beta | 1, 1, 1, 1 | 1, 1, 1, 2 |
| Linux CI, pinned | 1, 1, 8, 13 | 1, 1, 6, 12 |
| Linux CI, beta | 1, 1, 2, 12 | 1, 1, 2, 20 |
| Windows CI, pinned | 1, 1, 1, **155** | 1, 1, 32, 34 |
| Windows CI, beta | 1, 1, 1, 110 | 1, 1, 1, 81 |
| Coverage CI, under `llvm-cov` | (not sampled) | 2, 2, 16, 20 |
| macOS, 10 cores, 20 suite runs | median 2, max 113 | |

Instrumentation does inflate the sampled quantity, as expected — the coverage leg's floor is 2 where every other leg's is 1 — but its maximum lands an order of magnitude under Windows's.

`250_000` is ~1,600x the worst of those, and that ratio is discounted twice over:

- **The denominators differ in kind.** `HOLD_YIELDS`'s 39x is taken over a deliberately contended outlier; this is taken over whatever seven CI legs happened to be carrying.
- **The tail is thin.** 155 is one observation out of two runs of that leg, whose second run peaked at 34. `HOLD_YIELDS`'s doc calls its own 6664 outlier a *thirtyfold* tail; taking the larger of the two readings available for it, a thirtyfold tail on 155 puts the real worst near 4,650 and leaves roughly **54x** of headroom. That is the figure the doc tells a reader to carry, since 1,600x is the one they would otherwise quote.

## The margin stays in counts

Converting it to a duration would need the per-yield rate for the regime a late-but-correct commit sits in — a producer runnable but not yet scheduled — and no measurement of this loop in that regime exists. The rates that do exist put the bound at 7.5 ms (idle) and 105 ms (eight competing spinners); which applies is not established, so the ratio is left as a ratio.

## Why 250,000 rather than the "tens of thousands" #305 names

#305's direction rests on this premise: *"the passing path exits at the commit whatever the bound is, so a smaller value costs a passing suite nothing."* The measurement it asked for retires that premise. A correct-but-late run spends the bound too, and what it has to outlast is a scheduling quantum — `thread::yield_now` is `SwitchToThread` on Windows, reaching only threads ready on the calling processor, and `windows-latest` is Server with a ~187 ms long fixed quantum, so a producer queued behind a long-running thread on another core is reached by that thread's quantum expiring rather than by any number of yields.

That makes the value a threshold, not a slider:

| bound | idle rate | contended rate | vs ~187 ms quantum |
| --- | --- | --- | --- |
| 50,000 | 7.5 ms | 105 ms | under at **both** ends |
| 250,000 | 37.5 ms | 525 ms | over at one end |

`50_000` reports in 7.7 s and clears #305's stated target, but spends its whole budget inside one quantum at either rate. `250_000` reports in ~38 s and crosses at the contended end. The costs are asymmetric — a slow report is paid once by whoever broke the fixture and terminates either way, while a bound that expires on a correct tree is paid by every pull request on a required check.

Most of #305's complaint was about *diagnosis* rather than latency ("reads as a hang rather than as the failure it is"), and the panic rewrite below answers that half at any value.

**The doc states this as a decision rule, not a measurement.** `250_000` is not supported by evidence `50_000` lacks; both are choices under the same unmeasured tail, and the contended rate is borrowed from another host and build profile. It is chosen because where the uncertainty is unresolved, the cheaper thing to be wrong about is the slow report. **#342** carries the measurement that would replace the rule with a number.

## Also in this commit

`HOLD_YIELDS`'s paragraph cited this constant's 154 s as a bound too slow to be useful. It is rewritten so the ceiling stands as a rate calculation rather than a reference to a value that changed in the same commit, and so it says why the two constants land far apart: that one is pinned from below by a contended tail, with its reporting time a consequence rather than a target.

## Verification

- All six CI legs green on every revision of this branch, at both `50_000` and `250_000`.
- `cargo test --lib` (`build_features`) — 589 passed, repeated runs, no flake.
- `just clippy` (`--all-targets`), `just fmt-check`, and `RUSTDOCFLAGS=-D warnings cargo doc` clean.
- The failing path reports correctly at the new bound: with the gate deliberately never opened, the row fails with `the mid-batch handshake was not answered in 250000 scheduler yields. Either the gated producer never committed while the batch was running, or it was going to and this bound ran out first — see HANDSHAKE_YIELDS for which is which`.
- The loop's yield count was verified by instrumentation rather than by reading, at the time the invariant was established: `yields=50000 final_poll_ran=1 bound=50000` on the real failing path. The accounting is independent of the value; the panic text above was re-run at `250_000`.
- The measurement probe was removed before the commit.

## The diff is not doc-only

Worth naming, because an earlier version of this description implied it was. Alongside the constant and its rationale, the commit changes runtime behaviour in three fixtures:

- `open_and_await_commit` polled once per yield and never after the last one, so a commit landing in that final window was discarded and the row reported a failure. The loop now counts down: exactly `HANDSHAKE_YIELDS` yields across one more poll.
- `HOLD_YIELDS`'s two waits had the identical window. The flat one counts down the same way; the stall one now tests its bound *after* the exit check rather than as a `while` condition, which is what lets its last yield be read.
- The exhaustion panic no longer asserts `the gated producer never committed` as fact — at this bound that inference is exactly what the doc says can be false.

## History

The value went 1,000,000 → 50,000 → 250,000 → 50,000 → **250,000** across review rounds, and the rationale was rewritten with it. Three arguments of mine were refuted along the way — one contradicted by this PR's own measurement, one with the regime backwards, and a wall-clock "floor" that turned out to be scheduling noise — and none of them survive in the doc.

The last move is not a return to where the branch had already been. The earlier `250_000` rested on those refuted arguments; the reason now is the quantum threshold, which did not exist in the argument then, and which arrives with #305's "costs a passing suite nothing" premise retired by this PR's own measurement.

Per-round dispositions are in the comments below.

### DIFF
--- a/src/kernel/conformance.rs
+++ b/src/kernel/conformance.rs
@@ -44,13 +44,17 @@
 //! `view` calls the program under test counts, the driver owning its
 //! terminal and reporting nothing about frames (RFC 0008 §9.11).
 //!
-//! **Waiting.** No series sleeps, arms a timer, or reads a wall clock. Every
-//! wait is a bounded number of executor turns and fails the test on its
-//! bound, and both budgets that a script owns — `settle`'s and `confirm`'s —
-//! are named at the call site (RFC 0008 §9.6, §9.8). One wait is the
-//! exception and is documented where it lives: the mid-batch handshake's
-//! `recv` is an untimed block rather than a timed one, because a deadline
-//! would be a clock read.
+//! **Waiting.** No series sleeps, arms a timer, or reads a wall clock, and
+//! no wait can fail to end. Most are a bounded number of executor turns and
+//! fail the test on that bound, with both budgets a script owns —
+//! `settle`'s and `confirm`'s — named at the call site (RFC 0008 §9.6,
+//! §9.8). The mid-batch handshake counts scheduler yields instead, since
+//! the party it waits on runs on another thread, but it is bounded the same
+//! way. The gated-quiescence hold is the second form: it blocks on a
+//! release rather than a count, and what makes it terminate is structural —
+//! the type that drives the release owns it, so a script has no way to omit
+//! it. A deadline is ruled out in all of them for one reason, that it would
+//! be a clock read.
 //!
 //! **Four rows drive a multi-worker executor**, and take their determinism
 //! from an application-side handshake rather than from INV-RC14. A pass is a
--- a/src/kernel/conformance/support.rs
+++ b/src/kernel/conformance/support.rs
@@ -94,12 +94,163 @@ pub const THREADED_TURNS: usize = 4096;
 /// How many scheduler yields the reducer's half of a
 /// [`MidBatchHandshake`] spends waiting for the gated producer's commit.
 ///
-/// A hang guard whose value is mechanism, in the same style as
-/// [`THREADED_TURNS`]: the ordering the handshake establishes does not
-/// depend on how many yields it took to observe the signal, and the bound
-/// only decides whether a producer that never commits fails the test or
-/// stops it.
-const HANDSHAKE_YIELDS: usize = 1_000_000;
+/// A hang guard, in the same style as [`THREADED_TURNS`]: the ordering the
+/// handshake establishes does not depend on how many yields it took to
+/// observe the signal. Its *value*, unlike that one's, is not free. A
+/// passing run leaves this loop at the commit whatever the bound is, so no
+/// value here makes a healthy suite slower; what it decides is how long a
+/// real failure takes to report, and — if set below what a correct-but-late
+/// run can spend — whether a correct tree fails at all. The paragraphs
+/// below are that accounting.
+///
+/// **Reporting a real failure.** A yield is cheap when nothing else wants
+/// the calling core and dear when a peer is blocked on it: about 154 µs
+/// each in a debug build. That rate is the one this loop's exhaustion was
+/// measured at rather than extrapolated to — a row that spent the whole
+/// 1,000,000 was observed taking 154.46 s to report (issue #305) — and a
+/// hang guard that slow reads as the hang it exists to tell apart.
+/// `250_000` is a quarter of it: about 38 s.
+///
+/// **Not failing a correct tree.** The margin here is the yields a
+/// *successful* commit spends. A count is not a portable unit, so they are
+/// taken from a table rather than a host, the reason `HOLD_YIELDS` gives in
+/// this file's `mod tests`. Four rows reach this bound (`delivery`'s one,
+/// `quit`'s three), so a leg contributes four numbers. Two CI runs of each
+/// `test` leg, and one of the coverage leg:
+///
+/// | leg                          | first run    | second run   |
+/// | ---                          | ---          | ---          |
+/// | macOS CI, pinned             | 1, 1, 1, 1   | 1, 1, 1, 1   |
+/// | macOS CI, beta               | 1, 1, 1, 1   | 1, 1, 1, 2   |
+/// | Linux CI, pinned             | 1, 1, 8, 13  | 1, 1, 6, 12  |
+/// | Linux CI, beta               | 1, 1, 2, 12  | 1, 1, 2, 20  |
+/// | Windows CI, pinned           | 1, 1, 1, 155 | 1, 1, 32, 34 |
+/// | Windows CI, beta             | 1, 1, 1, 110 | 1, 1, 1, 81  |
+/// | Coverage CI, under `llvm-cov`| (not sampled)| 2, 2, 16, 20 |
+///
+/// Beside those, 20 local suite runs on a 10-core macOS host: median 2 over
+/// their ~80 observations, maximum 113. That row is a summary rather than
+/// four counts, which is why it is not in the table.
+///
+/// The coverage leg is sampled because it runs these rows too, under
+/// instrumentation, on the same runner class as the Linux legs — the
+/// environment most likely to inflate what is being measured. It shows in
+/// the lower order statistics — a floor of 2 where every other leg's is 1,
+/// and a third value of 16 against 8, 6, 2 and 2 for the Linux samples —
+/// but not at its own top: its maximum of 20 is exactly Linux beta's
+/// uninstrumented maximum from the same run. That is one sample on a runner
+/// class that never produced anything near 155, so it says instrumentation
+/// did not lift the Linux maximum, not that it leaves maxima alone.
+///
+/// `250_000` is about 1,600x the worst of those. `HOLD_YIELDS` takes 39x over
+/// its own worst observed, and the two ratios are not comparable, in two
+/// ways that both cut against this one. Its denominator is an outlier taken
+/// under contention — incidental rather than deliberate, a macOS run that
+/// happened to have a build beside it — where this is whatever seven CI
+/// legs were carrying. (The deliberately contended sampling in that table
+/// is its Linux column, whose maximum is 3511, and the 39x is not over
+/// that.) And its 6664 was measured on the caller that bounds a *stall* —
+/// a count that resets at every progress mark, so an arbitrarily long total
+/// wait passes as long as no single gap exceeds it — where this bounds a
+/// total that never resets. Measured against the more forgiving of the two
+/// shapes, that 39x is the more forgiving 39x, so the larger ratio here
+/// buys less than the arithmetic suggests. (`HOLD_YIELDS` has a second
+/// caller that is a flat total, like this one; the denominator is not from
+/// it.)
+///
+/// That margin is in counts, and it stays in counts deliberately. Turning
+/// it into a wall-clock tolerance would need the per-yield rate for the
+/// regime a late-but-correct commit sits in — a producer runnable but not
+/// yet scheduled — and no measurement of this loop in that regime exists.
+/// The rates that do exist bracket it widely: 0.15 µs idle and 2.1 µs
+/// against eight competing spinners, putting `250_000` at 37.5 ms and
+/// 525 ms — nowhere near the tens of seconds the blocked-peer rate
+/// implies. Two
+/// substitutions sit under those figures, and only one is signed. They are
+/// *release*-build measurements standing in for a debug suite, and the one
+/// debug rate available (154 µs) is three orders above the idle one, so
+/// that direction errs low. They were also taken on an Apple M1 Max, and
+/// `swtch_pri` on Darwin is not `SwitchToThread` on Windows — a
+/// substitution this file has been wrong about before (see `HOLD_YIELDS`:
+/// "sampling one platform and calling the ratio a margin is how this
+/// constant was first set, and it was wrong by an order of magnitude"), and
+/// which could err either way. So the count ratio is left to stand on its
+/// own rather than dressed up as a duration.
+///
+/// What the counts cannot show is a tail, which is the other half of why
+/// the ratio above is not the clearance it looks like. The worst entry is
+/// one observation out of two runs of that leg — the same leg's second run
+/// peaked at 34 — and the sampling that found `HOLD_YIELDS`'s outlier has
+/// no counterpart here: 100 Linux runs at 4x oversubscription, and a macOS
+/// run taken beside a running build. Instrumentation is the only extra load
+/// represented above, and oversubscription is not the same thing.
+/// Read 155 as the worst *sampled*, not the worst that exists. The tree's
+/// one measurement of that gap is `HOLD_YIELDS`'s macOS outlier of 6664,
+/// which stands at 18x its column's uncontended maximum of 373 while its
+/// own doc calls it a thirtyfold tail without saying which member the 30x
+/// is over. Take the larger, since the point is not to flatter the margin:
+/// a thirtyfold tail on 155 puts the real worst near 4,650, leaving
+/// `250_000` with roughly 54x of headroom rather than 1,600x. That is the
+/// figure to carry, because 1,600x is the one a reader would otherwise
+/// quote.
+///
+/// Borrowing that factor is a different move from comparing the two
+/// margins, which the paragraph above declines to do. A margin is a
+/// property of the bound — and a stall bound's is more forgiving than a
+/// total's, so 39x and 1,600x are not the same kind of number. A tail factor
+/// is a property of the *host*: how much worse an unsampled run can be than
+/// a sampled one under the same scheduler. That transfers across bound
+/// shapes in a way the margin does not. It transfers weakly all the same,
+/// since it comes from one macOS column and this bound's worst is a Windows
+/// observation, which is why 54x is the conservative end of a factor the
+/// tree does not pin down rather than a derived quantity.
+///
+/// **Why not smaller.** Issue #305 proposed "tens of thousands", reporting
+/// in single-digit seconds, on the premise that "the passing path exits at
+/// the commit whatever the bound is, so a smaller value costs a passing
+/// suite nothing". That premise does not survive the measurement #305 asked
+/// for, and retiring it is the main thing this accounting establishes.
+///
+/// A correct-but-late run spends this bound too, and what it has to outlast
+/// is a scheduling quantum. A Windows clock interval is ~15.6 ms and a
+/// quantum is a multiple of it — twelve intervals, about 187 ms, under the
+/// long fixed quantum Server defaults to, which is what `windows-latest`
+/// runners are. The mechanism is why a count cannot simply wait that out:
+/// `thread::yield_now` is `SwitchToThread` on Windows, which yields only to
+/// a thread already ready on the *calling* processor, so a producer queued
+/// behind a long-running thread on another core is not reached by yielding
+/// at all. It is reached when that thread's quantum expires, or when this
+/// thread is itself preempted — quantum-scale events that a sub-quantum
+/// bound can run out before.
+///
+/// That makes the value a threshold rather than a slider:
+///
+/// | bound   | idle rate | contended rate | vs ~187 ms quantum |
+/// | ---     | ---       | ---            | ---                |
+/// | 50,000  | 7.5 ms    | 105 ms         | under at both ends |
+/// | 250,000 | 37.5 ms   | 525 ms         | over at one end    |
+///
+/// `50_000` reports in 7.7 s and clears the target #305 names, but spends
+/// its whole budget inside one quantum at either rate. `250_000` reports in
+/// ~38 s and crosses the quantum at the contended end. The two costs are
+/// not symmetric, which is what decides it: a slow report is paid once, by
+/// whoever broke the fixture, and terminates either way, while a bound that
+/// expires on a correct tree is paid by every pull request, on a required
+/// check, indefinitely.
+///
+/// Most of #305's complaint was about *diagnosis* rather than latency — a
+/// bound that slow "reads as a hang rather than as the failure it is" — and
+/// the panic here now names the bound and both of its possible causes,
+/// which answers that half at any value.
+///
+/// **This is a decision rule, not a measurement.** `250_000` is not
+/// supported by evidence that `50_000` lacks; both are choices under the
+/// same unmeasured tail, and the contended rate above is itself borrowed
+/// from another host and build profile. It is chosen because where the
+/// uncertainty is unresolved, the cheaper thing to be wrong about is the
+/// slow report. Issue #342 carries the contended measurement that would
+/// replace this rule with a number, in either direction.
+const HANDSHAKE_YIELDS: usize = 250_000;
 
 /// How many turns a series hands the executor before asserting that
 /// something has **not** happened.
@@ -707,32 +858,48 @@ impl MidBatchHandshake {
     ///
     /// The wait is a counted number of scheduler yields with a poll for the
     /// signal between them, so it ends either at the commit or at its bound,
-    /// and never on a clock. The bound's value is mechanism, like every
-    /// other hang guard here: the *ordering* this establishes does not
-    /// depend on how many yields it took to observe.
+    /// and never on a clock. The *ordering* this establishes does not depend
+    /// on how many yields it took to observe; the bound's value does not
+    /// affect it either way. What the value does decide is the diagnosis
+    /// below — see [`HANDSHAKE_YIELDS`].
     ///
     /// # Panics
     ///
-    /// Panics when the bound is spent with no signal, which means the gated
-    /// producer never reached its send.
+    /// Panics when the bound is spent with no signal. At the bound's current
+    /// size that is *usually* a producer that never reached its send, but it
+    /// is not proof of one: the same exhaustion can come from a correct
+    /// producer that was not dispatched inside the window, which is why the
+    /// message names both.
+    ///
+    /// Panics too when the gate's sender is gone before a signal arrives —
+    /// the gated run torn down without reaching its send — which is a
+    /// distinct failure and says so.
     #[expect(
         clippy::panic,
         reason = "an exhausted bound fails the test, which is what a panic is here"
     )]
     pub fn open_and_await_commit(&self) {
         self.open.store(true, Ordering::SeqCst);
-        for _ in 0..HANDSHAKE_YIELDS {
+        // Counts down so the last pass can read without yielding again:
+        // the body polls *then* yields, so a plain `0..N` leaves its final
+        // `yield_now` in a window nothing looks at, and a commit landing
+        // there would be discarded and reported as a failure. This spends
+        // exactly `HANDSHAKE_YIELDS` yields — which keeps the panic below
+        // honest — across one more poll than that.
+        for remaining in (0..=HANDSHAKE_YIELDS).rev() {
             match self.committed.try_recv() {
                 Ok(()) => return,
-                Err(TryRecvError::Empty) => thread::yield_now(),
+                Err(TryRecvError::Empty) if remaining > 0 => thread::yield_now(),
+                Err(TryRecvError::Empty) => break,
                 Err(TryRecvError::Disconnected) => {
                     panic!("the gated producer's run ended before it reached its send")
                 }
             }
         }
         panic!(
-            "the mid-batch handshake was not answered in {HANDSHAKE_YIELDS} scheduler yields: \
-             the gated producer never committed while the batch was running"
+            "the mid-batch handshake was not answered in {HANDSHAKE_YIELDS} scheduler yields. \
+             Either the gated producer never committed while the batch was running, or it was \
+             going to and this bound ran out first — see HANDSHAKE_YIELDS for which is which"
         );
     }
 }
@@ -1326,9 +1493,15 @@ pub fn step_when_ready<P: Program, B: Backend>(
     source: WakeSource,
     max_turns: usize,
 ) -> StepReport<B::Error> {
-    for _ in 0..max_turns {
+    // Counts down so the last turn's result is read: a plain `0..max_turns`
+    // takes its final turn with nothing polling afterwards, so a pass that
+    // becomes ready *because of* that turn is reported as exhausted.
+    for remaining in (0..=max_turns).rev() {
         if let Ok(stepped) = driver.step_pass(source) {
             return stepped;
+        }
+        if remaining == 0 {
+            break;
         }
         // Exactly one turn: the predicate is false on its first reading and
         // true on the one after the turn.
@@ -1623,9 +1796,15 @@ pub async fn drive_to_ready<E, F>(probe: &ParkProbe, mut future: Pin<&mut F>) ->
 where
     F: Future<Output = Result<Exit, E>>,
 {
-    for _ in 0..TEST_TURNS {
+    // Counts down for the same reason `step_when_ready` does: the last
+    // turn's result has to be polled, or a future that completes on it is
+    // reported as exhausted.
+    for remaining in (0..=TEST_TURNS).rev() {
         if let Poll::Ready(output) = probe.poll(future.as_mut()) {
             return output;
+        }
+        if remaining == 0 {
+            break;
         }
         turn().await;
     }
@@ -1644,9 +1823,16 @@ where
     reason = "an exhausted bound fails the test, which is what a panic is here"
 )]
 pub async fn settle_until(mut condition: impl FnMut() -> bool, what: &str) {
-    for _ in 0..=TEST_TURNS {
+    // Counts down like the other bounded waits here, which for this one is
+    // an accounting fix rather than a coverage one: `0..=TEST_TURNS` already
+    // read after its last turn, but spent `TEST_TURNS + 1` of them while
+    // reporting `TEST_TURNS`.
+    for remaining in (0..=TEST_TURNS).rev() {
         if condition() {
             return;
+        }
+        if remaining == 0 {
+            break;
         }
         turn().await;
     }
@@ -1808,7 +1994,12 @@ mod tests {
     /// failures; the macOS column is 11 runs at default parallelism, ten of
     /// them between 68 and 373 and one at 6664 with a build running beside
     /// them. That single outlier is the number the margin is taken against,
-    /// because a thirtyfold tail is the thing a bound has to survive.
+    /// because a thirtyfold tail is the thing a bound has to survive. (What
+    /// the 30x is taken over is not recorded: 6664 against the ten runs of
+    /// 68-373 is anywhere from 18x to 98x, and 30x falls inside that band
+    /// without matching either end. `HANDSHAKE_YIELDS`'s doc reuses the 30x
+    /// as a tail factor, which errs conservative against the 18x that the
+    /// max-to-max reading would give.)
     ///
     /// A count is not a portable unit, which is why both columns exist: a
     /// `thread::yield_now()` yields the scheduling quantum on Darwin and
@@ -1818,15 +2009,28 @@ mod tests {
     ///
     /// Failing is the only thing a larger value delays, and it is measurably
     /// expensive: a yield costs about 154 µs in a debug build *with a blocked
-    /// peer*, which is the failing path's own condition, so
-    /// [`HANDSHAKE_YIELDS`] would report a regression 154 seconds later — a
-    /// bound that slow reads as the hang it replaces.
+    /// peer*, which is the failing path's own condition. A million yields at
+    /// that rate is 154 seconds of silence before a failure is named, which
+    /// reads as the hang the guard replaces — the ceiling any bound here has
+    /// to sit under.
     ///
-    /// So the value sits far above the worst stall and far below that:
-    /// `262_144` is about 39x over 6664 and 75x over 3511, for a failure
-    /// reported in around 45 seconds. A bound that merely clears the observed
-    /// maximum is how the budget this change removed came to fail in the
-    /// first place.
+    /// It is only a ceiling, not the value: what a stall here has to outlast
+    /// is a whole driver script's drop rather than one producer reaching its
+    /// send, so the number comes from this wait's own table. That is also why
+    /// it lands so far from [`HANDSHAKE_YIELDS`]'s — this one is pinned from
+    /// below by a tail 6664 yields deep, seen under incidental contention,
+    /// and a reporting time is what is left over rather than a target it was
+    /// chosen to hit.
+    ///
+    /// So the value sits far above the worst stall and far below that
+    /// ceiling: `262_144` is about 39x over 6664 and 75x over 3511, for a
+    /// failure reported in around 45 seconds — a figure that implies about
+    /// 172 µs per yield rather than the 154 µs quoted above, at which the
+    /// same bound reports in 40 s. Neither rate has ever been measured for
+    /// *this* wait, which is issue #342's business, so the number is left as
+    /// it was rather than rederived from a rate borrowed off another loop. A
+    /// bound that merely clears the observed maximum is how the budget this
+    /// change removed came to fail in the first place.
     const HOLD_YIELDS: usize = 262_144;
 
     // The gate stores its release rather than signalling it, which is what
@@ -1860,13 +2064,21 @@ mod tests {
         };
         thread::spawn(move || drop(held));
 
-        for _ in 0..HOLD_YIELDS {
+        // Counts down for the reason `open_and_await_commit` does: reading
+        // then yielding leaves the last `yield_now` in a window nothing
+        // looks at, so a mark landing there would be discarded and reported
+        // as a failure. This spends `HOLD_YIELDS` yields across one more
+        // read than that.
+        for remaining in (0..=HOLD_YIELDS).rev() {
             if quiesced.marked() {
                 assert!(
                     gate.entered(),
                     "the dismantling recorded itself on the way in"
                 );
                 return;
+            }
+            if remaining == 0 {
+                break;
             }
             thread::yield_now();
         }
@@ -2054,7 +2266,7 @@ mod tests {
 
         let mut seen = watch_progress.marks();
         let mut stalled = 0_usize;
-        while stalled < HOLD_YIELDS {
+        loop {
             if watch_exited.marked() {
                 let cause = reported
                     .lock()
@@ -2084,6 +2296,14 @@ mod tests {
             } else {
                 seen = marks;
                 stalled = 0;
+            }
+            // Tested after both reads rather than as a `while` condition, so
+            // the last yield's window is read like every other one: a mark
+            // landing there resets the count instead of being lost, and the
+            // loop still spends exactly `HOLD_YIELDS` yields before giving
+            // up.
+            if stalled > HOLD_YIELDS {
+                break;
             }
             thread::yield_now();
         }


akiomik:
Pushed a temporary probe commit (`b8ff17c`, **not for merge**) to take the measurement on every CI platform before settling the value.

Review round 1 pointed out that this PR sized the bound from a single 10-core macOS host, while `HOLD_YIELDS` — the sibling constant in the same file — carries a two-platform table and records explicitly that *"sampling one platform and calling the ratio a margin is how this constant was first set, and it was wrong by an order of magnitude"*. It also takes its margin against a 30x tail outlier rather than a typical run. That standard applies here too, so the single-host sample this PR shipped with is not enough to set the value on.

The probe panics at the commit with the yields actually spent, so each of the six test legs reports its own numbers. Expect this push to be red — that is what it is for. The value, the doc's table, and the probe's removal all land in the amended commit once the numbers are in.

akiomik:
Round 4 was right, and the correction matters enough to state plainly: the mechanism argument I added in round 3 was refuted by this PR's own measurement.

I claimed an unanswered handshake implies a ready peer, hence the expensive yield rate, hence a bound spent over seconds. But the table two paragraphs below it records the *unanswerable* run — gate never opened, producer busy-yielding and therefore ready by construction — spending 1,000,000 yields in 0.30 s on the idle host. That is 0.3 µs/yield: the cheap rate, with a ready peer. The claim was disproved by a number sitting in the same doc, and the `yield_now()` rule I stated to support it contradicted `HOLD_YIELDS`'s own ("returns almost immediately on Linux when the peer is already running").

Both are retracted. The doc now argues the ratio instead, which does hold: the worst commit seen and the bound are the same measurement on the same host, `155r` against `50_000r`, so the per-yield rate cancels and 320x is a wall-clock margin as much as a count one. What that does not cover is `r` differing between the sampled wait and a failing one — benign in the likely direction (a contended host makes yields dearer, buying more time), adverse only if the calling thread's core is idle while the producer's worker is starved on another. Nothing samples that split; the doc says so and names it as the first thing to check should this ever fail on a correct tree.

The "blocked peer" / "spinning peer" vocabulary split on the same 154 µs figure is reconciled too — the rate figure now cites one condition in both places.

On the suggested `250_000`: I have kept `50_000`. The issue asks for "tens of thousands ... single-digit seconds", `HOLD_YIELDS`'s own 39x standard applied to the worst observed here gives ~6,000, and `50_000` clears that eightfold. Happy to take `250_000` instead if the thinner sampling is judged to outweigh that — it is a one-token change and the doc's table and reasoning carry over unaltered.

akiomik:
Round 5 was right on both mediums, and taking them seriously changed the outcome of this PR.

**The "rate cancels" argument is retracted.** The 155 was spent on a wait that *ended at a commit* — the peer became runnable, so the yields were dear — while the bound is spent where the peer is not running and a yield returns at once. Not the same measurement, and the two rates differ in the adverse direction. So the count ratio bounds nothing in wall clock, exactly as you said.

**I then tried to size on a wall-clock floor instead, and the measurement refused to support it.** Driving the unanswerable wait three times at each of three bounds on this host:

| bound | seconds to report |
| --- | --- |
| 50,000 | 0.30, 0.33, 0.09 |
| 250,000 | 0.07, 0.07, 0.07 |
| 1,000,000 | 0.85, 1.47, 1.44 |

Non-monotone — 250,000 faster than 50,000 — because on an idle host that loop's wall clock is scheduling noise, not a yield rate. My earlier "0.30 s at 1,000,000, so ~0.3 us/yield" was a single draw from that noise, and the ~15 ms and ~75 ms floors I derived from it were arithmetic on noise. All of it is out of the doc.

**So the value is 250,000, and the doc no longer claims more than that supports.** It is 1,600x the worst observed count against `HOLD_YIELDS`'s 39x — two orders of magnitude of count margin, spent deliberately to cover a wall-clock margin nobody here has measured. The largest reduction that stays well inside the evidence rather than the smallest the counts would permit: a real failure reports ~4x sooner rather than ~20x, on the reasoning that a slow true failure costs one wait while a false one on a required check blocks every PR.

The measurement that would justify going further — contended sampling of this loop's own per-yield rate, and of a passing commit's cost on an oversubscribed runner, the shape `HOLD_YIELDS`'s decisive numbers came from — is filed as #342, along with the observation that `HOLD_YIELDS`'s own 45 s has never been checked against a rate measured for its own wait.

---

**Correction to an earlier version of this comment.** It closed by calling rounds 3-5 an oscillation and stopping the review loop on that basis. That was wrong on both counts, and is withdrawn.

Oscillation would be the same axis reversing — a change asked for in one round and undone in the next. Nothing here reversed: each round found a *different* defect in an argument I had written carelessly, which is a sequence of inadequate fixes, not a loop with no fixed point. Reaching for a stopping condition to end it was the wrong response.

And the state those rounds produced — a different value and a wholly rewritten rationale — had not been reviewed by anyone when I declared the loop finished. Review is running on it now, and continues until the findings are actually clear.

akiomik:
Review round on `6eaf051` returned 12 findings. Disposition for each, one to one — head is now `1dd6e32`.

**The three that changed the outcome**

1. **HIGH, the parenthetical disclaiming #305's measurement** — **fixed**. It was flatly wrong. #305 records the 154.46 s as the observed runtime of a row here that spent the whole 1,000,000 with a blocked peer; only the per-yield rate came from elsewhere. My local timings failed to reproduce it and I concluded the figure was borrowed, when what had actually failed was my reproduction — an idle 10-core host does not produce a blocked peer. The doc now cites it as this loop's own measurement.
2. **HIGH, the direction argument backwards** — **fixed**. The margin exists for a *false* failure, and there the producer is runnable and about to commit, so the yields are dear and the bound buys more real time, not less. The doc had flattened that into a blanket "adverse". The corrected paragraph is "a correct run that is merely late has the same shape as a failing one".
3. **MEDIUM, `250_000` misses the issue's single-digit-second target** — **fixed**, value back to `50_000` (~7.7 s). With 1 and 2 corrected there is nothing left arguing for the conservative value; it existed only to cover the hazard I had mis-signed.

**The rest**

4. **MEDIUM, "largest reduction that stays far inside the count evidence" is false** — **fixed**; the sentence is gone and the counts now justify `50_000` directly.
5. **MEDIUM, "two orders of magnitude clear" is 1.6, and the baselines are not comparable** — **fixed**; it reads "eight times that clearance" (320x against 39x), and the uncontended-vs-contended difference between the two denominators is stated in the tail paragraph.
6. **MEDIUM, the wall-clock table is an artifact and disagrees with the body's** — **fixed by removal**. You are right that two tables disagreeing 3-5x for the same bound is bad data rather than the "noise" I read it as. The argument no longer rests on any timing I took, so the table and the conclusion drawn from it are both out.
7. **MEDIUM, 155 is single-sampled** — **fixed**; the doc says to read it as the worst *sampled*, not the worst that exists, and names the missing contended row.
8. **MEDIUM, the `HOLD_YIELDS` paragraph anchors to a retired number and the ordering inverts** — **fixed**. At `50_000` the ordering holds again (7.7 s under its ~45 s), so the paragraph states what the figure bounds. Change-history phrasing removed per the project's convention of keeping history in Git; stray mid-sentence break repaired.
9. **MEDIUM, the PR body describes a different value than the head** — **fixed**; body rewritten against `50_000`, with the verification evidence that covers it and a note recording the value's path through this branch.
10. **LOW, "this host cannot measure it" over-strong** — **fixed by removal**, with 6.
11. **LOW, forward reference to a measurement that is not there** — **fixed by removal**, with 6.
12. **LOW, "154 us" vs "154 µs"** — **fixed**; both spellings normalised.

Nothing is being carried silently: 11 fixed in the tree, 1 fixed in this PR's body, none declined, and the contended sampling this cannot supply is #342 rather than an unstated gap.

akiomik:
Review round on `1dd6e32` returned 6 findings. Disposition for each — head is now `1f3a4b3`.

1. **MEDIUM, the cheap/dear rule is stated backwards relative to the conclusion drawn from it** — **fixed**. I had written the rule as a function of the *calling* core's readiness and then concluded the failing path is dear, which does not follow: a parked producer leaves nothing else ready. The rule now reads as #305 states it — cheap when nothing else wants the calling core, dear when a peer is blocked on it — and the 154 µs is presented as the rate this loop's exhaustion was *measured at* rather than as something derived from a rule.

2. **MEDIUM, "seconds of tolerance" applies the blocked-peer rate to the wrong regime** — **fixed**, and this was the worst error in the doc. Against #305's competing-spinners row (2.1 µs) `50_000` is ~100 ms, not seconds — a ~70x overstatement — and the paragraph then contradicted itself about which regime is the cheap one. The claim is withdrawn entirely. The doc now says the margin stays in counts on purpose, that converting it needs a rate for the runnable-but-unscheduled regime which nobody has measured for this loop, and quotes both bracketing rates (0.15 µs idle, 2.1 µs contended) with the explicit note that either would put the bound in the tens-to-hundreds of milliseconds rather than seconds.

3. **LOW, "~24 ms that 155 yields cost" multiplies a Windows count by a macOS blocked-peer rate, for a successful commit the rate provably does not apply to** — **fixed by removal**, with 2.

4. **LOW, "neither table has a contended row" refuted by its own next clause** — **fixed**; it now scopes to this doc's two sources (the CI legs and the local runs), which is what was meant, leaving `HOLD_YIELDS`'s contended Linux column standing as the contrast.

5. **LOW, the opening asserts a one-sided dependence and then retracts it** — **fixed**; it opens on "both paths the value touches" and the two are then headed separately.

6. **LOW, the doc does not cite #342 though the PR body says it does** — **fixed**; #342 is cited by number, matching the #305 citation style already in the doc.

Six fixed, none declined, nothing deferred. The measurement this cannot supply remains #342.

On the value: findings 1-3 removed the wall-clock framing on both sides of the argument, and what is left is the 320x count ratio plus a measured 7.7 s report time. Neither of those moved, so `50_000` stands — but it now stands on the two things that were actually measured rather than on a duration I could not substantiate.

akiomik:
Review round on `1f3a4b3` returned 3 findings, one of them carrying a second item. Disposition for each — head is now `c2257f2`.

1. **MEDIUM, my rewrite broke the antecedent of "that" in the following paragraph** — **fixed**. Before the diff that paragraph ended on the 154 s / 1,000,000 figure, so "far below that" bound to it; my version ended on `HANDSHAKE_YIELDS`, making the nearest referent 50,000, against which `262_144` is 5x *above*. The sentence now names the figure: "far below the 1,000,000 that costs 154 seconds".

   1b. **The same sentence's "around 45 seconds" is 40.4 s by multiplication (262,144 × 154 µs), an 11% overstatement** — **not fixed, deferred to #342**, which already records it ("whose 262,144 reports in ~45 s on the borrowed rate and has never been checked against a rate measured for its own wait"). The reason for declining rather than editing: I cannot tell from the tree whether 45 s was derived from the borrowed rate or observed for that wait. If it was observed, replacing it with the multiplication would substitute a derivation for a measurement — the exact error this PR spent several rounds undoing. #342 is where that gets resolved, by measuring rather than by arithmetic.

2. **LOW, "tens to hundreds of milliseconds" is not what the cited rates give** — **fixed**. 50,000 × 0.15 µs = 7.5 ms is single-digit, and × 2.1 µs = 105 ms only just reaches "hundreds". The doc now states both products outright rather than a range that rounds them outward.

3. **LOW, the 320x/39x comparison is made without its caveat, which appears 20 lines later** — **fixed**. The caveat is now attached where the comparison is made: the denominators are not the same kind of number, since `HOLD_YIELDS`'s is a deliberately contended outlier and this one is whatever six CI legs happened to be carrying, so the larger ratio buys less than the arithmetic suggests. The later tail paragraph was retrimmed so it extends that point instead of repeating it.

Three fixed, one declined with the reason and the place it is tracked. Nothing carried silently.

akiomik:
Review round on `c2257f2` returned 2 findings. Disposition — head is now `7855fbc`.

**1. MEDIUM, the twentyfold cut is also a twentyfold cut in tolerance for a correct-but-late run, in the regime the doc says it cannot bound — declined, with the risk written into the doc and the resolution tracked in #342.**

The mechanism is right and I am not disputing it. In the cheap-yield regime `50_000` is single-digit milliseconds of descheduling where `1_000_000` was ~150 ms, and no sampling here reaches that scenario.

Declined because reversing it means overriding the direction #305 settled. That issue's framing is explicit — "the passing path exits at the commit whatever the bound is, so a smaller value costs a passing suite nothing, and the only thing the value decides is how long a real failure takes to surface" — and it asks for "a bound in the tens of thousands [that] would report in single-digit seconds". Choosing a larger value trades a measured 20x improvement in reporting time against a hazard nobody has observed, on my judgement rather than on evidence. I did exactly that earlier in this PR (`250_000`), and the reasoning I used to justify it turned out to be two errors, both caught in the round after.

What has changed instead: the doc now names the risk as accepted rather than only noting the tail is unsampled — the twentyfold reduction in late-run tolerance is stated in those terms, with the millisecond figure, and with why it is being accepted. It is no longer something a reader has to derive.

The two suggested alternatives are recorded in #342 rather than dropped:
- **A larger count until #342 lands** — that is the value question above, and #342 is what settles it with a measurement.
- **A bounded park/backoff after the first N yields** — genuinely better than either value, because it decouples the two margins instead of trading them. It conflicts with `open_and_await_commit`'s contract that the wait ends "never on a clock", so it needs that contract revisited rather than a value edit; I have added it to #342 as a design option with that caveat.

**2. LOW, the `HOLD_YIELDS` ceiling now anchors on a value no longer in the tree, the disqualifying clause was dropped, and the two constants' acceptability thresholds diverge unexplained — fixed.** The ceiling is now a rate calculation that stands alone ("a million yields at that rate is 154 seconds of silence... which reads as the hang the guard replaces"), with no dependence on any constant's current value. The paragraph also now says why the two land far apart: `HOLD_YIELDS` is pinned from below by a contended tail 6664 yields deep, so its reporting time is a consequence rather than a target it was chosen to hit.

Also recorded in #342 while there: `HOLD_YIELDS`'s "around 45 seconds" is 40.4 s by the same multiplication, an 11% overstatement. Not corrected here for the reason given in the previous round — the tree does not say whether that figure was measured or derived, and substituting arithmetic for a possible measurement is the error this PR spent several rounds undoing.

One fixed, one declined with the risk documented in the tree and both alternatives filed. Nothing carried silently.

akiomik:
Review round on `7855fbc` returned 5 findings. Disposition — head is now `ae73470`.

1. **MEDIUM, the cut leaves ~7.5 ms of cheap-regime tolerance where 1,000,000 left ~150 ms, in a window nothing samples** — **declined, unchanged from the previous round**, and the review itself scopes it the same way ("a taken risk the doc names, not a defect"). The reasoning is in the doc and in the comment above: #305 settled the direction, and choosing a larger value would trade a measured 20x improvement against a hazard nobody has observed. #342 holds both the measurement that would settle it and the bounded-park design that would remove the trade instead of tuning it. Nothing new to weigh this round, so nothing changed.

2. **LOW, "320x" overstates the clearance once the tree's own contended precedent is applied** — **fixed**, and this is the most useful number anyone has put on this PR. `HOLD_YIELDS`'s macOS tail is 6664 against an uncontended max of 373, about 18x; applying that factor to 155 puts the real worst near 2,800 and leaves roughly **18x** of headroom, not 320x. The doc now carries that calculation and says explicitly that 18x is the figure to quote, because 320x is the one a reader would otherwise take away. It is also in the commit message and the PR body.

3. **LOW, #342 is stale in three ways and the doc points readers at it** — **fixed**. #342's body said the PR lowered the value to 250,000 (it is 50,000), asserted the cheap-regime premise this PR retracted, and claimed both constants' reporting times are extrapolated when this loop's is measured. Rewritten: the three per-yield rates are tabled with which one is measured for what, the retracted premise is gone, the tail factor from finding 2 is included, and the bounded-park option and the `HOLD_YIELDS` 45 s / 40.4 s discrepancy are folded into the body rather than left in a comment.

4. **LOW, the PR body still argued two claims the shipped doc declines** — **fixed**. The body had kept "seconds of tolerance" and "eight times that clearance" above the history section that says they were retracted. Rewritten against what actually ships, including the 18x figure.

5. **LOW, two reflow artifacts** — **fixed**; both lines rejoined. A scan for over- and under-length doc lines across the rewritten block is clean.

Four fixed, one declined on unchanged grounds. The declined item is documented in the tree, in the PR body, and in #342.

akiomik:
Review round on `ae73470` returned 3 findings. All three fixed — head is now `e1f23ba`.

1. **The `coverage` leg runs these same four rows under `llvm-cov` and was not in the table** — **fixed by measuring it**, since it was the leg most likely to inflate the sampled quantity and the one omitted. A second probe run reported it, and the hypothesis holds in direction: its floor is 2 where every other leg's is 1. Its maximum is 20, an order of magnitude under Windows's 155, so the worst observed does not move.

   The same run also gave a second sample of all six Test legs, so the table is now two CI runs of every leg rather than one. The second Windows-pinned run peaked at **34** against the first's 155, which is the most useful thing the extra run produced: it shows 155 is a tail observation rather than that leg's normal, and the doc now says so where it quotes it.

2. **The doc's 18x and the file's own "thirtyfold" describe the same 6664 outlier with different denominators** — **fixed**, taking the larger, because the point of the exercise is not to flatter the margin. A thirtyfold tail on 155 puts the real worst near 4,650, leaving roughly **11x** of headroom rather than the 18x I had or the 320x the raw ratio suggests. 11x is now the figure the doc tells a reader to carry.

3. **"around 45 seconds" sits two lines under its own contradicting derivation** — **fixed as documentation, not as arithmetic.** The paragraph now records that the figure implies ~172 µs per yield rather than the 154 µs quoted beside it, and that at 154 µs the same bound reports in 40 s. The number itself is unchanged, because neither rate has ever been measured for *that* wait — that is #342's business, and substituting a borrowed-rate derivation for a possible measurement is the error this PR spent several rounds undoing.

   Worth flagging on my side: my first attempt at this fix referenced a "rather than 11" phrase as being *below* it in the file. That phrase exists only in the commit the review traced the figure to, not in the tree. I caught it before pushing, but it is the same class of unverified-reference error the earlier rounds kept finding, so I would rather record it than let it pass unremarked.

Three fixed, none declined, nothing deferred beyond what #342 already carries.

akiomik:
Review round on `e1f23ba` returned 5 findings. All five fixed — head is now `d822c5a`.

1. **MEDIUM, the risk should be stated against the scheduler's granularity, not as a count** — **fixed**, and this is the sharpest form the risk has been given in any round. The doc now says it: 7.5 ms is the same order as a single Windows quantum (~15.6 ms) where 1,000,000 left many of them, and `thread::yield_now` is `SwitchToThread` there, which reaches only threads ready on the *calling* processor — so a producer queued behind a long-running thread on another core waits for that thread's quantum rather than for any number of yields, and a bound shorter than a quantum can expire before the producer is ever dispatched.

   It also now says plainly why the table cannot argue against this: every sample in it is a run where the producer *was* dispatched promptly, which is what a commit at 155 yields means, so "nothing sampled comes near the bound" is not evidence about a mode ordinary load does not reach. The two things that do sit against it are stated as what they are — the release/debug substitution errs low, and the Windows legs have run green at this bound on every revision carrying it.

2. **LOW, "two CI runs of every leg" contradicted by the coverage row's `(not sampled)`** — **fixed**; the prose now says two runs of each `test` leg and one of the coverage leg, and the row carries its single run rather than an empty cell.

3. **LOW, the `||` colspan does not exist in GFM, so the local row renders as a once-sampled CI leg** — **fixed** by taking the local row out of the table entirely. It was the only row that was a summary over ~80 observations rather than four raw counts, so it now sits in prose beside the table where that difference is visible.

4. **LOW, "an order of magnitude under Windows's" for 20 vs 155** — **fixed**; it is 7.75x and now reads "about eight times". Fair catch on a doc that is otherwise careful not to round its own margins the flattering way.

5. **LOW, the wall-clock bracket is release-build while the suite is debug** — **fixed**, including the direction, which is the part that mattered: the substitution errs *low*, since the only debug rate available is three orders above the release ones, so the real tolerance is larger than 7.5 ms by an unmeasured factor. That partly offsets finding 1, and the doc now says so rather than leaving a reader to work out which way it cuts.

Five fixed, none declined, nothing deferred beyond what #342 already carries.

akiomik:
Review round on `d822c5a` returned 5 findings. All five fixed — head is now `d4d5890`. One of them changes the input to the value decision, so it is flagged for the maintainer at the end rather than buried.

1. **MEDIUM, `~15.6 ms` is the clock interval, not a quantum, and `windows-latest` is Server** — **fixed**, and it was my error. A quantum is a multiple of the interval: twelve of them, ~187 ms, under the long fixed quantum Server defaults to. So 7.5 ms is between a quarter and a twenty-fifth of a quantum on the exact leg the paragraph singles out, not "the same order as" one — I understated the hazard by 4-25x while writing the paragraph whose purpose was to state it honestly. The mechanism sentence is corrected too: the producer is reached when the holder's quantum expires *or* when the reducer's own thread is preempted and frees its processor, both quantum-scale.

2. **LOW, the paragraph disqualifies "not observed" and then rests on it twice** — **fixed**, and this one compounds the first. The green CI runs are the same population as the table, so counting them was the argument I had just ruled out; "a hazard nobody has observed" was the same thing in one phrase. Both are gone. What survives is a single ground: the release-for-debug substitution errs low.

3. **LOW, "splitting it would trade a measured 20x improvement" is a false dichotomy** — **fixed**. `250_000` is a fourfold reporting improvement with five times the tolerance, so the trade is 20x against 4x. This branch carried that value for a round, so the doc had no excuse for the framing.

4. **LOW, "three orders above them" holds for one bracket endpoint only** — **fixed**: three orders above the idle rate, nearly two above the contended one.

5. **LOW, the instrumentation claim rests on the floor while the reassurance uses the wrong baseline** — **fixed**. The same-platform comparison is Linux beta's second run, whose maximum is also exactly 20, so instrumentation moves the floor and leaves the maximum — the only statistic this bound is sized against — untouched. Comparing 20 against Windows's 155 measured the platform gap, not the instrumentation gap.

---

**For the maintainer, since findings 1 and 2 together move the ground under the value.** When `50_000` was chosen, the doc said the tolerance was "the same order as" a quantum and listed three grounds for accepting the risk. After this round it says the tolerance is a quarter to a twenty-fifth of a quantum on the Windows legs, and two of the three grounds were circular. The remaining ground is that the 7.5 ms is a release figure standing in for a debug one and so errs low by an unmeasured factor.

I have not changed the value, because #305 states the direction and asks for "tens of thousands ... single-digit seconds", and reversing that on my own reading is what produced the `250_000` detour earlier in this branch. But this is the point where the evidence has moved enough that the call is worth making deliberately rather than inheriting. `250_000` costs 4x instead of 20x and buys 5x the tolerance; it is a one-token change and every table and argument in the doc carries over unaltered.

akiomik:
Review round on `d4d5890` returned 6 findings. All six fixed — head is now `08d960e`.

1. **MEDIUM, the value choice should be explicit, and the doc states the 20x/4x trade without saying why 4x is insufficient** — **fixed as documentation; value unchanged.** The gap was real: I had stated the trade and stopped there. The doc now gives the reason — `250_000` puts a real failure at ~38 s, and the complaint #305 files is that a bound slow enough to look like a hang is not a useful guard, asking for single-digit seconds by name; 38 s does not obviously clear that bar. It also names this as the judgement most worth overruling if #342's measurement lands the other way, so a later reader sees where the decision is soft rather than having to reconstruct it.

   The value question itself was surfaced to the maintainer in the previous round's comment and remains open there. This finding says "if the maintainer has already settled it, this is a no-change finding" — it is settled by #305's text, and now recorded as such.

2. **LOW, the coverage row is typeset in the first-run column while the prose says "the same run" as Linux beta's second** — **fixed**; the sample moves to the second-run column where it was actually taken, with `(not sampled)` in the first. The prose and the table now agree, and the earlier round's complaint about `(not sampled)` is answered by the sentence above the table saying two runs of each `test` leg and one of coverage.

3. **LOW, "shows up in the floor and nowhere else" is contradicted by the table** — **fixed**. Coverage's third value is 16 against 8, 6, 2 and 2 for the Linux samples, so instrumentation lifts the body of the distribution, not just the floor. Only the maximum coincides, and that is now what the sentence claims.

4. **LOW, a 118-character doc line** — **fixed**, reflowed. (A scan of the whole rewritten block for over-length lines is otherwise clean; the one remaining 81-character line in this file is outside this diff.)

5. **LOW, "the substitution errs low" asserted twice, the second time as independent support** — **fixed**; the first occurrence keeps the rates and drops the conclusion, which is drawn once, in the risk paragraph where it does work.

6. **LOW, the two docs characterise the same 6664 outlier differently** — **fixed** in `HOLD_YIELDS`'s own line, which now says which pairing its "thirtyfold" means: against the run of ten, where the same outlier is nearer 18x against that column's maximum, and that `HANDSHAKE_YIELDS`'s doc takes the larger.

Six fixed, none declined. The only thing still open on this PR is the value itself, which is the maintainer's call and is flagged in the previous comment and in the doc.

akiomik:
Review round on `08d960e` returned 6 findings. Five fixed, one declined and standing — head is now `dd17333`. This round found the first things in the PR that were wrong in the *tree* rather than in the prose, and they were all consequences of the change that I had not followed through.

1. **MEDIUM, the value** — **declined, unchanged**, and this is the second consecutive round to raise it. The framing here is the sharpest yet: a judgement about report readability traded against a required check that would block every merge if it flakes. That is a real asymmetry and I am not dismissing it.

   Still declining for the reason recorded twice above: #305 states the direction and asks for single-digit seconds by name, and overriding it on my own reading is what produced the `250_000` detour earlier in this branch. The doc now says where the judgement is soft and names `250_000` as the first thing to overrule if #342's measurement lands the other way. It is a one-token change either way and every table in the doc carries over. **Flagging for the maintainer: two independent rounds have now recommended reconsidering it.**

2. **MEDIUM, the exhaustion panic asserts a cause this PR makes unsound** — **fixed**, and this was the most valuable finding in the round. The message said `the gated producer never committed while the batch was running`. At the old bound that inference was safe; at this one it is exactly what the doc admits can be false, so a Windows flake would have told the engineer triaging it, as fact, to go hunt a kernel ordering bug that does not exist. It now names the bound as what was reached and both causes as what might have caused it. Verified by driving the failing path: the new text is what prints.

3. **LOW/MEDIUM, the `# Panics` clause carries the same equivalence** — **fixed** with 2, in the same paragraph pair.

4. **LOW, the constant's opening paragraph and the method doc still carry the pre-change model** — **fixed**. Both said the bound's value is "mechanism" and decides only whether a non-committing producer fails or hangs the test, which is the claim the section beneath it spends a page retiring. A reader who stops at the first paragraph was getting the old model.

5. **LOW, off-by-one: the last yield's window is never polled** — **fixed**, `0..=HANDSHAKE_YIELDS`. A commit landing during the final yield was discarded and reported as a failure. Negligible at this size, but the subject of this PR is margin precision, and it costs one `try_recv`.

6. **LOW, the `HOLD_YIELDS` cross-check names the wrong incomparability** — **fixed**, and the point is better than the one I had. `HOLD_YIELDS` bounds a *stall* that resets at every progress mark, so an arbitrarily long total wait passes it; this bounds a total that never resets. Its 39x is structurally the more forgiving 39x, which cuts against this margin rather than for it. The doc now gives both incomparabilities instead of only the contended-denominator one.

akiomik:
Review round on `dd17333` returned 3 findings. All three fixed — head is now `aa80e27`.

1. **MEDIUM, `0..=` does not deliver the invariant its comment claims** — **fixed**, and the catch is exactly right. The loop body reads *then* yields, so `0..=N` gives N+1 of each and the last `yield_now` is still followed by nothing. I moved the window one iteration later and wrote a comment saying I had closed it. The extra poll now sits after the loop, which is the only place it can be one *more* than the yields rather than one more of both.

2. **LOW, the panic's yield count was off by one against what the loop spent** — **fixed by 1**, and verified rather than reasoned: instrumenting the loop to count executed yields against the interpolated bound on the real failing path reports `actual_yields=50000 bound=50000`. Worth doing given the finding was that the file's load-bearing unit had drifted from what the code does.

3. **LOW, the "thirtyfold" provenance I asserted does not reach** — **fixed**. I wrote that the 30x is "against the run of ten", but 6664 against that run of 68-373 spans 18x to 98x and no member gives 30x. Both docs now say what is actually known: the outlier is 18x its column's maximum, its own doc calls it a thirtyfold tail without recording what over, and `HANDSHAKE_YIELDS` reuses the 30x as the conservative choice. The 11x headroom is now labelled as the conservative end of a factor the tree does not pin down, rather than as a derived quantity.

Three fixed, none declined. The value itself was explicitly not flagged this round; it remains the open item recorded in the two comments above and in the doc.

Worth noting what this round was: every finding was in code or claims I added in the *previous* round while fixing that round's findings — the off-by-one fix that did not fix it, the count it desynchronised, and the provenance I invented while reconciling two figures. That is a fair description of most of this PR's later rounds, and it is why the verification for finding 2 is an executed measurement rather than an argument.

akiomik:
`codecov/patch` went red on `aa80e27` at 0.00% (0 hits, 4 misses), and the content is worth stating rather than waving through, since `codecov.yml` says neither status is a merge gate precisely so that a red mark gets judged.

All four missed lines were the post-loop `match` I had just added: it only executes when the loop exhausts, which only happens on the failing path, which no test exercises — a hang guard's exhaustion is not something the suite is meant to reach. So it was 0% by construction, not by omission.

It also pointed at something real. Appending a second `match` after the loop duplicated the `Disconnected` panic arm and put four uncoverable lines in the tree to close a window worth one poll. Folding the extra poll into the loop instead — counting down, so the last pass reads without yielding again — gets the same invariant with:

- the duplicated `Disconnected` arm gone,
- the diff five lines shorter,
- **one** uncovered new line (`Err(Empty) => break`) instead of four; the other two uncovered lines in that region predate this PR.

Verified by instrumentation rather than by reading, since the previous attempt at this was wrong in exactly that way: driving the failing path reports `yields=50000 final_poll_ran=1 bound=50000` — the bound spent exactly, the extra poll executed, and the panic's count honest.

Head is now `f8828d2`.

akiomik:
Two review results landed together — one on `f8828d2` and one on the previous `aa80e27`. Deduplicated, six items were still live. All six addressed; head is now `505385a`.

**From the `f8828d2` review**

1. **`HOLD_YIELDS`'s two waits have the identical poll-then-yield window this PR just closed** — **fixed in the same commit**, rather than filed. Leaving a known off-by-one two hundred lines from the one I had just removed, in the same file, is the sort of debt that never gets paid. The flat wait counts down like `open_and_await_commit`; the stall wait now tests its bound *after* the exit check instead of as the `while` condition, which is what lets its last yield be read like every other one.

   My first attempt at that second one was wrong and I caught it before pushing: I appended a bare `if watch_exited.marked() { return; }` after the loop, which would have skipped the assertions the in-loop branch makes about the script thread not having panicked. Restructuring the loop keeps one copy of that branch.

2. **The PR body quoted the old panic text, so the shipped message had never been observed failing** — **fixed**. The body now carries the actual current output, plus the instrumented `yields=50000 final_poll_ran=1 bound=50000` reading.

3. **#342 carried a different tail factor (18x → ~2,800 → ~18x headroom) than the doc (30x → ~4,650 → ~11x)** — **fixed in #342**, which now gives the range and says the doc takes the conservative end, so following the link no longer produces a different number than the one the doc says to carry.

**From the `aa80e27` review**

4. **The ms figures are #305's Apple M1 Max rates, and the doc compares them against *Windows* granularity without flagging the substitution** — **fixed**, and this was the sharpest of the two reviews' doc findings. The file records that sampling one platform and calling the ratio a margin is how `HOLD_YIELDS` was first set wrong by an order of magnitude, and I had signed the release→debug substitution while leaving the Darwin→Windows one silent. Both are now named, with their directions: the first errs low, the second could err either way.

5. **"its 39x covers a stall" characterises the constant when it should characterise the measurement** — **fixed**. `HOLD_YIELDS` has two callers and one of them is a flat total exactly like this bound; the 6664 denominator comes from the resetting one. The sentence now says that, and notes the other caller.

6. **The PR body implied a doc-only diff** — **fixed**; it now has a section naming all three behavioural changes, since that is the most review-worthy part and a reader taking the old description at face value would have skipped it.

Also carried from the earlier round: the duplicated `Disconnected` literal is gone, as a side effect of folding the extra poll into the loop rather than appending a second `match`.

Six addressed, none declined. The value remains the one open item, flagged in three comments above.

akiomik:
`codecov/patch` is red at 50% on `505385a` (5 hits, 5 misses). Judging it by content, as `codecov.yml` asks, since neither status is a merge gate.

The changed lines that are uncovered are these three, and nothing else:

```
support.rs:878   Err(TryRecvError::Empty) => break,     open_and_await_commit
support.rs:2046  break;                                 HOLD_YIELDS, flat wait
support.rs:2257  break;                                 HOLD_YIELDS, stall wait
```

All three are the bound-exhaustion arm of a hang guard. Reaching one means a row failed, so a passing suite cannot execute them by construction — the same reason their neighbouring `panic!`s were already uncovered before this PR.

They exist because closing the unpolled-final-window bug needs the loop to distinguish "out of budget" from "keep going", and that distinction is a branch. The alternative shape — a trailing `match` after the loop — was worse on this axis too: it put four uncoverable lines in the tree instead of one, and duplicated the `Disconnected` arm.

`codecov/project` passes at 95.65%, and the absolute miss count rises by 7 lines across the three restructured loops.

akiomik:
Review round on `505385a` returned 4 findings. Three fixed, one declined and standing — head is now `a558bb4`.

1. **MEDIUM, the value** — **declined, unchanged.** Third consecutive round to raise it, and the mechanism is stated more concretely each time. Recorded and flagged for the maintainer in three comments above; the doc names it as a risk taken and #342 is the resolution path. Nothing new to weigh this round.

2. **LOW, my stall-loop fix was incomplete against its own comment** — **fixed**, and the catch is exact. I moved the bound test after the *exit* check but left it before the `watch_progress.marks()` read, so the final window read one of the two marks and not the other: a script publishing its progress mark during that last yield would have been reported as `"262144 scheduler yields without progress"` immediately after making progress. The bound is now tested after both reads, and the loop still spends exactly `HOLD_YIELDS` yields.

3. **LOW, a permanent comment embedded this PR's review history** — **fixed**, and thank you for catching it: `` `250_000`, which this branch carried for a round `` is unresolvable from `main`, which is the class this repo has had to correct four times. It now reads as a plain value comparison. A grep for branch/PR/round references across the file comes back empty.

4. **LOW, two siblings keep the defect class this PR names** — **fixed rather than filed.** `step_when_ready` and `drive_to_ready` each took their last turn with nothing polling afterwards, so a pass that became ready *because of* that turn was reported as exhausted. Both count down now. Writing down a rule and leaving two violations of it in the same file is worse than not writing it down, and the whole point of the earlier rounds was that this class is easy to miss.

So the defect this PR started from turned out to run through the file, and all five instances are closed: `open_and_await_commit`, `HOLD_YIELDS`'s flat and stall waits, `step_when_ready`, `drive_to_ready`. (`settle_until` already used `0..=` and was fine.) All are test fixtures, none changes what a row asserts — only whether a result arriving in the last window is seen or discarded.

Verified with the full integration suite alongside `--lib`, since this round touched shared driver helpers rather than one loop: 589 lib rows and every integration target green.

akiomik:
`codecov/patch` red again on `a558bb4`, and the content is the same shape as before, now across five loops instead of one. The uncovered changed lines are exactly these:

```
support.rs:878        Err(TryRecvError::Empty) => break,   open_and_await_commit
support.rs:1488-1490  if remaining == 0 { break; }         step_when_ready
support.rs:1792       break;                               drive_to_ready
support.rs:2058       break;                               HOLD_YIELDS, flat wait
support.rs:2283       break;                               HOLD_YIELDS, stall wait
```

Every one is the budget-exhaustion arm of a bounded wait. Executing it means the row failed, so a passing suite cannot reach any of them — the same reason the `panic!`s they fall through to were already uncovered before this PR.

They are the cost of the fix: closing the unpolled-final-window bug requires the loop to tell "out of budget" from "keep going", and that distinction is a branch. The alternative shape for the two turn loops — poll once before the loop and keep it at N — avoids the branch but duplicates the poll expression, which is the trade this PR already made the other way once (folding the extra `try_recv` into `open_and_await_commit`'s loop rather than appending a second `match`, which cut four uncoverable lines to one and removed a duplicated diagnostic literal).

`codecov/project` passes.

akiomik:
Review round on `a558bb4` returned 4 findings. Three fixed, one declined and standing — head is now `64b5fc0`.

1. **MEDIUM, the bound can be spent inside one Windows Server quantum** — **declined, unchanged.** Fourth consecutive round on this, and the arithmetic here is the tightest statement of it yet: even at #305's eight-spinner rate the whole budget is ~105 ms against a ~187 ms quantum, so both bracket rates land under it. Nothing about that is disputed and the doc says it. Still declining for the reason recorded each time — #305 settled the direction and the alternative trades a measured 20x for an unmeasured hazard — and it remains the maintainer's call, flagged in four comments now.

2. **LOW, `settle_until` was left out of the sweep** — **fixed.** Its `0..=TEST_TURNS` already read after the last turn, so no signal was lost, but it spent `TEST_TURNS + 1` turns while printing `TEST_TURNS`. It counts down like the rest now, so every bounded wait in the file agrees with the number it reports. Six loops touched in total: the four that lost a result, this one that miscounted, and `HOLD_YIELDS`'s stall wait that read one of its two marks.

3. **LOW, `conformance`'s module doc still calls this wait an untimed block** — **fixed.** It listed the handshake as the one exception to "every wait is a bounded number of executor turns and fails the test on its bound", which stopped being true when the wait became a counted `try_recv` loop and which this commit's own prose contradicts at length. It now says what the wait is — scheduler yields rather than executor turns, because the party it waits on is on another thread — and that it is bounded and fails on its bound like the rest.

4. **LOW, "deliberately contended outlier" overstates what 6664 is** — **fixed in three places.** That number came from *incidental* contention — a macOS run that happened to have a build beside it — while the deliberately oversubscribed sampling is the Linux column, whose maximum is 3511 and which the 39x is not taken over. My own paragraph got this right thirty lines later and wrong where the argument rests on it. Corrected in `HANDSHAKE_YIELDS`'s doc, in `HOLD_YIELDS`'s, and in #342, which had inherited the same wording.

Three fixed, one declined. The declined item is the value, recorded in the tree, the PR body, and #342.

akiomik:
Review round on `64b5fc0` returned 5 findings. Four fixed, one declined and standing — head is now `b1f90e9`.

1. **MEDIUM, the residual sub-quantum exposure** — **declined, unchanged.** Fifth round on it, and the review itself scopes it as "a flag on the residual exposure rather than a defect". Recorded in the tree, the PR body, and #342; flagged for the maintainer in five comments now.

2. **LOW, the 11x re-imports the comparison the doc had just disqualified** — **fixed, by drawing the distinction rather than dropping the number.** The catch is right that the two uses were run together. They are different: a *margin* is a property of the bound, and a stall bound's is more forgiving than a total's, which is why 39x and 320x are not the same kind of number; a *tail factor* is a property of the host — how much worse an unsampled run is than a sampled one under the same scheduler — which does carry across bound shapes. The doc now says that, and says why it still carries weakly: the factor comes from a macOS column while this bound's worst is a Windows observation. So 11x is labelled the conservative end of an unpinned factor, not a derived quantity.

3. **LOW, the coverage-leg sentence overstates a single observation** — **fixed.** It claimed instrumentation "leaves the one statistic this bound is sized against where it was", but that statistic is 155 on Windows, while the comparison is 20 against 20 on Linux from one run, on a runner class that never produced anything near it. It now says what the sample supports: instrumentation did not lift the Linux maximum.

4. **LOW, the rewritten `# Panics` names only one of two reachable panics** — **fixed.** `Disconnected` — the gated run torn down before reaching its send — is a distinct failure with its own message, and the section that was rewritten in this diff is where it belongs.

5. **NIT, a duplicated opening point and two unwrapped fragments** — **fixed.** The two opening paragraphs made the same argument; they are one now. Both fragments are reflowed.

Four fixed, one declined.

akiomik:
**The value is now `250_000`.** Head is `c8fc97c`. I had declined that across six rounds; a maintainer-solicited review made the case that broke the decline, and I have no rebuttal to it.

**What broke it.** My standing reason was that #305 settled the direction. It settles it on this premise:

> the passing path exits at the commit whatever the bound is, so **a smaller value costs a passing suite nothing**

That premise does not survive the measurement #305 itself asked for. The Windows mechanism these rounds established — `yield_now` is `SwitchToThread`, reaching only threads ready on the calling processor, and `windows-latest` is Server with a ~187 ms long fixed quantum — says a correct-but-late run spends the bound too. So there *is* a cost on the passing path, and "sized against the failing path" was resting on a premise this PR retired. Following #305 means following the measurement it asked for, not the provisional number it proposed before anyone had taken one.

**Why 250,000 specifically.** The quantum makes this a threshold rather than a slider:

| bound | idle rate | contended rate | vs ~187 ms quantum |
| --- | --- | --- | --- |
| 50,000 | 7.5 ms | 105 ms | under at **both** ends |
| 250,000 | 37.5 ms | 525 ms | over at one end |

`50_000` clears #305's stated target at 7.7 s but spends its whole budget inside one quantum at either rate. `250_000` reports in ~38 s and crosses at the contended end. The costs are asymmetric — a slow report is paid once by whoever broke the fixture and terminates either way, while a bound that expires on a correct tree is paid by every PR on a required check — and this repository's own stopping rule names "CI false positive" as a merge blocker, which is the category `50_000`'s risk sits in.

It also matters that most of #305's complaint was about *diagnosis* rather than latency ("reads as a hang rather than as the failure it is"), and the panic rewrite in this PR answers that half at any value by naming the bound and both causes.

**Stated honestly in the doc.** `250_000` is not supported by evidence `50_000` lacks — both are choices under the same unmeasured tail, and the contended rate is borrowed from another host and build profile. The doc says so, and frames the choice as a decision rule: where the uncertainty is unresolved, the cheaper thing to be wrong about is the slow report. #342 replaces the rule with a number.

**Not a return to where this branch already was.** It carried `250_000` for one round earlier, on two arguments that the following round refuted — a mechanism claim contradicted by this PR's own measurement, and a wall-clock "floor" derived from scheduling noise. Neither survives, and neither is the reason now. The reason now is the quantum threshold, which did not exist in the argument then.

Every figure downstream of the value is updated (38 s report, 37.5 ms / 525 ms tolerances, 1,600x raw and ~54x tail-adjusted headroom), and the panic was verified against the real failing path at the new bound.

akiomik:
Review round on `c8fc97c` returned 2 findings, both low. Both fixed — head is now `f8b3045`.

1. **The module-doc fix installed a new overclaim** — **fixed.** Correcting "one wait is the exception, an untimed block" by deleting the carve-out left the paragraph asserting that *every* wait is bounded and fails on its bound. `QuiescenceGate::wait` is a bare `Condvar` loop, and the file's own module doc has always described it as the second form: it terminates because the type driving its release owns it, not by a count. The paragraph now names both forms, so fixing one false universal did not create another. Exactly the failure mode this PR has been correcting all the way through, and I walked into it while correcting an instance of it.

2. **The PR body's evidence was gathered at a different constant than the one shipping** — **fixed.** The History line ended at `50,000` and the Verification section quoted the panic and the instrumentation reading at `50000`, so the body read as though `50_000` were final and as though the failing-path evidence covered the shipped value. The panic output is now the one re-run at `250_000`; the `yields=50000 final_poll_ran=1 bound=50000` reading is kept but labelled for what it is — the check that established the poll/yield accounting, which is a property of the loop shape and independent of the value. The History line carries the fifth move and says why it is not a return to where the branch had been. The "five instances" / "six loops" off-by-one is corrected too.

Both fixed, none declined. Nothing open on this PR.

akiomik:
Review round on `f8b3045` returned **no correctness defects in the diff**. One finding, outside it, plus one observation the doc already covers.

1. **`TestDriver::settle_kernel` has the same off-by-one** (`src/testing/driver.rs`) — **filed as #343 rather than fixed here.** Verified before filing: it polls before each turn and never after the last, so a terminated kernel draining on turn 1024 returns `None` and asserts `"the terminated kernel's join set did not drain"` — a correct drain reported as a hang, with a message naming a cause that did not happen.

   Not fixed in this PR because `testing` is `pub mod` — shipped test-support that dependants use — while this PR is scoped to `#[cfg(test)]` conformance fixtures and says so in its body. Extending it there would change what the PR is and would deserve reviewing as a public-surface edit rather than riding along on a fixture sweep. #343 carries the fix, the countdown form, and a note to check the rest of `src/testing/` for the same shape, which the conformance sweep did not cover.

2. **The quantum-threshold argument holds at the contended rate and not the idle one** — no action; the review filed it as an observation rather than a finding, and it is what the doc says about itself. `250_000` is "over at one end", the doc's own table shows both ends, and the paragraph that follows states the choice as a decision rule under an unmeasured tail rather than as a measurement. #342 is what replaces it with a number.

## Loop state

This round reviewed the current head and returned nothing that changes it, which is the fixed point I take as the stopping condition — not "no blockers", but "a complete round on this head asks for no edit to it".

Sixteen rounds of dispositions are recorded above. Every finding across them was fixed, explicitly declined with its reason in the tree, or filed: #342 for the contended measurement that would replace the value's decision rule, #343 for the shipped-code instance of the off-by-one.

PR #344, opened 2026-09-03
fix(testing): read the last turn in the kernel settle wait

### LINKED ISSUE #343: TestDriver::settle_kernel never polls after its last turn, so a drain on the final turn reports as exhausted
## The defect

`TestDriver::settle_kernel` (`src/testing/driver.rs`) polls the settle future *before* each turn and never after the last one:

```rust
for _ in 0..TURN_BUDGET {
    if let Poll::Ready(report) = futures::poll!(settle.as_mut()) {
        return Some(report);
    }
    executor_turn().await;
}
None
```

`TURN_BUDGET` turns are spent and `TURN_BUDGET` polls are made, so the final `executor_turn()` opens a window nothing reads. A terminated kernel whose join set drains *on* that turn returns `None`, and the driver asserts:

```
bounded settle exhausted after 1024 executor turns: the terminated kernel's join set did not drain
```

— a correct drain reported as a hang, with a message that states a cause that did not happen.

Low probability at `TURN_BUDGET = 1024`, and it has always been there.

## Why it is filed rather than fixed in place

#341 found this class in `src/kernel/conformance/support.rs` and closed all six instances there: `open_and_await_commit`, `HOLD_YIELDS`'s flat and stall waits, `step_when_ready`, `drive_to_ready`, and `settle_until`. The fix is the countdown form:

```rust
for remaining in (0..=TURN_BUDGET).rev() {
    if let Poll::Ready(report) = futures::poll!(settle.as_mut()) {
        return Some(report);
    }
    if remaining == 0 {
        break;
    }
    executor_turn().await;
}
```

which spends `TURN_BUDGET` turns across `TURN_BUDGET + 1` polls, so the count the panic prints stays true.

That PR is scoped to `#[cfg(test)]` conformance fixtures and says so. `testing` is `pub mod` — shipped test-support that dependants use — so extending it there would change what the PR is, and the change deserves reviewing as a public-surface edit rather than riding along.

## Worth checking at the same time

Whether any other bounded wait in `src/testing/` has the same shape. The conformance sweep covered its own file only.

akiomik:
Closes #343.

## The defect

`TestDriver::settle_kernel` (`src/testing/driver.rs`) polled the settle future *before* each turn and never after the last one:

```rust
for _ in 0..TURN_BUDGET {
    if let Poll::Ready(report) = futures::poll!(settle.as_mut()) {
        return Some(report);
    }
    executor_turn().await;
}
None
```

`TURN_BUDGET` turns were spent and `TURN_BUDGET` polls made, so the final `executor_turn()` opened a window nothing read. A terminated kernel whose join set drained *on* that turn returned `None`, and the driver asserted:

```
bounded settle exhausted after 1024 executor turns: the terminated kernel's join set did not drain
```

— a correct drain reported as a hang, with a message stating a cause that had not happened.

## The fix

The bound moves before the turn, which is the shape `confirm` (`:812`) and `settle` (`:922`) in this same file already use. `TURN_BUDGET` turns across `TURN_BUDGET + 1` polls, so the count the assertion prints stays true.

## Verification

Forcing the bound and counting what the loop actually spends:

```
PROBE polls=1025 turns=1024 budget=1024
```

Exactly `TURN_BUDGET` turns, one more poll than that. Before the change the same probe reports `polls=1024`, which is the missing read.

Also: `cargo test --lib` 589 passed, `just clippy` (`--all-targets`, `build_features`) and `just fmt-check` clean.

## The rest of `src/testing/`

The issue asked for a sweep. There is one other bounded loop in the module (`:1766` after this change), and it is **not** affected — it re-reads `wake_source_ready` in the assertion after the loop, so its last turn is observed. `settle` and `confirm` were already assert-before-turn. So this was the only instance.

## Why it was not in #341

That PR swept this defect class out of `kernel::conformance`'s fixtures, where it appeared six times, and is scoped to `#[cfg(test)]` code. `testing` is `pub mod` — shipped test-support that dependants use — so this belongs in a change that can be reviewed as a public-surface edit rather than riding along on a fixture sweep.

### DIFF
--- a/src/testing/driver.rs
+++ b/src/testing/driver.rs
@@ -1118,9 +1118,24 @@ impl<P: Program, B: Backend> TestDriver<P, B> {
         let settled = executor.block_on(async {
             let settle = kernel.settle();
             tokio::pin!(settle);
-            for _ in 0..TURN_BUDGET {
+            // One read more than turns. `0..TURN_BUDGET` read before each
+            // turn but never after the last, so a join set that drained *on*
+            // turn `TURN_BUDGET` was reported as one that never drained. The
+            // bound sits between the read and the turn, counting up like
+            // `settle` and `confirm` above; the turns still number exactly
+            // `TURN_BUDGET`, which is what the assertion prints.
+            //
+            // Those two end on the assertion itself, where this breaks and
+            // asserts outside the executor. Not because it must — a panic
+            // inside `block_on` propagates fine — but so the message is
+            // raised on the caller's frame rather than from inside a
+            // future being driven.
+            for turns in 0..=TURN_BUDGET {
                 if let Poll::Ready(report) = futures::poll!(settle.as_mut()) {
                     return Some(report);
+                }
+                if turns == TURN_BUDGET {
+                    break;
                 }
                 executor_turn().await;
             }


akiomik:
Review round on `3e17bdb`: **no blocking findings**, one optional note. Fixed — head is now `90a4bf2`.

1. **The comment's opening clause misdescribes what changed** — **fixed.** "Bounded before the turn rather than around it" is wrong about the prior shape: the old loop also read before its turn. What it lacked was the read *after* the last one. The comment now opens on that ("one read more than turns") and says what the old range did, so the first sentence is the accurate one rather than a redundant lead-in to it.

   The same note pointed out that "like `settle` and `confirm` above" holds for the bound's *placement* but not its *mechanism* — they end on an assertion where this breaks, because the loop has to hand an `Option` back out of `block_on`. The comment now says that too.

On `codecov/patch`: red, and the same content as #341's — the uncovered changed line is the `if turns == TURN_BUDGET { break; }` arm, which is the budget-exhaustion path of a bounded wait. Reaching it means the assertion below fires, so a passing suite cannot execute it. `codecov/project` passes and `codecov.yml` makes neither a gate.

Also noted from the review, for the record rather than as a discrepancy: `cargo test --lib` reports 544 rows on default features against the 589 this PR quotes with `build_features`. Same tree, different feature set.

akiomik:
Review round on `90a4bf2`: **no correctness issues**, two non-blocking notes. One fixed, one declined with the reason — head is now `a1bdda5`.

1. **The comment's reason for breaking rather than asserting does not hold** — **fixed.** I wrote that the loop "has to hand an `Option` back out of `block_on`", which is not why: a panic inside `block_on` propagates perfectly well, so an `assert!` in the async block would work. The actual reason is that the message should be raised on the caller's frame rather than from inside a future being driven, and the comment now says that instead.

2. **Three spellings of one shape now live in the tree** — count-up here, countdown in `settle_until`, assert-based in `settle`/`confirm` — **declined, with the reason in the comment.**

   Count-up is this file's existing idiom: `settle` and `confirm` sit two hundred lines above and are already `for turns in 0..=max_turns` with the bound checked before the turn. Matching them locally seemed worth more than matching a different file, and the comment now says "counting up like `settle` and `confirm` above" so the choice reads as deliberate.

   Not filed as an issue either, because unifying would mean editing the conformance loops that merged with #341 for no behavioural gain, and the shape's correctness is verified per site rather than by uniformity. If a single spelling is wanted across the tree, that is a call worth making once over all three rather than as a side effect of this fix — happy to file it if you would prefer that.

akiomik:
Review round on `a1bdda5`: **no findings**. One non-blocking note, fixed in the description only — the tree is unchanged, so the head is still `a1bdda5`.

- **The PR body had `settle` and `confirm`'s line references swapped.** Verified before fixing: `:812` is `confirm`'s loop and `:922` is `settle`'s, and I had written them the other way round. Corrected, along with the sweep section's line reference for the unaffected loop, which had shifted with this diff. The in-code comment names both without line numbers and was already right.

That is the fixed point: a complete round on this head asked for no edit to it. Across three rounds every finding was fixed except one, the three-spellings-of-one-shape observation, which is declined with its reason recorded in the comment above and in the code.

PR #345, opened 2026-09-03
test(observability): make the census's rise guards defend themselves

### LINKED ISSUE #339: Close producer_gauges_rose's two remaining vacuity paths: an empty expected_active, and the census guard's own missing row
## Background

#335 (PR #337 → #338) moved the INV-LC7 census's rise half into `tests/common/gauges.rs` as `producer_gauges_rose` and gave it the same census-membership guard `producer_gauges_are_zero` carries. Two vacuity paths in the new helper were left open deliberately, both latent today, both surfaced in that PR's review rounds.

## 1. An empty `expected_active` passes vacuously

`producer_gauges_rose` loops over `expected_active` and asserts nothing when it is empty.

That is *not* the same trade `producer_gauges_are_zero` makes. There, an empty slice is a documented reading ("a no-producers reading, not an INV-LC7 check") and the census's own unit row `an_active_gauge_with_no_events_is_not_settled` passes `&[]` on purpose to pin the tolerant branch. `producer_gauges_rose` has no such caller and no such meaning.

The risk is that both halves go tolerant at once. `tests/lifecycle.rs`'s `assert_two_stage_postconditions` passes the *same* slice to both, and its `expected_active` is built dynamically (`let mut expected_active = vec!["keyed_commands"];` plus conditional pushes). It cannot be empty today — it is seeded with one element — but a row that ever reached the helper with an empty slice would get a vacuous rise check *and* a settle loop that returns on iteration zero: exactly the silenced-instrument vacuity the census exists to prevent, now reachable through the helper written to prevent it.

An `assert!(!expected_active.is_empty(), ...)` in `producer_gauges_rose` closes it, and it does not touch the item #335 explicitly declined to make strict.

## 2. The membership guard has no row of its own

The guard is the only thing keeping `producer_gauges_rose` from a false pass on a non-census name: `TraceRecorder::u64_values` reads a field name off the whole `tears::runtime::load` target, and every gauge event also carries `seq` (bumped before capture, so `>= 1`) and `runtime_id` (allocated from 1, with 0 reserved). Without the guard, `producer_gauges_rose(&recorder, &["seq"])` passes by asserting that the ordering counter rose.

It was verified by a throwaway probe during review, so nothing in the tree fails if the guard is later removed — unlike its sibling's strict/tolerant split, which `an_active_gauge_with_no_events_is_not_settled` pins.

A row asserting the panic would make the guard self-defending. The cost to weigh: `tests/common/gauges.rs` is included by `#[path]` into every gauge-reading target, so one `#[should_panic]` row here becomes one per including binary — against the module's stated policy that an item only one target needs stays with its caller, and against #306's open question about `#[should_panic]` rows and panic-hook serialization.

## Why deferred

#338's review closed with no correctness findings; both of these are latent-only and neither is reachable from a call site that exists. They are recorded here so the choice is made deliberately rather than by omission.

akiomik:
Closes #339.

## Two vacuity paths and one undefended guard

**An empty `expected_active` asserted nothing.** That is not the trade its sibling makes: `producer_gauges_are_zero` documents empty as a tolerant no-producers reading and has a row that passes `&[]` on purpose. This half has no such meaning and no such caller.

The hazard is that both halves go vacuous together. `assert_two_stage_postconditions` hands the *same* slice to each, so an empty one gives a rise check that asserts nothing **and** a settle loop that returns on iteration zero — the silenced-instrument vacuity the census exists to prevent, reached through the helper written to prevent it. It cannot be empty today (its `expected_active` is seeded with `"keyed_commands"`), which is why #339 filed it rather than #338 fixing it.

**The membership guard had no row.** It is the only thing between this helper and a false pass on a non-census name: `u64_values` reads a name off the whole `tears::runtime::load` target, and every gauge event carries `seq` and `runtime_id`, both above zero by construction — so `producer_gauges_rose(&recorder, &["seq"])` would pass by asserting that the ordering counter rose. That was verified by a throwaway probe during #338's review, which left nothing in the tree to fail if the guard were later removed.

## Verification

Both rows fail when their guard is disabled:

```
test gauges::tests::an_empty_rise_check_is_refused - should panic ... FAILED
test gauges::tests::a_non_census_name_cannot_satisfy_the_rise_half - should panic ... FAILED
```

Unmutated: `cargo test --test lifecycle --test observability` — 17 + 5 passed. `just clippy` (`--all-targets`, `build_features`) and `just fmt-check` clean.

## On `#[should_panic]`, which #339 flagged as the cost to weigh

The guards *are* asserts, so pinning them means catching their panic; a check on an extracted predicate would pin the predicate and not its use.

#339 raised #306's hazard against this. It does not transfer, and I checked rather than assumed: #306's 31 rows share the **lib** binary with `panic`'s recording-hook row, which installs a counting hook and asserts it ran exactly once, so a foreign panic inside that window corrupts the count. This module compiles into `tests/lifecycle.rs` and `tests/observability.rs` — separate binaries — and neither has a hook-counting row. `tests/lifecycle.rs` does include `common/panic_hook.rs`, but that helper only silences reporting and asserts nothing about invocations (`grep` over `tests/` finds no hook-count assertion at all).

The per-binary duplication is the module's stated policy rather than an exception to it: its doc already says each row is "compiled per including target, so this row runs once per binary — the deliberate cost of each target pinning the census it links against."

### DIFF
--- a/tests/common/gauges.rs
+++ b/tests/common/gauges.rs
@@ -74,15 +74,29 @@ pub const SETTLE_STEPS: usize = 1_000;
 /// those and *pass*, asserting that the ordering counter rose. The guard is
 /// what keeps this half unable to satisfy itself on a non-gauge.
 ///
+/// An empty `expected_active` is refused, which is the opposite of what
+/// [`producer_gauges_are_zero`] does with one — read the two together, since
+/// a caller usually hands the same slice to both. Empty is a meaningful
+/// reading there (no producers to settle) and no reading at all here, and
+/// letting it through would take both halves of the census vacuous at once:
+/// a rise check that asserts nothing beside a settle loop that returns on
+/// its first iteration.
+///
 /// Written as a plain loop rather than an iterator chain for the same
 /// `#[track_caller]` reason as [`producer_gauges_are_zero`].
 ///
 /// # Panics
 ///
-/// If any name is outside [`PRODUCER_GAUGES`], or any named gauge has no
-/// reading above zero.
+/// If `expected_active` is empty, if any name in it is outside
+/// [`PRODUCER_GAUGES`], or if any named gauge has no reading above zero.
 #[track_caller]
 pub fn producer_gauges_rose(recorder: &TraceRecorder, expected_active: &[&str]) {
+    assert!(
+        !expected_active.is_empty(),
+        "an empty rise check asserts nothing, and its caller hands the same slice to \
+         `producer_gauges_are_zero`, where empty is the tolerant reading — so both halves \
+         of the census would go vacuous at once"
+    );
     assert!(
         expected_active
             .iter()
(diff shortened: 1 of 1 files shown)


akiomik:
Review round on `5ebb494` returned 3 findings. All three fixed — head is now `1f7eb56`.

1. **MEDIUM, the non-census row pinned the guard's message rather than the hazard** — **fixed, and this was the finding worth the round.** The fixture was a bare `TraceRecorder` recording nothing, so removing the guard made the row fail on the *rise* assert finding no readings — never reaching the state the comment described. The row asserted that the guard's string exists, not that the hazard is real.

   It now installs the recorder and emits a gauge-shaped event on `tears::runtime::load`, with a premise assertion that `u64_values("seq")` reads `[1]`. Verified in both directions: unmutated the row passes; with the guard removed it reports **"test did not panic as expected"** — the false pass itself, which is what the row exists to witness.

2. **LOW, the contract did not mention the new refusal** — **fixed.** `# Panics` now names the empty case, and the doc body says the refusal is the *opposite* of what `producer_gauges_are_zero` does with an empty slice, and why: a caller usually hands the same slice to both, empty is a meaningful reading there and no reading at all here, and letting it through takes both halves vacuous at once. That contrast is the part a reader of the two side by side needs.

3. **LOW, the new comment block displaced an existing one** — **fixed.** The strict/tolerant paragraph is back on `an_active_gauge_with_no_events_is_not_settled`, the guard rows have their own heading which names that row rather than saying "the row below", and the module comment says "these rows" now that there are three.

Three fixed, none declined.

akiomik:
Review round on `1f7eb56`: **no findings**, and the round independently re-ran the mutation check rather than taking the PR body's word for it — deleting the membership assert flips `a_non_census_name_cannot_satisfy_the_rise_half` to "did not panic as expected", which confirms both that the row defends the guard and that its premise holds.

One non-blocking observation, recorded rather than acted on: in the `lifecycle` binary these panicking rows can overlap a `with_silent_panic_hook` window and lose their stderr location line. No action, because libtest matches `should_panic` on the unwind payload rather than on stderr, and prints the payload itself on a mismatch — so nothing about the rows' verdict or their diagnostic depends on the hook that happens to be installed. `docs/testing.md` already rules that ordinary panicking rows need not take `PANIC_HOOK_GUARD`.

That is the fixed point: a complete round on this head asked for no edit to it. Across two rounds every finding was fixed, none declined.

PR #347, opened 2026-09-03
test(observability): read gauge events through one per-event accessor

### LINKED ISSUE #332: TraceRecorder: expose per-event gauge access so multi-runtime reads stop hand-rolling positional zips
## Background

#318 gave the test `TraceRecorder` a per-event `u64_events` store (each event's unsigned-integer fields kept together) and a `current_u64(field)` accessor that reads the greatest-`seq` snapshot within a single runtime instance. `current_u64` deliberately **panics** when the matched events span more than one `runtime_id`, because `seq` is comparable only within one instance.

That leaves two consumers that genuinely need per-event or per-runtime data still hand-rolling it, because no accessor exposes the grouping the store now holds:

- **`gauge_trace` (`src/kernel/producer.rs:361`)** rebuilds per-event `(subscriptions, unkeyed, keyed)` triples by positionally `zip`ping three separately-flattened `u64_values` vectors (`:368-370`), then asserts exact literals like `vec![(0, 1, 0), (0, 0, 0)]`. `zip` truncates to the shortest input and never reports a length mismatch, so a schema change that stops a count riding every snapshot silently shifts the tuples or shortens the vector — the `is_empty()` assertions pass vacuously and the exact-sequence ones compare misaligned data.
- **The gauge-ordering contract rows in `src/runtime/load.rs`** (e.g. `gauge_seq_strictly_increases_within_each_runtime_id`, `each_observer_gets_its_own_runtime_id`) deliberately observe **two** `LoadObserver`s under one recorder and reconstruct each instance's current value with a hand-rolled `BTreeMap<runtime_id, …>` partition-then-max. They cannot call `current_u64` (it would panic on the second `runtime_id`), so they re-implement the contract's read rule independently — the reader and the contract's own enforcement rows cannot witness each other's bugs.

## Proposal

Expose the per-event grouping the recorder already stores, so both consumers read it instead of re-deriving it:

- A raw per-event accessor (`u64_events() -> &[Vec<(String, u64)>]` or a `snapshots()` view), and/or
- A `current_u64_by_runtime(field) -> BTreeMap<u64, u64>` that returns the greatest-`seq` value per `runtime_id`, with `current_u64` as the `len() <= 1` convenience over it.

`gauge_trace`'s hand-rolled zip then reads real per-event tuples, and `load.rs`'s contract rows consume the same accessor they exist to enforce.

## Why deferred from #318

#318's scope was the `.last()` → `current_u64` migration; adding a partition API is a separate change, and the two-runtime enforcement rows must keep hand-rolling the partition *if* the concern is that they should not depend on the convenience reader they justify — this issue is to decide that trade deliberately rather than by omission. Surfaced repeatedly across #318's review rounds.

akiomik:
Closes #332.

`TraceRecorder` has kept each event's unsigned-integer fields grouped since the `current_u64` migration, but nothing exposed that grouping — so the two consumers that need it re-derived it from the flattened per-field views.

## What was wrong

- **`gauge_trace` (`src/kernel/producer.rs`)** rebuilt its per-event `(subscriptions, unkeyed, keyed)` triples by positionally zipping three separately flattened `u64_values` vectors. `zip` pairs by index, truncates to its shortest input and reports no mismatch, so a count that stopped riding every snapshot would shift the triples or shorten the vector — and the two `is_empty()` rows beside it would then pass on an emission outage rather than on the absence they assert.
- **`gauge_seq_strictly_increases_within_each_runtime_id` (`src/runtime/load.rs`)** observes two `LoadObserver`s under one recorder on purpose, so it cannot call `current_u64` (the cross-instance guard panics). It re-implemented the contract's own read rule — partition by `runtime_id`, then take the greatest `seq` — beside the emitter it exists to hold to that rule.

## What this adds

- **`u64_event_values([...]) -> Vec<[u64; N]>`** — the named fields grouped per event, ordered as named. An event recording none of the names is skipped; one recording *some* of them panics, naming the missing field and the whole event. That is the alignment a zip could only assume.
- **`current_u64_by_runtime(field) -> BTreeMap<u64, u64>`** — the contract's read rule itself: one current value per `runtime_id`, each the greatest-`seq` reading among that instance's own events. `current_u64` is now the one-instance convenience over it, and its cross-instance panic names it as the remedy instead of telling the caller to partition by hand.

Both recorder copies get both accessors, per the duplication policy; `docs/testing.md` gains both rules.

## The trade #332 asks to decide

The contract row's recovery half now consumes the reader it justifies. What that gives up is recovered on both sides:

- the strict-increase half above it still builds its own partition from the raw per-event log, so the reader cannot cover for an emitter that stopped ordering; and
- the reader's own rule is pinned on scripted events in the recorder's unit rows, where no emitter is involved at all.

The per-event read there also settles something the old form could only assume: `u64_event_values(["runtime_id", "seq", "subscriptions", "keyed_commands"])` refuses a gauge event carrying only some of the four, which is the premise the recovery half rests on — that every gauge event carries every count. The old form checked it with a length assert per field, which narrows that assumption rather than removing it.

## Dispositions

- `each_observer_gets_its_own_runtime_id`, also named in the issue, is unchanged: it asserts arrival-order attribution over a single field. It hand-rolls nothing the new accessors provide, and a by-runtime map cannot express that claim.
- The issue's alternative shape — a raw `u64_events()` view — is not added. It would hand every caller the field matching back, which is the part being centralised; a caller that needs the whole event still has `field_name_sets`.

## Verification

`just pre-commit` green (fmt, clippy, the full test matrix, the loom mirrors, doctests), plus `RUSTDOCFLAGS=-D warnings cargo doc`.

Two mutations, both run:

1. Making the by-runtime read take the latest arrival instead of the greatest `seq` fails both recorder rows (`current_u64_reads_by_seq_not_by_arrival`, `current_u64_by_runtime_reads_each_instance_by_its_own_seq`). The load contract row keeps passing under it — its scripted emissions happen to arrive in `seq` order per instance — which is exactly why the reader's rule is pinned on its own scripted events and not on that row.
2. Dropping `keyed_commands` from `GaugeSnapshot::dispatch` now fails at the accessor, naming the missing field and the whole event, in the rows that would previously have compared a silently shortened trace.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -146,13 +146,25 @@ tracing output.
 Read a gauge's *current value* through `current_u64`, never through
 `u64_values(...).last()`: the gauge contract orders events by `seq`, not by
 arrival. The accessor assumes one runtime instance per recorder and
-panics otherwise; a deliberately multi-instance test partitions by
-`runtime_id` itself, the way `src/runtime/load.rs`'s contract rows do.
+panics otherwise; a deliberately multi-instance test reads
+`current_u64_by_runtime`, which applies the same greatest-`seq` rule inside
+each `runtime_id`'s own events and which `current_u64` is the one-instance
+convenience over. A row asserting something else about those events — their
+arrival order, or that `seq` strictly increases inside one partition — reads
+the log itself instead; `src/runtime/load.rs`'s two-instance rows do both.
 The rule covers the gauge snapshot's fields — the ones whose events carry
 a `seq`. Other unsigned-integer fields on the same target (`pulled`,
 `updated`, `shared_pending`, `wait_us`) are per-event readings with no
 current value to take; `current_u64` returns `None` for them, so keep those
 on `u64_values`.
+
+Read several fields of the *same* event through `u64_event_values([...])`
+rather than by zipping as many `u64_values` logs: the per-field logs are
+flattened across events, so zipping them pairs values by position and
+truncates to the shortest input without reporting the mismatch — a field
+that stops riding every event then shifts the tuples or shortens the vector
+instead of failing. `u64_event_values` keeps the per-event grouping and
+refuses an event that records only some of the names.
 
 The INV-LC7 producer-gauge census in `tests/common/gauges.rs` is included
 with the same `#[path]` mechanism as the recorder above; see its module
--- a/src/kernel/producer.rs
+++ b/src/kernel/producer.rs
@@ -343,8 +343,22 @@ mod tests {
         RunKind::Sub(subscription.id().clone())
     }
 
-    /// The four producer gauges as `tears::runtime::load` reports them over
-    /// one run's guard: the values while it is held, then after it drops.
+    /// The three run-scoped producer gauges as `tears::runtime::load` reports
+    /// them over one run's guard: the values while it is held, then after it
+    /// drops. (`blocked`, the fourth field on the same snapshot, belongs to
+    /// the bounded-send layer and no run guard moves it.)
+    ///
+    /// Read per event, not by zipping three flattened per-field logs: `zip`
+    /// pairs by position and truncates to its shortest input without
+    /// reporting the mismatch, so a count that stopped riding every snapshot
+    /// would shift these triples or shorten the vector — and a shortened one
+    /// takes the `is_empty()` rows below with it, passing them on an emission
+    /// gap rather than on the absence they asserted. `u64_event_values`
+    /// refuses a snapshot carrying only some of the three instead.
+    ///
+    /// A snapshot carrying *none* of them is still skipped, so this trace
+    /// still empties out under a total gauge outage — what refuses that is
+    /// the two rows below that expect values, not this shape.
     fn gauge_trace(kind: &RunKind) -> Vec<(u64, u64, u64)> {
         let recorder = TraceRecorder::new().with_target("tears::runtime::load");
         let _guard = recorder.set_default();
@@ -352,15 +366,10 @@ mod tests {
 
         drop(gauge_guard(&observer, kind));
 
-        let subscriptions = recorder.u64_values("subscriptions");
-        let unkeyed = recorder.u64_values("unkeyed_commands");
-        let keyed = recorder.u64_values("keyed_commands");
-
-        subscriptions
+        recorder
+            .u64_event_values(["subscriptions", "unkeyed_commands", "keyed_commands"])
             .into_iter()
-            .zip(unkeyed)
-            .zip(keyed)
-            .map(|((subscriptions, unkeyed), keyed)| (subscriptions, unkeyed, keyed))
+            .map(<(u64, u64, u64)>::from)
             .collect()
     }
 
--- a/src/runtime/load.rs
+++ b/src/runtime/load.rs
@@ -232,7 +232,8 @@ impl Gauges {
 /// followed by RFC 0006 §4.4's four counts.
 ///
 /// The names are the contract, and they are matched *by name* by consumers
-/// that cannot see each other — the test recorders' `current_u64` markers,
+/// that cannot see each other — the test recorders'
+/// `current_u64_by_runtime` markers (which `current_u64` reads through),
 /// the integration census's `PRODUCER_GAUGES`, and the bench subscriber's
 /// `LoadVisitor`. `tracing`'s field names are macro syntax rather than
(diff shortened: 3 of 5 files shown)


akiomik:
## Review rounds (3 at `high`), with a disposition per finding

No round found a correctness defect. Every note below has a disposition.

**Round 1 — 4 low notes**

1. *Self-referential coverage*: the recovery half of `gauge_seq_strictly_increases_within_each_runtime_id` now asserts the read rule through the reader it justifies. **Accepted deliberately** — this is the trade #332 asks to decide, and the rationale sits in the code beside it. Round 3's mutation check sharpens the record: a reader patched to compare `seq` *across* instances (dropping the partition) still fails that row, so the row is not self-satisfying; what it stops witnessing is only the within-partition max-`seq` choice, which the recorder's own scripted rows pin.
2. `recover`'s message said "max-seq **event**" (singular) while the computation is one scan per gauge. **Fixed** — the contract's rule is per gauge, so the messages now claim a recovered partition, with the reason in the comment above them.
3. `gauge_trace`'s doc overclaimed what the accessor closes. **Fixed** — it now states that a *total* gauge outage still empties the trace and names the rows that refuse that instead.
4. `docs/testing.md` sent multi-instance readers to "load.rs's contract rows" when only one half of one row reads that way. **Fixed**.

**Round 2 — 2 low notes**

1. `GAUGE_EVENT_FIELDS`'s doc still named `current_u64` as the by-name consumer after this PR moved the marker literals into `current_u64_by_runtime`. **Fixed** — that array's doc is the map a renamer navigates by.
2. The partial-event panic asserted a categorical cause the recorder itself can falsify: `record_i64` keeps a negative field's *name* and drops its value, so such a field would read as "missing". **Fixed** by naming the condition the log actually holds rather than by special-casing the case. Latent otherwise — every numeric field on these targets is `usize`/`u64`.

**Round 3 — no findings.** Two cosmetic notes left as they are: one extra `BTreeMap` allocation per `current_u64_by_runtime` call (same asymptotics, in a test helper), and one loosely-reading phrase in `gauge_trace`'s doc.

**Stopping condition met**: a round producing no finding at the merge-blocking bar — shipping impact, a CI false positive or hang, or an invariant left unverified while passing — and no inaccuracy in text this PR authored.

Independent mutation checks run during review, beyond the two in the description: patching the reader to compare `seq` across instances fails the load contract row, and dropping `keyed_commands` from the emitter fails all three `u64_event_values` call sites at the accessor with the field named.

CI: 22/22 green on `59eee8567a0d4ffadffccc795a9e67c932b20659`.

akiomik:
## Round 4 — findings: none

Reviewed at `89014902047c69df86df23f2236fc1314467d62d`, which is the current head; nothing has changed since. CI 22/22 green on the same SHA.

The maintainer's four notes on round 3's head are dispositioned as follows.

1. **`N == 0` degraded the guarantee to silence** — *fixed*. `u64_event_values([])` would have answered `vec![]` whatever the log held, the one shape where "refuses a partial event" says nothing. It is now `const { assert!(N > 0, …) }`, i.e. a compile error at the call site (confirmed: `error[E0080]: evaluation panicked: u64_event_values needs at least one field name`). No row can pin it, because the failure mode was a *value* rather than a panic — the comment beside it records that as the reason it is refused where the caller is written.
2. **Duplicate names in `fields`** — *behaviour kept, doc corrected*. `["left", "left"]` repeats the column, which is harmless; the doc now says so, so it no longer reads as a stronger promise than the code makes.
3. **The refusal is vacuous at `N == 1`** — *doc fixed*. The doc now states what two or more names buy, that one name is `u64_values` narrowed to each event's first occurrence, and that none is a compile error.
4. **`current_u64` no longer breaks early** — *no change*. Recorded as an accepted consequence: the full scan is what lets the panic report every observed id, and the cost is a test helper over a handful of events. `load.rs`'s two `current_u64_by_runtime` calls are two scans of the same log for the same reason.

Round 4's reviewer also confirmed, by its own mutation, the trade this PR asks to accept consciously: patching the reader to take the latest arrival instead of the greatest `seq` fails only the recorder's scripted rows, not the load contract row. That is the deliberate half of #332 — the greatest-`seq` choice within a partition is pinned on scripted events; the partition itself, and the emitter's strict increase, are still checked against the live emitter (round 3's mutation, dropping the partition, does fail the contract row).

One non-code note from the round: the working tree is on this PR's branch rather than `main`. That is expected — it is where the branch was written — and the tree is clean.

Issue #322, opened 2026-08-31
Add a runnable reducer composition example and performance guide


akiomik:
## Summary

Add a runnable example and user guide for the reducer composition API introduced in 0.11.0.

The runtime and composition contract already exist: `Reducer`, `Program`, `scope`, `for_each`, `presented`, `into_program`, `Keyed`, `Slot`, `ProgramRuntime`, automatic scope application, and removal teardown are implemented by RFC 0014. What is missing is a production-facing path that shows how those pieces fit together in an application.

## Scope

- Build one runnable application with sibling features, a keyed collection, and optional presentation state.
- Close the reducer stack with `into_program` and run it with `ProgramRuntime`.
- Show child message routing and aggregation of child commands and subscriptions.
- Demonstrate that users do not manually apply scope metadata or teardown removed children.
- Add a test that drives the same composition through `TestDriver`.
- Document the cost of routing high-frequency messages through deep reducer stacks, including `Command::map` boxing, source-side coalescing, and `without_redraw()`.

This is documentation and example work. It should not redesign the runtime or reopen RFC 0014.

## Acceptance criteria

- [x] A runnable example covers `scope`, `for_each`, `presented`, and `into_program` in one coherent application.
- [x] The example includes child command and subscription output.
- [x] Removal of a `Keyed` row and dismissal/replacement of a `Slot` demonstrate automatic teardown.
- [x] A deterministic test uses `TestDriver` against the same composition.
- [x] User documentation explains when to keep a simple `Application` and when to move to composed reducers.
- [ ] Performance guidance covers high-frequency input and deep mapping chains.

akiomik:
#360 covers the example, the test and the user documentation. Staying open for
the performance guidance, which is now tracked as sub-issue #361.

What landed:

- [x] A runnable example covers `scope`, `for_each`, `presented`, and
      `into_program` in one coherent application — `examples/dashboard_composed.rs`,
      which is `examples/dashboard.rs` written a second time so the two differ in
      structure and in nothing else.
- [x] The example includes child command and subscription output.
- [x] Removal of a `Keyed` row and dismissal/replacement of a `Slot` demonstrate
      automatic teardown — all four removal shapes, asserted through the
      `on_teardown` hooks each child registers.
- [x] A deterministic test uses `TestDriver` against the same composition — the
      same reducer value the binary runs, closed with the same `init` and `view`.
- [x] User documentation explains when to keep a simple `Application` and when to
[...]

PR #360, opened 2026-09-04
docs: write dashboard.rs a second time, with composed reducers

akiomik:
Closes part of #322 — the runnable example, the `TestDriver` test, and the user
guide. The performance guidance stays open there.

## What this is

The composition API has been runnable since 0.11.0 and had no worked
application to read. A new example on a new domain would have shown the API
without answering the question a reader actually has, which is whether their
application wants it — so this is `dashboard.rs` written a second time with the
wiring moved off the root.

The two files run the same application: same panes, same focus cycle, same task
operations. The structure is what the diff is for. What else differs — the
children's runs, three keys, the lines the activity log gains, the status line
naming a request instead of an outcome, and the tasks arriving as messages — is
listed in the example's own module doc rather than left for a diff to turn up.

| `dashboard.rs` | `dashboard_composed.rs` |
| --- | --- |
| root updates `NavigationState` | `scope(Navigation, Segment::Navigation)` |
| root updates `ActivityLogState` | `scope(Activity, Segment::Activity)` |
| `TaskListState { tasks: Vec<Task> }` | `for_each` over `Keyed<TaskId, TaskState>`, one child reducer per row |
| `DetailState`, always present | `presented` over `Slot<DetailsState>` |
| `StatusState`, updated by the root | unchanged — not everything wants a boundary |

Every timer is declared on one interval and every request keyed under one
`CommandId`, deliberately. Written by hand those are collisions: one timer
shared by every row, one sync slot for the whole application. Composed they are
not, and no `.scoped(...)` call appears anywhere in the file.

## Two rules the rewrite makes explicit

- **A child's own work starts as a message routed through the boundary.**
  `init`'s command is the root's and crosses none, so a hook registered there
  would anchor where no teardown reaches it (`TaskMessage::Watch`).
- **Work spanning two children stays the root's.** A child is handed its own
  projected state and can reach neither the collection beside it nor the slot
  around it (`Message::SaveNotes`).

## Tests

`TestDriver` tests drive the same stack the binary runs — the same reducer
value, closed with the same `init` and `view`, started from a `Setup` carrying a
scripted input.

- `identical_child_identities_stay_apart_under_their_boundaries` reads the run
  identities each step started: three timers declared on one interval are three
  subscriptions, three requests keyed on one id are three slots, and none of
  them is the unqualified id the reducers wrote.
- `every_removal_tears_its_instance_down` walks all four removal shapes —
  `Slot::present` over an occupied slot, `Keyed::insert` over an occupied key,
  `Keyed::remove`, `Slot::dismiss` — and asserts one teardown per removal and
  none for the successor row still present. It is observed through the
  `on_teardown` hooks each child registers, since a finalizer produces no
  message by design.

Both were mutation-checked: moving a row's hook registration from the child to
the root makes the second test fail on its `settle` budget, so it is sensitive
to the scope the hook anchors at rather than to the hook existing.

## Example naming

The list in the README gains the two headings the files were already sorted by.
Structure examples are named after the application, because a reader choosing a
structure knows the shape of their problem and not the name of the API; feature
examples are named after the API item, because that is what they arrive looking
for; `http_todo.rs` carries both, as a feature that needs a real application to
be worth showing. No existing file is renamed.

`dashboard.rs` is not renamed or removed either. `Application` is run through an
adapter over the same kernel rather than a legacy path, and the guide needs it
as the shape to keep while a child is a single instance.

## Verification

`cargo test --lib --tests --examples`, `just test-doc`, `just clippy-loom`,
`just test-mirrors`, `cargo clippy --all-targets`, `cargo fmt --check`, `typos`,
and `RUSTDOCFLAGS=-D warnings cargo doc` all pass. The bench targets are not
run here — this touches examples and docs only, and `clippy --all-targets`
covers their compilation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,12 +227,27 @@ Create custom subscriptions by implementing the `SubscriptionSource` trait.
 
 ## Examples
 
-Check out the [`examples/`](examples/) directory for more examples:
+Check out the [`examples/`](examples/) directory. They fall into two groups,
+and each group is named after what a reader arrives knowing.
+
+### Application structure
+
+Named after the application, because someone choosing a structure knows the
+shape of their problem and not yet the name of the API. Read them in order:
+each one is the previous one at the next scale.
 
 - [`counter.rs`](examples/counter.rs) - A simple counter with timer and keyboard input
-- [`panic_hook.rs`](examples/panic_hook.rs) - Restoring the terminal on panic with `install_panic_hook`
 - [`views.rs`](examples/views.rs) - Multiple view states with navigation and conditional subscriptions
-- [`dashboard.rs`](examples/dashboard.rs) - Structured state management with nested state and child messages
+- [`dashboard.rs`](examples/dashboard.rs) - Structured state management, with the root owning the wiring
+- [`dashboard_composed.rs`](examples/dashboard_composed.rs) - The same application with composed reducers: a keyed collection of child reducers, an optionally-present child, automatic scope application and teardown of removed children
+
+### Framework features
+
+Named after the API item they demonstrate, because that is what a reader
+arrives looking for. `http_todo.rs` carries both, as a feature that needs a
+real application to be worth showing.
+
+- [`panic_hook.rs`](examples/panic_hook.rs) - Restoring the terminal on panic with `install_panic_hook`
 - [`signals.rs`](examples/signals.rs) - OS signal handling with graceful shutdown (SIGINT, SIGTERM, etc.)
 - [`command_timeout_retry.rs`](examples/command_timeout_retry.rs) - Enforcing a `Command` deadline with `timeout` and recovering from failures with `retry`
 - [`command_cancellation.rs`](examples/command_cancellation.rs) - Cancelling superseded in-flight commands with `cancellable`/`cancellable_with` and `CancelPolicy`
@@ -246,9 +261,10 @@ Run an example:
 
 ```bash
 cargo run --example counter
-cargo run --example panic_hook
 cargo run --example views
 cargo run --example dashboard
+cargo run --example dashboard_composed
+cargo run --example panic_hook
 cargo run --example signals
 cargo run --example command_timeout_retry
 cargo run --example command_cancellation
@@ -287,6 +303,14 @@ runnable tests ship with these examples:
 cargo test --example command_timeout_retry
 cargo test --example command_cancellation
 cargo test --example dashboard
+```
+
+A composed `Program` is driven with `tears::testing::TestDriver` rather than
+`TestStore`, which takes an `Application`; `examples/dashboard_composed.rs`
+carries two worked `TestDriver` tests:
+
+```bash
+cargo test --example dashboard_composed
 ```
 
 Repository-wide test conventions live in [docs/testing.md](docs/testing.md).
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ cargo test --example dashboard
 
 A composed `Program` is driven with `tears::testing::TestDriver` rather than
 `TestStore`, which takes an `Application`; `examples/dashboard_composed.rs`
-carries two worked `TestDriver` tests:
+carries worked `TestDriver` tests:
 
 ```bash
 cargo test --example dashboard_composed
--- a/README.md
+++ b/README.md
@@ -233,10 +233,10 @@ feature, one child per row of a `Keyed` collection, or an optionally-present
 child in a `Slot` — and `into_program` closes the stack into a runnable
 `Program` for `ProgramRuntime`. Every boundary qualifies the identities its
 child produces with its own segment, so rows declaring the same subscription or
-keying the same command do not collide. The two that hold a collection —
-`for_each` and `presented` — additionally tear a removed child's runs down, so
-no removal leaks a run; `scope` composes one child in place, where there is no
-collection for an instance to leave.
+keying the same command do not collide. The two whose child can leave —
+`for_each` over a `Keyed` collection and `presented` over a `Slot` —
+additionally tear a departed child's runs down, so no removal leaks a run;
+`scope` composes one child in place, and that child is always there.
 
 An `Application` is run through an adapter over the same kernel, so this is a
 way to write a program rather than a second runtime. See
--- a/README.md
+++ b/README.md
@@ -325,9 +325,9 @@ cargo test --example command_cancellation
 cargo test --example dashboard
 ```
 
-A composed `Program` is driven with `tears::testing::TestDriver` rather than
-`TestStore`, which takes an `Application`; `examples/dashboard_composed.rs`
-carries worked `TestDriver` tests:
+`TestStore` takes an `Application`, so a composed `Program` is driven with
+`tears::testing::TestDriver` instead; `examples/dashboard_composed.rs` carries
+worked `TestDriver` tests:
 
 ```bash
 cargo test --example dashboard_composed
--- a/README.md
+++ b/README.md
@@ -231,10 +231,12 @@ A `Reducer` owns one state transition and the subscriptions that state declares.
 `scope`, `for_each` and `presented` compose one under a parent — a sibling
 feature, one child per row of a `Keyed` collection, or an optionally-present
 child in a `Slot` — and `into_program` closes the stack into a runnable
-`Program` for `ProgramRuntime`. Each boundary qualifies the identities its child
-produces with its own segment and tears a removed child's runs down, so rows
-declaring the same subscription or keying the same command do not collide and no
-removal leaks a run.
+`Program` for `ProgramRuntime`. Every boundary qualifies the identities its
+child produces with its own segment, so rows declaring the same subscription or
+keying the same command do not collide. The two that hold a collection —
+`for_each` and `presented` — additionally tear a removed child's runs down, so
+no removal leaks a run; `scope` composes one child in place, where there is no
+collection for an instance to leave.
 
 An `Application` is run through an adapter over the same kernel, so this is a
 way to write a program rather than a second runtime. See
@@ -250,8 +252,9 @@ and each group is named after what a reader arrives knowing.
 ### Application structure
 
 Named after the application, because someone choosing a structure knows the
-shape of their problem and not yet the name of the API. Read them in order:
-each one is the previous one at the next scale.
+shape of their problem and not yet the name of the API. The first three are a
+scale, each one the previous grown; the fourth is the third one rewritten, so
+read it beside `dashboard.rs` rather than after it.
 
 - [`counter.rs`](examples/counter.rs) - A simple counter with timer and keyboard input
 - [`views.rs`](examples/views.rs) - Multiple view states with navigation and conditional subscriptions
--- a/README.md
+++ b/README.md
@@ -225,6 +225,23 @@ Tears follows **The Elm Architecture (TEA)** pattern:
 
 Create custom subscriptions by implementing the `SubscriptionSource` trait.
 
+### Composing Reducers
+
+A `Reducer` owns one state transition and the subscriptions that state declares.
+`scope`, `for_each` and `presented` compose one under a parent — a sibling
+feature, one child per row of a `Keyed` collection, or an optionally-present
+child in a `Slot` — and `into_program` closes the stack into a runnable
+`Program` for `ProgramRuntime`. Each boundary qualifies the identities its child
+produces with its own segment and tears a removed child's runs down, so rows
+declaring the same subscription or keying the same command do not collide and no
+removal leaks a run.
+
+An `Application` is run through an adapter over the same kernel, so this is a
+way to write a program rather than a second runtime. See
+[docs/composition.md](docs/composition.md) for when the rewrite is worth it, and
+[`examples/dashboard_composed.rs`](examples/dashboard_composed.rs) for
+`dashboard.rs` written the other way.
+
 ## Examples
 
 Check out the [`examples/`](examples/) directory. They fall into two groups,
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -86,6 +86,15 @@ the root reducer, so the root keeps exactly the messages that are its own.
   instance's scope. The removed instance's subscriptions stop, its in-flight
   commands are cancelled, and the cleanup hooks it registered run.
 
+- **It discards what it cannot route.** A message addressed to a key the
+  collection no longer holds, or to a slot with no occupant, reaches no reducer
+  and is dropped — with no diagnostic, and with no way for the sender to learn
+  it went nowhere. That is reachable from ordinary code: a root that returns
+  `Command::message` for a row, and an update that removes the row before the
+  message is delivered, leaves the row's work simply not done. Where that
+  matters, keep the decision and the work in one reduce rather than splitting
+  them across a message.
+
 Building initial state records nothing: `Keyed::from_iter` and an insert into an
 absent key remove no instance, so growing a collection during `init` is fine.
 The four shapes that *do* record belong inside a `reduce`, where the boundary
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -138,15 +138,14 @@ itself.
 `TestStore` takes an `Application`, so a composed program is driven with
 `tears::testing::TestDriver` instead: it constructs from the same inputs the
 production entry point takes, boots the program, and steps whole passes, with
-the sends a producer makes released one grant at a time. Four tests in
-`dashboard_composed.rs` drive that example's own stack that way — the same
-`Reducer` value `main` runs, closed with the same `init` and `view`, and started
-from a `Setup` carrying a scripted input instead of the binary's seed. They are
-the ones whose names begin `identical_child_identities`, `every_removal`,
-`a_repeated_setup_message` and `replacing_a_row`, and they are the shape to
-copy. The rest of that file's rows call the root reducer and the key decoder
-directly, which is a cheaper loop and a different claim: it never crosses a
-boundary, so it would pass with the composition taken out.
+the sends a producer makes released one grant at a time. The rows in
+`dashboard_composed.rs` that build a `TestDriver` drive that example's own stack
+that way — the same `Reducer` value `main` runs, closed with the same `init` and
+`view`, and started from a `Setup` carrying a scripted input instead of the
+binary's seed. They are the shape to copy. The rest of that file's rows call the
+root reducer and the key decoder directly, which is a cheaper loop and a
+different claim: it never crosses a boundary, so it would pass with the
+composition taken out.
 
 Two observations are worth designing tests around, because they are what
 composition changes:
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -1,0 +1,139 @@
+# Composing Reducers
+
+`Application` and the composition API describe the same program. An
+`Application` is run through an adapter that makes the application value the
+state and `update` the `reduce`, on the same kernel a composed program runs on,
+so moving between them changes how a program is *written* and not how it is
+executed.
+
+This guide is about when that rewrite pays for itself. The worked example is
+one application written both ways: [`examples/dashboard.rs`](../examples/dashboard.rs)
+with the root owning the wiring, and
+[`examples/dashboard_composed.rs`](../examples/dashboard_composed.rs) with
+boundaries owning it. The reference for each item is the
+[`tears::reducer`](https://docs.rs/tears/latest/tears/reducer/) module.
+
+## Keep the `Application`
+
+A single `Application` is the right shape while the root can still answer
+every question about the whole program:
+
+- One state struct, or a root struct whose child structs are plain data the
+  root updates directly.
+- A `Message` enum the root matches exhaustively, forwarding child variants by
+  hand where it helps (`dashboard.rs` is this, at the size where it still reads
+  well).
+- Commands and subscriptions whose identities the root can keep distinct by
+  choosing distinct values — one timer per interval, one command id per
+  concern.
+
+Nothing here is improved by adding boundaries. `TestStore` drives an
+`Application` directly, which is the shortest test loop the crate offers.
+
+## Move to composed reducers
+
+The signal is not size, it is **repeated identities**. Compose when a child
+feature exists in more than one instance at a time, or comes and goes:
+
+- **A collection of children.** Every row wants the same subscription and the
+  same command id, and only the row it belongs to tells two of them apart.
+- **An optionally-present child.** A modal, a detail pane, an editor: its
+  subscriptions must start when it appears and stop when it is dismissed or
+  replaced, and the replacement's runs must not inherit the old occupant's.
+- **A child you want to write once and place twice.** A reducer over its own
+  state and message type is complete on its own; the boundary supplies the
+  projection and the message mapping at each placement.
+
+Hand-written, the first two are where the bugs are. Removing a row means
+remembering every command id and subscription that row had; replacing a modal
+means tearing down the old occupant *before* the new one starts. Both are
+correct-by-omission problems: the code that forgets them still compiles, still
+passes single-instance tests, and leaks a run per removal.
+
+## The three boundaries
+
+Each combinator wraps a parent reducer and one child, and the result is still a
+reducer over the root's state and message — which is why they chain.
+
+| Combinator | Child state lives in | Segment |
+| --- | --- | --- |
+| `scope` | a field of the parent's state | a fixed value you choose |
+| `for_each` | a `Keyed<K, ChildState>` | the row's key |
+| `presented` | a `Slot<ChildState>` | a fixed value you choose |
+
+`into_program` closes the stack with the two root-level functions composition
+has no place for: `init` and `view`.
+
+A child with no commands and no subscriptions of its own has no identities to
+qualify, so `scope` buys it code organisation rather than separation — which is
+a fine reason to reach for it, and a reason not to expect anything more.
+
+The outermost boundary sees a message first. What no boundary claims reaches
+the root reducer, so the root keeps exactly the messages that are its own.
+
+## What a boundary does, so you do not
+
+- **It qualifies identities.** Everything identity-bearing in the command a
+  child returned — spawn keys, explicit cancels, cleanup registrations — and
+  every subscription the child declares is qualified with that boundary's
+  segment. Two rows declaring the same timer are two subscriptions; two rows
+  keyed on the same command id occupy two slots. Application code writes no
+  `.scoped(...)`, and cannot omit or double-apply one.
+- **It tears removed instances down.** `Keyed` and `Slot` record a removal when
+  one happens — `Keyed::remove`, `Slot::dismiss`, and the two replacing shapes,
+  `Keyed::insert` over an occupied key and `Slot::present` over an occupied slot
+  — and the boundary turns each recorded removal into one teardown of that
+  instance's scope. The removed instance's subscriptions stop, its in-flight
+  commands are cancelled, and the cleanup hooks it registered run.
+
+Building initial state records nothing: `Keyed::from_iter` and an insert into an
+absent key remove no instance, so growing a collection during `init` is fine.
+The four shapes that *do* record belong inside a `reduce`, where the boundary
+drains them in the same update.
+
+## Three things stay at the root
+
+**`view` does.** `Reducer` has no `view` and only `Program` does: pane layout,
+draw order and area allocation are decisions about the whole frame, so
+composing child views is ordinary function calls over the root state.
+
+**`init`'s command does.** It is the root's command and crosses no boundary, so
+nothing it starts is scoped to a child. Work that belongs to a child — the
+first fetch, a cleanup hook that must anchor at the child's scope — starts as a
+message routed *through* the boundary. In the worked example that is
+`TaskMessage::Watch`: the root inserts the row and returns
+`Command::message(Message::Task(id, TaskMessage::Watch))`, and the row's own
+reduce registers its hook.
+
+The same rule explains `Command::on_teardown`'s placement. A registration
+anchors at the scope of the boundary it is built at, and one built at the root
+anchors where no teardown reaches it.
+
+**Work that spans two children does.** A child is handed its own projected
+state and nothing else, so it can reach neither the collection it sits beside
+nor the slot it sits in. Anything that touches two of them is a root message.
+In the worked example that is `Message::SaveNotes`: the details pane's edited
+notes are written onto the task the pane was opened for by the root, which then
+asks the row to sync through the boundary rather than starting that command
+itself.
+
+## Testing a composition
+
+`TestStore` takes an `Application`, so a composed program is driven with
+`tears::testing::TestDriver` instead: it constructs from the same inputs the
+production entry point takes, boots the program, and steps whole passes, with
+the sends a producer makes released one grant at a time. The two tests in
+`dashboard_composed.rs` drive that example's own stack — the same `Reducer`
+value `main` runs, closed with the same `init` and `view`, and started from a
+`Setup` carrying a scripted input instead of the binary's seed — and are the
+shape to copy.
+
+Two observations are worth designing tests around, because they are what
+composition changes:
+
+- The run identities a step started, which is where "these two rows did not
+  collide" is visible.
+- A cleanup hook's own side effect. `Command::on_teardown` takes a future whose
+  `Output` is `()`, so a finalizer sends no message; give it a sink the test can
+  read, and let the view render the same sink if the application wants to show
+  it.
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -138,11 +138,15 @@ itself.
 `TestStore` takes an `Application`, so a composed program is driven with
 `tears::testing::TestDriver` instead: it constructs from the same inputs the
 production entry point takes, boots the program, and steps whole passes, with
-the sends a producer makes released one grant at a time. The tests in
-`dashboard_composed.rs` drive that example's own stack — the same `Reducer`
-value `main` runs, closed with the same `init` and `view`, and started from a
-`Setup` carrying a scripted input instead of the binary's seed — and are the
-shape to copy.
+the sends a producer makes released one grant at a time. Four tests in
+`dashboard_composed.rs` drive that example's own stack that way — the same
+`Reducer` value `main` runs, closed with the same `init` and `view`, and started
+from a `Setup` carrying a scripted input instead of the binary's seed. They are
+the ones whose names begin `identical_child_identities`, `every_removal`,
+`a_repeated_setup_message` and `replacing_a_row`, and they are the shape to
+copy. The rest of that file's rows call the root reducer and the key decoder
+directly, which is a cheaper loop and a different claim: it never crosses a
+boundary, so it would pass with the composition taken out.
 
 Two observations are worth designing tests around, because they are what
 composition changes:
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -91,7 +91,7 @@ absent key remove no instance, so growing a collection during `init` is fine.
 The four shapes that *do* record belong inside a `reduce`, where the boundary
 drains them in the same update.
 
-## Three things stay at the root
+## What stays at the root
 
 **`view` does.** `Reducer` has no `view` and only `Program` does: pane layout,
 draw order and area allocation are decisions about the whole frame, so
@@ -105,12 +105,13 @@ message routed *through* the boundary. In the worked example that is
 `Command::message(Message::Task(id, TaskMessage::Watch))`, and the row's own
 reduce registers its hook.
 
-**Handle such a setup message idempotently.** Nothing guarantees it arrives
-once — a second key press decided against a state the first message has not
-been applied to yet produces a second — and a teardown fires *every*
-registration its scope holds, so a child that arms on each one reports two
-teardowns for one removal. A flag on the child's state is enough; the successor
-instance a replacement creates gets a fresh one.
+Such a setup message has to be handled *idempotently* by the child, which is
+the child's side of the same rule. Nothing guarantees it arrives once — a
+second key press decided against a state the first message has not been applied
+to yet produces a second — and a teardown fires *every* registration its scope
+holds, so a child that arms on each one reports two teardowns for one removal.
+A flag on the child's state is enough; the successor instance a replacement
+creates gets a fresh one.
 
 The same rule explains `Command::on_teardown`'s placement. A registration
 anchors at the scope of the boundary it is built at, and one built at the root
@@ -129,7 +130,7 @@ itself.
 `TestStore` takes an `Application`, so a composed program is driven with
 `tears::testing::TestDriver` instead: it constructs from the same inputs the
 production entry point takes, boots the program, and steps whole passes, with
-the sends a producer makes released one grant at a time. The two tests in
+the sends a producer makes released one grant at a time. The tests in
 `dashboard_composed.rs` drive that example's own stack — the same `Reducer`
 value `main` runs, closed with the same `init` and `view`, and started from a
 `Setup` carrying a scripted input instead of the binary's seed — and are the
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -85,7 +85,6 @@ the root reducer, so the root keeps exactly the messages that are its own.
   — and the boundary turns each recorded removal into one teardown of that
   instance's scope. The removed instance's subscriptions stop, its in-flight
   commands are cancelled, and the cleanup hooks it registered run.
-
 - **It discards what it cannot route.** A message addressed to a key the
   collection no longer holds, or to a slot with no occupant, reaches no reducer
   and is dropped — with no diagnostic, and with no way for the sender to learn
--- a/docs/composition.md
+++ b/docs/composition.md
@@ -105,6 +105,13 @@ message routed *through* the boundary. In the worked example that is
 `Command::message(Message::Task(id, TaskMessage::Watch))`, and the row's own
 reduce registers its hook.
 
+**Handle such a setup message idempotently.** Nothing guarantees it arrives
+once — a second key press decided against a state the first message has not
+been applied to yet produces a second — and a teardown fires *every*
+registration its scope holds, so a child that arms on each one reports two
+teardowns for one removal. A flag on the child's state is enough; the successor
+instance a replacement creates gets a fresh one.
+
 The same rule explains `Command::on_teardown`'s placement. A registration
 anchors at the scope of the boundary it is built at, and one built at the root
 anchors where no teardown reaches it.
--- a/examples/dashboard.rs
+++ b/examples/dashboard.rs
@@ -665,6 +665,16 @@ mod tests {
         Message::Terminal(Event::Key(KeyEvent::new(code, KeyModifiers::empty())))
     }
 
+    /// The same for the release a terminal that reports both would send after
+    /// it.
+    fn key_release(code: KeyCode) -> Message {
+        Message::Terminal(Event::Key(KeyEvent::new_with_kind(
+            code,
+            KeyModifiers::empty(),
+            KeyEventKind::Release,
+        )))
+    }
+
     /// Sending the child message directly walks focus through the panels.
     #[test]
     fn focus_next_cycles_through_panels() {
@@ -734,6 +744,40 @@ mod tests {
         store.send(key(KeyCode::Char('q')));
         store.receive_matching(|msg| matches!(msg, Message::Quit));
         store.receive_quit();
+        store.finish();
+    }
+
+    /// The details panel is a text field, so 'q' is a character there rather
+    /// than a request to quit. Tab is how you leave that focus.
+    #[test]
+    fn quit_key_is_text_while_the_details_panel_has_focus() {
+        let mut store = TestStore::<App>::new(());
+        store.send(Message::FocusNext);
+        store.send(Message::FocusNext);
+        assert_eq!(store.state().focus, Focus::Details);
+        let notes = store.state().details.notes.clone();
+
+        store.send(key(KeyCode::Char('q')));
+        store.receive_matching(|msg| matches!(msg, Message::Details(DetailMessage::Input('q'))));
+
+        assert_eq!(store.state().details.notes, format!("{notes}q"));
+        store.finish();
+    }
+
+    /// A terminal that reports releases as well as presses sends two events
+    /// per keystroke. Only the press is an instruction.
+    #[test]
+    fn a_key_release_asks_for_nothing() {
+        let mut store = TestStore::<App>::new(());
+        assert_eq!(store.state().focus, Focus::Navigation);
+
+        store.send(key_release(KeyCode::Tab));
+
+        assert_eq!(
+            store.state().focus,
+            Focus::Navigation,
+            "the release is not a second Tab"
+        );
         store.finish();
     }
 }
--- a/examples/dashboard.rs
+++ b/examples/dashboard.rs
@@ -168,10 +168,9 @@ impl Application for App {
             // Not releases. A terminal that reports them — Windows, or a
             // session with the kitty keyboard protocol on — sends two events
             // for one press, and the second would be acted on. A *repeat* is
-            // kept, and is the other thing entirely: a held key is repetition
-            // the reader is asking for. Testing for `Press` would drop those
-            // with the releases, and a held Down would move the selection
-            // once.
+            // kept: a held key is repetition the reader is asking for, and
+            // testing for `Press` would drop those with the releases, leaving
+            // a held Down to move the selection once.
             Message::Terminal(Event::Key(key)) if key.kind != KeyEventKind::Release => {
                 return handle_key_event(self.focus, key);
             }
@@ -654,8 +653,9 @@ impl StatusState {
     }
 }
 
-/// Every binding here is for an unmodified key. Raw mode delivers no SIGINT, so
-/// a reader pressing Ctrl+C to leave would otherwise have run whatever `c` is
+/// Every binding here is for an unmodified key, apart from a shifted character,
+/// which is how an uppercase letter arrives. Raw mode delivers no SIGINT, so a
+/// reader pressing Ctrl+C to leave would otherwise have run whatever `c` is
 /// bound to — clearing the activity log — and Ctrl+D would have deleted a task.
 fn handle_key_event(focus: Focus, key: KeyEvent) -> Command<Message> {
     // The way out of raw mode, from any focus: the details editor holds `q`
@@ -663,14 +663,18 @@ fn handle_key_event(focus: Focus, key: KeyEvent) -> Command<Message> {
     if key.code == KeyCode::Char('c') && key.modifiers.contains(KeyModifiers::CONTROL) {
         return Command::message(Message::Quit).into();
     }
-    // Every binding below is for an unmodified key, so a chord this
-    // application does not bind is answered here rather than falling through
-    // to the bare key's. Above the match and not on the character arms,
-    // because `Enter`, `Up`/`Down`, `Tab` and `Esc` arrive with modifiers too.
-    if key
-        .modifiers
-        .intersects(KeyModifiers::CONTROL | KeyModifiers::ALT)
-    {
+    // Every binding below is for an unmodified key, with one exception: a
+    // shifted character on its way into the details editor, which is how an
+    // uppercase letter arrives. Anything else carrying a modifier is a chord
+    // this application does not bind and must not fall through to the bare
+    // key's. Testing for `CONTROL` and `ALT` alone leaves four of the six
+    // through — Shift+Down would move the task selection. Letting `SHIFT`
+    // through generally is no better: it would reach the task bindings, where
+    // `d` deletes.
+    let shifted_text = focus == Focus::Details
+        && matches!(key.code, KeyCode::Char(_))
+        && key.modifiers == KeyModifiers::SHIFT;
+    if !key.modifiers.is_empty() && !shifted_text {
         return Command::none();
     }
     match key.code {
@@ -897,13 +901,14 @@ mod tests {
         store.finish();
     }
 
-    /// A `CONTROL` chord runs no bare-key binding, and Ctrl+C leaves.
+    /// A modified key runs no bare-key binding, and Ctrl+C leaves.
     ///
     /// Raw mode delivers no SIGINT, so Ctrl+C arrives as an ordinary key
     /// event; without the guard it would have cleared the activity log, and
-    /// Ctrl+D would have deleted the selected task.
+    /// Ctrl+D would have deleted the selected task. Shift+Down would have moved
+    /// the selection.
     #[test]
-    fn a_control_chord_runs_no_bare_binding_and_ctrl_c_leaves() {
+    fn a_modified_key_runs_no_bare_binding_and_ctrl_c_leaves() {
         let mut store = TestStore::<App>::new(ExitReason::default());
         store.send(Message::FocusNext);
         assert_eq!(store.state().focus, Focus::Tasks);
@@ -919,7 +924,13 @@ mod tests {
             KeyCode::Tab,
             KeyCode::Down,
         ] {
-            for modifiers in [KeyModifiers::CONTROL, KeyModifiers::ALT] {
+            for modifiers in [
+                KeyModifiers::CONTROL,
+                KeyModifiers::ALT,
+                KeyModifiers::SHIFT,
+                KeyModifiers::SUPER,
+                KeyModifiers::META,
+            ] {
                 store.send(chord_with(code, modifiers));
             }
         }
--- a/examples/dashboard.rs
+++ b/examples/dashboard.rs
@@ -254,7 +254,7 @@ impl App {
                     task.notes.clone_from(&self.details.notes);
                     self.activity
                         .push(format!("updated notes for {}", task.title));
-                    self.status.set_info("Saved task notes");
+                    self.status.set_info(outcome.status);
                 } else {
                     // Silence here would leave the footer reporting the edit
                     // that was not saved.
@@ -976,6 +976,30 @@ mod tests {
             store.state().status.message,
             "No task selected",
             "the save wrote nothing and says so"
+        );
+        store.finish();
+    }
+
+    /// Saving reports through the outcome the panel returned rather than a
+    /// copy of it, so the two cannot drift.
+    #[test]
+    fn saving_reports_the_panel_s_own_status() {
+        let mut store = TestStore::<App>::new(ExitReason::default());
+        store.send(Message::FocusNext);
+        store.send(Message::FocusNext);
+        assert_eq!(store.state().focus, Focus::Details);
+
+        store.send(Message::Details(DetailMessage::Input('x')));
+        store.send(Message::Details(DetailMessage::Save));
+
+        assert!(
+            store.state().tasks.tasks[0].notes.ends_with('x'),
+            "the edit reached the task"
+        );
+        assert_eq!(
+            store.state().status.message,
+            "Saved task notes",
+            "and the footer says what the panel asked it to"
         );
         store.finish();
     }
--- a/examples/dashboard.rs
+++ b/examples/dashboard.rs
@@ -686,7 +686,7 @@ impl StatusState {
 fn handle_key_event(focus: Focus, key: KeyEvent) -> Command<Message> {
     // The way out of raw mode, from any focus: the details editor holds `q`
     // but nothing holds this.
-    if key.code == KeyCode::Char('c') && key.modifiers.contains(KeyModifiers::CONTROL) {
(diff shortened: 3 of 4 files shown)


akiomik:
All five addressed in 931e0fd and f128d1f. `take_under` returning every
registration under the prefix is confirmed, so #2 was real as described.

**1 — stale `TaskId`.** Fixed as suggested: the reduce allocates, `AddTask`
carries a label only. Worth recording that batching is not even required — one
extra hop is enough, because the second `Terminal` is decoded before the first
`AddTask` is *delivered*. The reduce appends the key it chose to the label, so
two `n` presses no longer produce two rows with the same display name either.
The variant's doc says why the caller does not get to choose the key.

**2 — doubled setup hook.** `TaskMessage::Watch` and `DetailsMessage::Open` are
now idempotent on a flag the successor instance of a replacement gets fresh.
New test `a_repeated_setup_message_arms_one_hook` covers exactly the sequence
described — both `present` calls land before either `Open` is delivered, so the
replaced occupant reports nothing and both `Open`s reach the survivor. Dropping
the guard reproduces the symptom:

```
left:  ["closed details: alpha #1", "closed details: alpha #1"]
right: ["closed details: alpha #1"]
```

**3 — `ReloadTask` leaving a stale pane open.** It now closes a pane opened on
the instance it replaces, through the same helper `DeleteTask` uses. New test
`replacing_a_row_closes_a_pane_opened_on_it`; dropping the rule exhausts its
`settle` budget. That test asserts which hooks fired and not their order,
because both teardowns come from one reduce and the finalizers finish in
executor order.

**4 — focus dead end.** The helper restores focus as well as dismissing, so all
three dismissal paths — `CloseDetails`, the implicit one in `DeleteTask`, and
now `ReloadTask` — go through it.

**5 — unguarded `q`.** Guarded in both files. It is not newly introduced in
`dashboard.rs`, but the panel is an editable text field there too, and fixing
only one would make the pair differ in something other than structure, which is
the one thing the pair is for.

One knock-on worth flagging: fixing #3 changed the script order in
`every_removal_tears_its_instance_down` — `CloseDetails` now runs before
`ReloadTask`. With the pane still open, `ReloadTask` fires two finalizers in one
pass and the ordered `assert_eq` would have become a race between them. The
reordering keeps one teardown per pass, so the assertion stays an order claim
rather than a set one; the comment in the script says so.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass.

akiomik:
All seven addressed in 83675e8 and 57701b6.

**A — `q` dead on an empty pane.** Fixed as suggested, via a named predicate:

```rust
fn editing_notes(state: &App) -> bool {
    state.focus == Focus::Details && state.details.is_present()
}
```

`is_present()` alone would have been wrong in the other direction — it would
stop `q` quitting while a pane is open and the focus is on Tasks — so the guard
is the conjunction, and the other editor keys (`Char`, `Backspace`, `Enter`) now
ask the same question rather than the focus. The regression row asserts all
three positions: empty pane at Details focus quits, occupied pane at Details
focus types, occupied pane at Tasks focus quits. Reverting the predicate to the
focus alone fails its first assertion, which is the state f128d1f shipped.

**B — false comment in `dashboard.rs`.** Correct: Esc there maps to
`DetailMessage::Reset` and never touches the focus, and the panel is always
present, so there is nothing to leave. The comment now says Tab is the only way
out of that focus and that Esc reloads the notes. The claim was true of
`dashboard_composed.rs` and was copied to a file where it is not. f128d1f's
message carries the same wrong sentence; I have left the commit alone rather
than rewriting a branch you have already reviewed, and 83675e8 states the
correction.

**C, D — stale counts.** Both dropped rather than updated, along with the same
sentence in the PR body. The number was doing no work and I had already made it
wrong once.

**E — heading counted four.** The idempotence rule is the child's side of the
`init` rule, not a fourth thing the root keeps, so it is prose under that item
now and the heading no longer counts.

**F — `key.kind`.** Added to both, matching `http_todo.rs` and `websocket.rs`.
Agreed on the reasoning: inherited or not, a doubled `AddTask` or `OpenDetails`
produces exactly the spurious teardown lines 931e0fd removed, in the file whose
subject is which removals produce which teardowns.

**G — divergent seed titles.** Resolved rather than accepted: the key moved out
of the title and onto the row's state, which a child needs anyway to name itself
in a hook it registers, and the panes and log lines print it beside the title
(`#1 alpha`). Seeded titles are now character-for-character `dashboard.rs`'s.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass.

akiomik:
All four addressed in 1edbe26 and dab85f1.

**A — "next scale" vs. the entry below it.** Reworded as suggested: the first
three examples are a scale, the fourth is the third one rewritten, and it is
read *beside* `dashboard.rs` rather than after it.

**B — `scope` does not tear down.** Correct, and the README was the only place
that said otherwise. Split as recommended: every boundary qualifies identities,
and the two that hold a collection additionally tear a removed child's runs
down. The clause names why `scope` is not among them — one child in place, no
collection for an instance to leave — so the omission does not read as an
oversight.

**C — more than the wiring differs.** Named both differences where the
invitation to diff is. The keys, as you listed them: `r`, Enter from the task
list, and Esc closing the pane rather than reloading the notes. And one you did
not list that is larger: every child here has runs of its own, a timer and a
keyed request, where `dashboard.rs` has neither — its only `Subscription::new`
is the root's terminal source and it constructs no effect at all. That one is
load-bearing rather than incidental, since qualification and teardown are about
runs, so a child with none would give a boundary nothing to do; the doc now says
so rather than leaving a reader to find it in a diff.

**D — Esc claiming a close that did not happen.** Took the first branch: Esc
stays guarded on the focus, because a `Details` focus with an empty slot behind
it still needs a way out, and the status now follows what `close_details`
actually did — `"Closed the details pane"` only when an occupant was dismissed,
`"Left the details pane"` otherwise. Reporting a close for an empty slot
described a teardown that never happened, in the file whose subject is which
removals produce which teardowns. Covered by
`escaping_an_empty_pane_leaves_it_without_claiming_a_close`, which fails if the
status goes back to being unconditional.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass.

akiomik:
All seven addressed in 28605e4 and 8b7d9b3.

**A — "Every child here has runs of its own".** Correct, and the same file
contradicted it: `Navigation` and `Activity` declare no subscription and return
no effect, and `Navigation`'s own doc says so. Narrowed to the row children and
the pane's occupant, and the other two are now named in the same paragraph as
the case where `scope` buys organisation rather than separation — which is worth
saying rather than working around, since it is the honest reason a boundary can
be worth adding to a child with nothing to qualify.

**B, C — undeclared behavioural differences.** Took the first branch for both:
aligned to `dashboard.rs` rather than growing the doc's list. The selection
clamps, and a delete keeps the position, so the row that moved up is selected
and a deleted last row falls back to the new last. The only part that stays
composition's is that the position is expressed over keys instead of indices.
`the_selection_clamps_and_survives_a_delete_in_place` covers the four positions
that tell the two behaviours apart; restoring either the wrap or the jump fails
it.

**D — `close_details`'s doc.** Both halves rewritten. It returns whether there
was an occupant to dismiss, which is what the status branch reads; and the focus
reset guards nothing now that `editing_notes` asks the slot — it only keeps the
focus off a pane the user cannot see. That sentence had gone stale exactly when
`editing_notes` landed, as you say.

**E — `Watch`'s re-entry path.** Made inert, matching `Open`. Thank you for the
correction; the asymmetry is the part that matters, and "the run does not leak"
was answering a different question than the one the doc's **Handled
idempotently** raises. The first `Watch` an instance sees still asks for the
sync, so nothing is lost.

**F — indistinguishable `synced:` lines.** `Sync` now names the row the way the
teardown line does (`#1 alpha`). The `n` label stays `"New task"`, so the titles
still match `dashboard.rs` character for character.

**G — header advertising a key the editor has taken.** Both headers now ask the
question their key handler asks. The composed one offers `esc: close pane` while
an occupant has the keyboard; `dashboard.rs`, whose panel is always present,
offers `Tab: leave the editor`.

One thing that came out of this rather than from the review: extracting the
collection arms into functions is what keeps `Root::reduce` under
`clippy::too_many_lines` after the delete-position change. The bodies are
unchanged, and the commit says so.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
All four addressed in 41870ea, with C and D taken further than "add a clause".

**A, B — row-scoped lines without the key.** Both fixed. `save_notes` writes
`updated notes for #1 alpha` and the occupant's load writes
`loaded details: #1 alpha`. Thank you for B; the previous pass fixed `synced:`
and stopped at the one line the finding named, which is the same partial-sweep
mistake twice in a row on my part.

**C, D — an enumeration claiming completeness.** Took both halves rather than
just the doc:

- The activity log now carries the entries `dashboard.rs` writes and this file
  did not — `added:` on a task, `completed:`/`reopened:` on a toggle. The log
  differs by the teardown lines, which is the difference worth seeing, rather
  than by an arbitrary subset.
- The layout constants are `dashboard.rs`'s. They differed for no reason, in a
  file whose doc sends readers to diff them.
- The doc list is no longer a closed set. It is what is worth knowing before
  reading either file, it names the log and the message-seeded tasks alongside
  the runs and the keys, and it moved after the explanation that gives a reader
  the vocabulary for it instead of interrupting the first paragraph.
- The PR body's "same panes, same keys, same activity log" is rewritten. It now
  says what is genuinely the same and points at the module doc for what is not,
  so the claim has one home rather than two that can drift.

The seed rows stay as they are, per your reading: uniform notes and none marked
done follow from the tasks arriving as messages, which follows from `init`'s
command crossing no boundary. That is now a line in the list rather than
something a diff turns up.

One consequence worth flagging: the new `added:` lines put reduce-written
entries in front of the teardown lines the tests assert on. The rows whose
subject is the teardowns read them through a filter now;
`every_removal_tears_its_instance_down` still asserts the whole log, because the
interleaving of a reduce's own `deleted:` line with the hook fired by the
teardown that same reduce originated is part of what that row documents.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
All five addressed in a17b370 and 15cbe2b.

**A — the status line silent on three operations.** Fixed as recommended: the
root sets the status where it dispatches the child message, in the `Terminal`
arm that decoded the key. Worth naming what that costs, because it is
composition's and not an oversight — the line can only name the *request*
("Toggled the task") where `dashboard.rs` names the *outcome* ("Marked task as
completed"), since the outcome is the child's and the child has not run when the
root sets the line. Both the constant and the doc list say so. Covered by
`a_child_addressed_key_still_moves_the_status_line`; removing the dispatch-time
assignment reproduces the stale line exactly as you described it:

```
left:  "Added a task"
right: "Selected another section"
```

**B — "as a result" not reaching its claim.** Correct, and the substitute is the
one you name: `AddTask` carries a title and nothing else. The doc now says a
setup message carrying more would cross the same boundary just as well, so it
stops implying a constraint composition does not impose.

**C — silent discard.** Added to "What a boundary does, so you do not", with the
sequence you gave: a root asking a row for work through `Command::message`, and
an update removing the row before it is delivered. Also the reason it belongs
there rather than only in the library rustdoc — a reader who has not been told
looks for the bug in the child.

**D — dangling relative clause.** Borrowed the guide's word order, which has
nowhere for it to attach wrongly.

**E — two log entries with no counterpart, and a stray capital.** The list now
says the log gains *two* kinds this file has and `dashboard.rs` cannot: the
teardown lines, and the completion lines the children's runs write (`synced:`,
`loaded details:`), which have no counterpart because that file runs no
commands. `dashboard.rs`'s notes entry was the one capitalised and quoted line
in either file and is now neither, like its five neighbours. The PR body's "same
activity log entries" is gone; it now claims only what is genuinely the same and
points at the module doc for the rest.

Thank you for the table — enumerating every `push` site is what turned E from a
guess into a check, and it is the form I should have used when I claimed the
logs matched.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
All five addressed in 5d0ee6b and 75863ba, and thank you for the retraction —
the ordering argument is right, and the status write on the removal always lands
after the dispatch write.

**A — the press guard's reasoning.** This one is mine twice over: I moved the
key allocation into the reduce in 931e0fd and then, in 8b7d9b3, wrote a comment
justifying a *different* guard with the counts from before that move. Both
halves were wrong as you say — two `AddTask`s are no removals, and two
`OpenDetails` are one removal and no teardown line, which
`a_repeated_setup_message_arms_one_hook` asserts a thousand lines down. The
guard stands; the comment now says what doubling actually costs (a row the
reader did not ask for, and a pane replaced by a second pane on the same row).

**B — the two `dashboard.rs` changes untested.** Correct, and the mechanism is
exactly as you diagnosed: `quit_key_requests_shutdown` never reaches
`Focus::Details`, and the `key` helper's `KeyEvent::new` stamps
`KeyEventKind::Press`, so nothing reached the filter. Two rows now do, each
mutation-checked — dropping the guard makes the store see a `Quit` where the row
expects `Input('q')`, and dropping the filter lets a release move the focus a
second time.

**C — `reload_task` writing no line.** Added `reloaded: #{id} {title}`. Your
repro is the argument: clearing `done` without a line leaves `completed:`
standing above a checkbox that has gone.

**D — "three operations" against four arms.** Third countable mismatch on this
branch, so the sentence no longer counts anything.

**E — "names the request" against past-tense strings.** Took the second option.
The strings are ordinary UI text and reworking them into a gerund voice would
buy nothing; what is actually true is that the line is fixed before the child
runs, so nothing the child computes can appear in it. Both the doc bullet and
`child_status`'s rustdoc say that instead.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
All five addressed in 5681753 and ce56fbd.

**A — Esc, wrong twice.** Took your recommendation and fixed the code rather
than the comment. `DetailMessage::Reset` now rereads the selected task's notes,
which is what the variant's name, the panel's hint, and my comment all already
claimed. The panel cannot reach the task, so the outcome carries a
`DetailAction` — `Keep`, `Save`, `Reset` — where it carried a `save_requested`
flag, and the root does the reread the way it already did the save; an enum
rather than a second bool, since only one of the three can be wanted at a time.

That choice is the point. A comment corrected twice is a comment that will be
wrong a third time; a behaviour a row asserts is not. `escape_rereads_the_selected_task_s_notes`
fails if `Reset` goes back to setting only the status:

```
left:  "Check docs, examples, and CI status.x"
right: "Check docs, examples, and CI status."
```

**B — the same claim repeated as a baseline.** True now, and updated to say
"rereads the selected task's notes" rather than "reloads the notes".

**C — "two kinds" undercounting.** Fourth countable mismatch on this branch, so
that item names the kinds and stops counting them. Both of the ones you found
are in it: `reloaded:`, for a key `dashboard.rs` does not have, and the
terminal-error line, which goes to stderr there.

**D — the details buffer following the selection.** Added, and it earns its
place better than most of the list: `dashboard.rs` rereads the panel on *every*
task message, so moving the selection throws an unsaved edit away, while a
`Slot`'s occupant is fixed to the task it was opened for. That is composition
showing through — a slot holds an instance rather than a view of whatever is
selected — rather than an incidental difference.

**E — the assertion on the eviction boundary.** Went with the filtered view:
that row now compares without the script's own `added:` lines, which were setup
rather than subject. Raising `ACTIVITY_LIMIT` would have reintroduced a
difference from `dashboard.rs` to fix a test's framing, which is the wrong
direction on this branch.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
Both fixed in 0ec05cb.

**A — `select` silent on an empty list.** The assignment moved above the guard.
The row that covers the ends covers the empty list too now, and without the move
it reproduces your repro exactly:

```
left:  "Deleted the task"
right: "Selected another task"
```

**B — loose list.** Blank line removed. I also swept the file for the same shape
elsewhere and there is none.

Two things from your verification worth recording, because they close gaps I had
left open rather than confirming what I already knew:

- The `KeyEventKind::Press` guard's justification was mine by assertion — I
  named Windows and the kitty protocol from the other examples' precedent
  without checking what either actually delivers. Your read of
  `event/sys/windows/parse.rs` and of `REPORT_EVENT_TYPES` is what makes the
  guard's *safety* — that it drops nothing a user can generate — a checked
  claim rather than a plausible one.
- Likewise the FIFO lane ordering behind the module doc's "Enter would open the
  pane and then replace it": I reasoned that from the message hop and did not
  trace `src/kernel/lane.rs`. That the release beats the reduce's own
  `Command::message` is the part that makes the sentence true, and it is now
  checked.

The wall-clock independence of `every_removal_tears_its_instance_down` was a
property I relied on when I chose a 900 ms stub — the sends are grant-gated, so
nothing in that row waits on real time — but I stated it in a commit message
rather than establishing it. Good to have it verified against the driver's
executor rather than taken on my word.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
All three fixed in 77396d9.

**A — the guard dropped `Repeat`.** This is the sharper form of a mistake I made
two rounds ago: I named the kitty protocol as the guard's justification without
checking what it sends, and it sends exactly the kind the guard was discarding.
`!= Release` now, which is the rule the comment states. A row in each file
covers both kinds; narrowing it back to `== Press` fails them.

Agreed that the impact is zero today — nothing in `src/` or `examples/` pushes
`PushKeyboardEnhancementFlags`, so `Repeat` never arrives. The defect is that
the code and its stated reason disagreed, and the reason was the part I had
already been corrected on once.

**B — silent no-op on an unselected row.** Fixed as recommended, aligned with
`select`. Space, `d`, `r` and Enter decode to a `NoTaskSelected` message the
root reports, rather than to nothing. `dashboard.rs`'s two strings for this
("No task selected" / "No task to delete") are now one, so the pair reports the
same thing.

**C — "because it runs no commands".** Correct: `handle_key_event` returns
`Command::message` from every arm and `Command::quit` from two. The claim rests
on nothing there completing with a value later, and that is what it says now.

One correction on the citation, since you checked it and reported it missing:
the line *is* in the file. `examples/dashboard.rs` around the test module's doc
reads "the key handler produces follow-up | messages via `Command::message`" —
wrapped across two lines, so a single-line grep for the whole phrase finds
nothing. The review agent quoted it accurately. It does not change C, which
stands on the code either way, and it is worth flagging only because a grep that
finds nothing reads as evidence of absence.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
All three fixed in 4ab83c2.

**A — a reason that did not reach its claim.** Correct, and the contradiction
with the guide is the clearest way to see it: `Keyed::from_iter` in `init` is
something `docs/composition.md` explicitly blesses, so a bullet implying it is
unavailable is teaching the opposite of the file next to it. What `init`'s
command cannot do is start work *under a child's scope* — that is why the row's
setup is a message — and it says nothing about how the row itself gets there.
The rows are messages so the seed and the `n` key take one path, which is the
actual reason and is now the stated one.

**B — `done` and saved notes dropped in silence.** Took your reading: not
carrying them across is the correct teaching, since a replacement tears the old
instance down and starts the successor fresh, and that is the subject of the
file. So the fix is saying it rather than changing it. `ReloadTask`'s doc says
it, and the status line says it where someone driving the binary can see it,
because a cleared checkbox was the only other sign.

**C — the comment's first half outliving the change to its second.** Mine, and
the mechanism is worth naming: I widened the guard in the previous commit and
edited only the sentence about the widening, leaving the sentence about the
original narrow rule in place. The filter removes a second *event* for one
press; a held key is repetition the reader is asking for, and holding Enter does
replace the pane, over and over. The two are distinguished now rather than
conflated under "performed once".

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree.

akiomik:
Both fixed in 143389e.

**A — Ctrl chords running bare bindings.** Fixed in both files with the shape
`command_cancellation.rs` uses. Ctrl+C leaves from any focus, including the
notes editor that holds the bare `q`; every other `CONTROL` chord decodes to
nothing rather than falling through. A row in each file covers both halves and
fails without the guard.

Agreed that the composed file is where this matters most, and for the reason you
give: `d` and `r` are this branch's own additions, and `r` discards a row's done
flag and its saved notes. `dashboard.rs` takes the same guard, since leaving it
would put the pair out of step on exactly the axis this PR claims they are not.

**B — the silent save.** `else` added, reporting "No task selected" — the string
its sibling reports for the same condition. Worth recording how it survived: I
rewrote that arm in this branch when I introduced `DetailAction`, so I had the
`if let` open in front of me and did not ask what the absent branch did, in the
same session where I was adding `NoTaskSelected` to the other file for precisely
this failure mode.

**On the test counts.** Yours is right and the agent's was not: composed was 11
before this commit and `dashboard` was 8. This commit takes them to 12 and 10.

Verification on the pushed tree: `cargo test --lib --tests --examples`,
`just test-doc`, `cargo clippy --all-targets`, `cargo fmt --check`, `typos`,
`RUSTDOCFLAGS=-D warnings cargo doc` — all pass.

Also noted, with thanks: the `presented`-replacement experiment is a check I
should have run myself when I wrote "its timer stops and its in-flight sync is
cancelled" into the module doc. A fixed segment shared by an occupant and its
successor is exactly where that sentence could have been false, and I asserted
it from the teardown contract rather than from a run.

akiomik:
Both fixed in 9b770e8, and thank you for the retraction — the focus move in
`open_details` is the part I missed, and it is what makes a held Enter open once
and then save.

**A — the comment overstating a held key.** Rewritten. It now distinguishes the
two cases rather than offering the wrong one as the example: the duplicate the
filter removes is a press and its release, both decoded before either lands, so
both see the focus that opens; a held Enter is not that.

Worth naming how this one arrived, since it is the second time in three rounds:
I widened the guard, then wrote a justification for the widening without
re-deriving what the widened behaviour actually is.

**B — the terminal error reported to a screen about to disappear.** Went past
the recommendation, because `eprintln!` from inside the run is written while the
alternate screen is up and goes with it — the same failure the finding names,
one channel over, and `dashboard.rs`'s existing `eprintln!` is not obviously
better off. `main` prints the recorded line *after* `ratatui::restore()`,
reading the activity log it already holds a handle to. That is a report that
survives the screen, which the reduce cannot provide from inside the run.

A row covers the half the program owns — that the reason is written down at all
— and dropping the record fails it. The module doc's framing of this as a
deliberate stderr difference is replaced by what each file actually does.

`dashboard.rs` keeps its in-run `eprintln!`. Making its error survive the
restore needs a handle its `Application` does not have, so it is a change of
shape rather than a line, and it is pre-existing rather than this branch's. Say
the word if you want it in scope and I will thread one through; otherwise it is
worth its own issue.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree. Composed is 13 tests now, `dashboard` 10.

akiomik:
All six fixed in 0a77de2.

**A, B — the guard on two arms, the comment on all of them.** Lifted above the
match in both files, with Ctrl+C kept in front. `ALT` is included, so Alt+q no
longer quits and Alt+c no longer clears the log. Ctrl+Enter is the one that
mattered most here for the reason you name: it opens the pane, which replaces an
occupant, which is a removal — in the file about which removals happen.

Your retraction is the useful part of this one. "Not destructive" was the wrong
question to ask of `OpenDetails` in *this* file, and it is the same question I
had been getting wrong in the other direction all round — treating a replacement
as an update rather than as a removal.

The rows are widened to what their names claimed: every binding, character or
not, under both modifiers. They failed against the previous guard.

**C — `Reset` reporting a reload of nothing.** `else` added, reporting the
string the `Save` arm beside it reports. Covered.

**E — `dashboard.rs`'s error still lost.** Taken into scope, since your citation
settles what I was uncertain about when I raised it: `init.rs:400` enters the
alternate screen and `:558` leaves it, so an in-run `eprintln!` goes with the
screen. `App` gains the handle its sibling has — `ExitReason` as its `Flags` —
and `main` prints after `restore()`. Both files now report a terminal error the
same way.

**D — a reason that was not the operative one.** Rewritten: the entries eviction
would drop first are the two the filter already removes, so the failure would
surface as an unexpected extra rather than a missing head. The filter stays,
for the reason that does hold.

**F — the two ends joined by a literal.** `TERMINAL_ERROR` is a constant both
sides use, so a change to the recorded format cannot silently stop `main`
matching it.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree. Composed is 13 tests, `dashboard` 11.

akiomik:
All four fixed in 44d354e. Took C first as you suggested, and it did decide A.

**C — one job, two mechanisms.** Aligned on `ExitReason`, which is the better of
the two rather than merely the one that was there first: the log's route is
bounded by `ACTIVITY_LIMIT` and joined to `main` by a string prefix that only a
constant kept honest, and an `Option<String>` is neither. The constant and the
prefix scan are gone with it.

The mechanism split is mine from one round ago: I gave `dashboard.rs` the handle
while leaving the composed file on the log, in a commit whose whole subject was
making the two report the same way, and did not go back to the file I had
already written.

**A — the difference list describing what it had just changed.** Deleted rather
than rewritten: with both files placing the reason the same way there is no
difference left to list.

**B — "the tests … are the shape to copy", of four in thirteen.** Named the
four, and said what the other nine are: they call the root reducer and the key
decoder directly, which crosses no boundary and would pass with the composition
taken out. That is worth stating rather than eliding, since a reader following
"copy this" into `the_selection_clamps_and_survives_a_delete_in_place` learns the
wrong lesson about what a composition test proves. I verified the four names
against the four `scripted()` call sites rather than from memory.

**D — `App` losing `Debug`.** Restored; `ExitReason` derives it. You are right
that it was the only struct in the file without one, and only because this
branch took it away.

On the line number: the guide is 155 lines and the sentence is at 141. Noted for
what it says about a citation I cannot check by opening the file myself —
same class as the `follow-up messages` grep two rounds ago, in the other
direction.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the pushed tree. Composed is 13 tests, `dashboard` 11.

akiomik:
Pausing this PR pending #362.

The three findings above are real and cheap, but fixing them one at a time is
what has not been converging. Sorting the sixteen rounds gives 66 findings, and
by my classification 4 of them are about `scope`/`for_each`/`presented`/
`into_program`. The other 62 are the TUI shell built to host those four, and the
prose explaining it.

The prose half has one shape: every claim that failed had its truth somewhere
other than the lines it annotated — a framework item's semantics restated
locally, the other example's behaviour, what the same file does three functions
away, counts. The prose that never failed describes the statement below it, and
`docs/composition.md`, the densest prose here, has had almost no findings
because its subject is the library.

About ten of the 66 exist only because of my own earlier fixes on this branch:
the press-filter comment wrong in three consecutive rounds, the `ExitReason`
split created by the commit whose subject was unifying that reporting, the
`CONTROL` guard fixed twice, the `q` guard's fix generating the next round's
finding. Each fix added prose and shell behaviour, so I was widening the surface
while clearing it.

#362 fixes the rule rather than the instances: where each kind of claim belongs,
and that a claim about what the file does elsewhere in itself becomes a test
rather than a comment. The evidence for that being the operative move is this
PR's own record — claims moved into tests stayed fixed, claims corrected in
prose came back.

Two decisions taken with the maintainer and applied here once #362 settles:

- The comparison between the two files lives in `dashboard_composed.rs`'s module
  doc and nowhere else. `dashboard.rs` keeps a pointer and asserts nothing about
  the other file. Today's A and C, and the previous round's A and C, are all
  prevented by that alone.
- The composed example states its own behavioural rules in its module doc and
  applies each everywhere, instead of answering each situation as it comes up.
  Today's B is rule four of four — "a replacement says what it discarded" —
  already applied in `reload_task` and not in `open_details`. Six findings across
  the sixteen rounds were this same rule discovered separately.

I will hold this branch until #362 is settled, then apply the result here in one
pass rather than continuing round by round.

akiomik:
Rebased onto `main` now that #362 has landed, and the three findings from the
round this branch paused on are resolved.

**The rebase was clean.** #362's README changes and this branch's do not overlap:
it rewrote the lead-in above the `cargo test --example` block and added a
Contributing paragraph, while this branch adds the paragraph below that block and
the Composing Reducers section. Both are present on the rebased tree; I checked
rather than assuming, since I had predicted a conflict.

**A — the held-Enter rationale** was already fixed, in 9b770e8, before the pause.

**B — `open_details` replacing an occupant in silence.** Fixed. That present is a
replacement, a replacement is a removal, and the occupant it removes can hold
edits nobody saved — which is why `reload_task` three functions above says what
it discarded. The status distinguishes the two cases now, a row covers both, and
reporting the open unconditionally fails it.

**C — the test module's doc.** It said its tests are driven by `TestDriver`, of a
module where most of them are not. It describes both kinds now, by what a row
builds rather than by how many there are. `docs/composition.md` named four rows
by name for the same claim; it does the same, since a name rots the way a number
does.

**What #362 changed about the plan I recorded here.** I said two decisions would
be applied on this branch: the comparison between the two files lives only in
`dashboard_composed.rs`'s module doc, and the composed example states its own
behavioural rules and applies each everywhere.

The first no longer has a rule behind it. #362 tried to write that rule and
withdrew it: it forbade an example describing another example's behavior and
then required exactly that for a pair that did not exist yet. So the single-home
arrangement stays as a judgment about this pair rather than as a rule being
applied, and I am not going to argue it from a convention that is not there.

The second is not a rule either, but it is what B was an instance of, and it
holds: an operation that could not do what was asked, or that discarded
something, says so.

`cargo test --lib --tests --examples`, `just test-doc`, `cargo clippy
--all-targets`, `cargo fmt --check`, `typos` and `RUSTDOCFLAGS=-D warnings cargo
doc` all pass on the rebased tree. Composed is 14 tests, `dashboard` 11.

akiomik:
All three in b4cbecc.

**A — and this one is mine twice.** You reported it, I said last round it had
been fixed in 9b770e8, and it had not: the sentence was byte-identical to the
one you quoted. 9b770e8 fixed the *second* half of that comment and I read the
finding as being about the half I had touched. The Enter example is gone; `n`
and `d` carry the point, and the timing claim that contradicted the paragraph
under it goes with it.

**B, C — the guard answered two of six.** Verified: `KeyModifiers` has `SHIFT`,
`CONTROL`, `ALT`, `SUPER`, `HYPER`, `META`, and `intersects(CONTROL | ALT)` let
four through to the bare key's binding. Shift+Down is reachable today with no
protocol setup, so the task selection moved; Shift+Enter would have opened the
pane, replacing an occupant, which is a removal.

Fixing it turned up the reason the wide form is also wrong, which I had not
anticipated: letting `SHIFT` through generally reaches the row bindings, where
`d` deletes. My first attempt did exactly that and the widened row caught it —
`Char('d')` with `SHIFT` reached `DeleteTask`. The exception is narrower than
"a shifted character": it is a shifted character on its way into the notes
editor, which is how an uppercase letter arrives and the only modified key
either application takes.

Both rows now run all five constructible modifiers rather than the two their
names claimed, and both cover the exception. Narrowing the guard back to
`CONTROL | ALT` fails them in both files.

Full verification on the pushed tree: `cargo test --lib --tests --examples`,
`just test-doc`, `cargo clippy --all-targets`, `cargo fmt --check`, `typos`,
`RUSTDOCFLAGS=-D warnings cargo doc`. Composed is 14 tests, `dashboard` 11.

akiomik:
All four in db3cbef.

**A–C.** Fixed all three rather than weighing the reachability, because your
last argument settles it independently of frequency: this file names the hazard
in `Message::AddTask`'s own doc — a key handler reads the selection off a state
the messages already in flight have not been applied to — so a branch that goes
quiet when that hazard fires contradicts the file. It also contradicts what the
file spends effort on elsewhere: `NoTaskSelected`, `select` setting its status
above the guard, `CloseDetails` telling its two cases apart.

They say the row is gone. A row covers all three and fails if any goes back to
silence.

For the record on the split verdict: the round that called these unreachable was
reasoning from `state.selected` alone, which is right about `selected` and wrong
about the message in flight. `d` then `r` are both decoded against
`selected == Some(1)`, and only the first is applied before the second arrives.

**D.** The gap is exactly where you place it. `Segment::Details` is the third
variant and `beta` is `TaskId(2)`, so that pair's segments erase to the same
value and nothing but the type carried alongside separates them — and it was the
one pair the row did not compare, while claiming no two collide. A regression
dropping the type from the structural key would have kept it green. Two
assertions added, with the reason in a comment so the next person does not read
them as redundant with the alpha ones.

On the note about `every_removal_tears_its_instance_down`: agreed, and your
reading matches mine — the filter removes the two `added:` lines, which are the
first the buffer sheds, so a ninth entry surfaces as an extra element rather
than a missing head. No change.

Full verification on the pushed tree. Composed is 15 tests, `dashboard` 11.

akiomik:
All three in 9964961.

**A — the fourth of four, and mine.** Last round I fixed three early returns and
left this one, in the same commit whose subject was that this file must not go
quiet there. Its path is the one the file documents: two `d` presses decoded
against one state name the same row, the first removes it, the second finds
nothing. It says so, and the row that holds the contract runs all four
operations rather than three.

**B — a status built and never read.** Correct: the `Save` arm set the same
literal itself while the `Reset` arm beside it read the outcome, so editing the
panel's wording changed nothing. The arm reads it. There was no coverage of that
wiring either — changing the outcome string failed nothing — so a row asserts it
now, and hardcoding a different literal in the arm fails that row.

**C — an unreachable branch holding a user-facing string.** Taken as you frame
it. Both paths that remove a row close a pane opened on it first, and a reload
puts a row back under the same key, so nothing reaches it through the keys. I
left the branch and the comment and dropped the status: a line nobody can reach
is a line nobody can check, which is the argument against having put one there.

I did not use `unreachable!`. If my reasoning about the two paths is wrong, a
panic in a runnable example is a worse outcome than a no-op, and the comment
records the claim where the next reader can test it.

**On the dispatch-time status.** Agreed, and your ordering is the reason: a
child-addressed message's status is written at dispatch, and the removal that
would falsify it lands afterwards and overwrites. I had not traced it that far
either — I wrote `child_status` and reasoned about which arms are unreachable,
not about which write wins. Worth recording that this is the third pass over the
same point.

Full verification on the pushed tree. Composed is 15 tests, `dashboard` 12.

akiomik:
A and B in 0192c12. C is a decision and I have taken it: no entry.

**A — one condition, two answers.** Fixed, and the history is the finding's
point: the status went above `select`'s empty guard in round 8 because at the
time that branch said nothing at all, and `NoTaskSelected` arrived in round 12
and took over the case without the earlier fix following it. Both paths go
through one function now, so the line has one place and cannot come to disagree
with itself, and `select` sets its own status after the guard again — where it
means a move that happened.

**B — the untested half of a pair.** Right, and the commit that gave
`dashboard.rs` the handle had "report a terminal error the same way in both
dashboards" as its subject while covering only one of them. It has a row;
dropping the record fails it.

**C — no changelog entry.** `docs/releasing.md` frames `Added` as public API a
user can reach, this branch adds none, and the file has no docs-only entry in
its history — the two examples it mentions are sub-items under feature entries,
as you say. Your vote and mine agree, and I have recorded the disposition in the
commit message rather than only here, so the reasoning is in the tree.

If you would rather it were listed anyway, the argument on the other side is
real and one line reverses this.

Full verification on the pushed tree. Composed is 15 tests, `dashboard` 13.

On the flake note: agreed, and the mechanism you traced is the one I relied on
when I chose a 900 ms stub without saying why it was safe. `settle` issues no
grant and asserts none is outstanding, and `deliver` grants a named run, so the
sync's send waits at the gate however much wall-clock passes. That is now
checked rather than assumed, which it was when I wrote it.

akiomik:
Both in 9242644, and I changed how rather than only what.

## Why this round is different

Five rounds on this branch have found instances of one rule — the footer says
what happened — and I have fixed them one at a time. Silence on an empty
selection, then on four keys needing a row, then on a save with no task, then a
reset, then a replacement, then three vanished rows, then the fourth, then two
answers for one condition. Round 3's finding was created by round 2's fix
leaving one of four; this round's A was created by round 4's fix making a guard
unreachable and two comments false.

So I enumerated instead: every status write and every early return in both
files, checked against what its path can actually have done.

| | count | untrue |
| --- | --- | --- |
| `dashboard_composed.rs` status writes | 12 | 1 |
| `dashboard_composed.rs` early returns | 8 | 0 |
| `dashboard.rs` status writes | 20 | 2 |
| positional comments | 2 | 2 |

The three untrue ones are the same one: a selection clamped at an end stayed
where it was and said it had moved. Both files say "Nothing further that way",
and a row in each fails if either goes back to reporting the move. Everything
else in the enumeration is either true, or a child's, which cannot reach the
status.

## The findings

**A.** The guard is gone — the root answers that case, which is what the last
round moved there — and both comments say what is true rather than where code
sits. Describing code by position is why they inverted when it moved; that is
the correction worth keeping from this one.

**B.** Fixed in both files, since `dashboard.rs` had it twice. Parity is kept by
fixing rather than by declaring, because "reports a move that did not happen" is
not a difference worth preserving.

**The dispatch-time status, fourth time.** Added to `child_status`'s doc as you
suggest, in the form your ordering gives: a line set at dispatch cannot be left
standing as a lie, because a root message that removes what the child was
addressed to writes its own status when it lands, and a landing is always after
the dispatch that queued it.

## What I expect

The generator for this class is the enumeration, not a rule, and it is done. A
finding in the next round would be a different class, which is information the
last five rounds did not produce. If it is this class again, my enumeration was
incomplete and that is what I would go back to rather than the instance.

Full verification on the pushed tree. Composed is 15 tests, `dashboard` 14.

akiomik:
All four in 1530ede.

## The enumeration was incomplete, and here is how

I said last round that another finding of this class would mean my enumeration
was wrong. It was, and the defect is precise: **I counted the places a status is
written, not the conditions each value claims.** `child_status` is one write
site returning four strings, and it entered the table as one row with a tick.

Redone by string:

| string | verdict |
| --- | --- |
| "Selected another section" | false — the list clamps, and Up on a fresh run is the repro |
| "Cleared the activity log" | false — an empty log clears nothing |
| "Toggled the task" | true, but phrased as an outcome |
| "Editing the notes" | true |
| `dashboard.rs` Up/Down on an empty list | false — an empty list is not an end |

**A.** The three that claimed outcomes name what was asked for now, which is
what `child_status`'s doc already said they did — the strings had not caught up
with it. The doc says why an outcome cannot go there: the line is written before
the child runs.

**B.** `dashboard.rs` answered an empty list as an end while `Toggle` and
`Delete` answered it as "No task selected" — one file, two answers for one
state, which is the finding this branch closed on the other side two rounds ago.
Answered before the clamp.

**C.** Verified the mechanism before rewording: `redraw_pending` is set at
bootstrap and from a command's plan, and a finalizer produces neither. The doc
says the pane carries the line at the next draw rather than showing it as the
hook runs.

**D.** On the bullet that already says the tasks arrive as messages, since that
is what causes it: adding a task selects it, so the binary opens on the last of
three.

## Two process notes

My first mutation check on `dashboard.rs` did not apply — rustfmt had collapsed
the arms I was matching against, so the edit silently did nothing and the test
"passed". I re-ran it against the formatted text and it fails as it should. A
mutation that does not apply reads exactly like a test that does not catch.

And my verification script had no `set -e`, so `clippy` failed into the middle of
the log while `test exit=0` sat at the end. I committed on that and would have
reported it green. It fails fast now, and the lint it was hiding —
`match_same_arms` on the two empty-list arms — is fixed in the same commit.

Composed is 15 tests, `dashboard` 14. fmt, typos, clippy, test (20 binaries),
test-doc and doc all pass on the pushed tree.

akiomik:
Both in bbf2b77.

**A.** The unconditional resync predates this branch; the status line that
contradicts it does not, which is what makes it this branch's to fix. It rereads
where the selected task can have changed — the index moved, or a row was added
or removed under it. The index alone is not enough: deleting a row leaves it
pointing at a different task. A toggle changes neither and a clamped move
changes nothing at all.

A row covers it, and reverting to the unconditional reread fails it. The
difference list said the reread follows the selection, which is narrower than
what it does; it names the three.

**B — and this one is a process failure, not a judgment.** The edit was in last
round's script as a string replacement with no assertion behind it. It matched
nothing, changed nothing, and I reported it as done. Every replacement in that
script had an `assert` except this one.

That is the second round running where an operation silently did nothing and I
read the silence as success — last round it was a mutation that did not apply
because rustfmt had reshaped the target. Same shape: an edit that no-ops looks
exactly like an edit that worked.

Every replacement in this round's script asserts, and I checked the line is in
the file by grep rather than by having run the script.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree. Composed is 15 tests, `dashboard` 15.

akiomik:
Both in ba5ac1e.

**A.** The guard asked what kind of message this was, and that is the wrong
question — the property it needs is whether anything changed. A `Delete` on an
empty list removes nothing, moves nothing, and returns "No task selected", and
the message kind cannot tell it from a `Delete` that deleted. It asks the state
now: the selection moved, or the list grew or shrank. That covers an add, a
delete that deleted, and every move, and leaves out a delete that deleted
nothing and a toggle.

A row covers it, and going back to the message kind fails it with the typed
character gone.

Thank you for the correction on your own check — I made the same one. I traced
`Delete` clamping `selected` to 0 when the last row goes, concluded the index
comparison alone was insufficient, and reached for the message kind instead of
for the count. Testing a proxy for the property is what let both of us miss the
same case.

**B.** `HYPER` is in both arrays, so the rows run the six the comment beside
them counts.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree. Composed is 15 tests, `dashboard` 15.

akiomik:
All three in 0c74b09, plus the two sweeps I said I would run.

## The findings

**A.** `contains(CONTROL)` against a comment saying anything else carrying a
modifier is unbound. It is `== CONTROL` now. The first attempt at covering it
asserted the focus had not moved — which a spurious quit does not move either —
so the composed row adds the pairs to its chord list and `dashboard.rs` gets a
row where `finish` is the assertion, a quit request being output and `finish`
failing on output left unaccounted for. Both fail under `contains`.

**B.** Both absence claims were read before anything could have written: a
hook is a spawned run and does not write in the pass that tore its scope down.

**C.** Removed. Only `handle_key_event` reads the focus and that row sends
messages directly.

## Sweep 1 — assertions of absence

Every one in the composed rows. Two read the activity log without spending
turns, as you found.

- The one after the two adds now spends turns, and it fails when the key
  allocation collides so the second add replaces the first. It is sensitive.
- The one after the two opens is **deleted**. It cannot fail under any reachable
  change: a hook is armed by the child's own `Open`, which is not delivered at
  that point, and one armed at the root would anchor where the slot's teardown
  cannot reach it — this file's own rule. The reasoning is kept as prose; the
  assertion was not carrying it.

The `boot.terminated.is_none()` rows are reads of a returned value, not of a
log, so nothing has to have happened first.

## Sweep 2 — comments stating a guard, against the guard

A was the instance. The rest agree with their code: the modifier guard and its
`shifted_text` exception, the release filter keeping repeats, `editing_notes`
asking the slot, the `save_notes` branch documented as unreachable.

## Two process failures this round

Two mutations did not apply and read as passes. The first was rustfmt reshaping
the text I matched against — the same failure as two rounds ago. The second was
mine in a different way: I built a `Command::on_teardown` at the root and never
returned it, so nothing was armed and the mutation tested nothing. I now check
the edit landed before reading the result, which is what caught both.

## The exit condition I stated

Both sweeps are done. If the next round finds this class again — a comment
disagreeing with its guard, or an assertion that cannot fail — I will stop
fixing instance by instance, file the remainder, and propose merging rather
than spending another round.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree. Composed is 15 tests, `dashboard` 16.

akiomik:
The finding is fixed in 922c360, and the observation is disposed of as #364.

**The finding.** You are right about the mechanism, and I checked it rather than
taking it: `finish_fails_on_an_unreceived_ready_message` in `src/testing.rs` is
what that row lands on, and `check_exhaustive`'s quit arm is never reached
because the quit is never applied — the key handler returns
`Command::message(Message::Quit)` and the row never receives it. The doc says
that now. The row was sound and its stated reason was not, in a file read for
its reasons.

**The observation — not added to the guide, filed as #364.** Three reasons:

- Nothing currently written is false. `docs/composition.md` says what a boundary
  does with a *removed* instance, and all of that holds. What is missing is a
  statement about the *successor*.
- That statement's home is the combinator's own rustdoc — `ReducerExt::presented`
  and `for_each` — not a guide about when to compose. A reader hits this while
  writing a replacement, not while deciding whether to have boundaries.
- Every explanatory half-sentence added to this branch and to #362 has produced
  the next round's finding. This one would be a claim about `reconcile`'s
  ordering, which is exactly the kind that has not survived.

I verified the ordering in `src/kernel/pass.rs` before writing the issue rather
than repeating your trace, and the issue records the mechanism, the condition
that makes it visible (a replacement returning `Command::none()`; both sites
here return a command, so it never shows), and why it belongs in the library's
documentation rather than in the example.

Both items are now disposed of — one fixed, one filed.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree.

akiomik:
Fixed in 4936991, with the sweep.

**The finding.** Took (b). The silence itself is ordinary — most keys are unbound
in most focuses, and nothing here reports every unbound key — so the defect is
the hint printed in that pane naming one of them. It names Tab to the task list
now, which is reachable from where it is printed, and does not count how many
Tabs.

**Sweep of the other hints.** The pane titles list their own pane's keys, which
work when that pane has focus; both editor hints and both headers name keys
bound where they are shown. This was the only one where a printed hint pointed
at a key that is inert and silent from that pane.

**No row covers it, and I would rather say why than add a weak one.** The hint
is built inside `render_details`, so asserting it compares a literal against
itself. The check that would mean something — every key a hint names is bound
from where it prints — needs the hint parsed, and that is more machinery than
this example is worth. The reason is in the commit message rather than only
here.

**On the brittleness note:** your reading matches the code and the test's own
comment. The assertion is on `written`, which drops the two `added:` lines, and
those are the first the buffer sheds — so there are two entries of headroom, not
zero. Nothing to do.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree.

akiomik:
All three in 898e742.

**A.** Made the code symmetric rather than annotating the asymmetry: forward
from nowhere is the first row, backward is the last. It is unreachable today, as
you say, so this changes no behaviour — and a comment saying "unreachable" would
be one more claim to keep true, which is the thing that has cost this branch the
most. A row pins the arm instead, and dropping the direction fails it.

**B.** Both files, since either alone would be the asymmetry. Thank you for
checking the composed side — the agent named only `dashboard.rs` and the
duplicate was in both. The copy on the guard stays; the function doc says only
what the guard cannot, that raw mode delivers no SIGINT so Ctrl+C arrives as an
ordinary key event.

**The wall-clock question, third round.** Added next to the assertion, in your
terms: the requests sleep in real time, `deliver` grants a named run and
`settle` grants none, so a request that finishes waits at the gate and its line
is never written. Same move that stopped `child_status` recurring — the answer
belongs where the question gets asked.

One process note. An `assert` on a string replacement caught a mismatched anchor
this round, so the edit failed loudly instead of silently doing nothing. That is
the failure I reported two rounds ago as having shipped a "fixed" item that was
never applied; the guard is now doing its job.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree. Composed is 16 tests, `dashboard` 16.

akiomik:
Both in 43575e5, and the convergence estimate you asked for.

**A.** Correct, and the README contradicted itself two lines apart. What
separates those two from `scope` is that their child can leave, not that they
hold a collection — `presented` holds a `Slot`, which the crate defines as at
most one child state. It names the property.

**B.** The panel is emptied either way, so that arm is not the `Save` arm's
no-op and the comment above it said it was. The footer says what became of the
notes; the row asserts the panel is empty as well as what the footer reads.

## Convergence

Findings per round since the rebase:

```
3, 4, 3, 3, 2, 4, 2, 2, 3, 1, 1, 2, 2
```

| span | mean |
| --- | --- |
| rounds 1–7 | 3.0 |
| rounds 8–13 | 1.83 |
| last four | 1.5 |

The trend is real, and the kind has changed with it: no correctness defect for
more than ten rounds, and the last six findings are all precision of statements
— a test doc's mechanism, a hint naming an inert key, an unreachable arm's
asymmetry, one rule written twice, the README's "collection", this round's
footer. The behaviour has been right for a long time; what is left is the
surface of English, and this artifact carries roughly 14 status strings, 8
hints, 32 test docs and ~150 comments.

My estimate: reaching a zero round in one or two more is plausible, in three to
five likely, at roughly 70k tokens each. The uncertainty is not the review's
yield but my own fix-induced rate — 6 or 7 of the 26 post-rebase findings were
created by my previous fix, though only one of the last four was.

**The bar I will hold myself to:** if the next two rounds do not reach zero or
one, I will stop fixing instance by instance, file what remains, and propose
merging. Past that point the decay does not pay for the rounds.

fmt, typos, clippy, test (20 binaries), test-doc and doc all pass on the pushed
tree.

One process note: this round's mutation check did not apply on the first attempt
— rustfmt had wrapped the line I matched — and the `assert` caught it, so the
"ok" that followed was meaningless rather than reassuring. Re-run against the
formatted text, it fails as it should.

akiomik:
Both in c11db27, along with a third site the same rule reaches.

**A.** Correct, and the rule was already written down one function over —
`open_details`'s comment says a removal that may take edits with it names what
it discarded. Esc was the one removal not following it, and the pane's own hint
steers readers into it.

I did not take the wording as recommended, though. `save_notes` is the only
thing that moves the buffer onto the task, so right after a save the buffer and
the task agree and nothing is outstanding — an unconditional "discarding its
unsaved notes" would tell a reader who just pressed Enter that their save was
lost. That is the same class as the footer this round's B and last round's Esc
were about: a line claiming something that did not happen. So the discard is
reported only when the two actually differ:

```rust
let unsaved = state.details.get().is_some_and(|open| {
    state.tasks.get(&open.task).is_some_and(|task| task.notes != open.notes)
});
state.status = match (close_details(state), unsaved) {
    (true, true) => "Closed the details pane, discarding notes nobody saved",
    (true, false) => "Closed the details pane",
    (false, _) => "Left the details pane",
};
```

**B.** Your reading, not the count. The pane is drawn whenever the slot is
empty — at every focus, including from the task list, where no Tab is needed.
`"No details open. Press Enter on a task in the task list."` names where the key
has to land, which holds at every focus and counts nothing.

**C, not reported.** Applying A meant `open_details`'s replacement was left as
the site that states the rule without following it: replacing an occupant
discards its edits exactly as Esc does, and the string said only "replacing the
one that was open". Same test, same tail, so the two read alike:

- `(true, true)` — `"Replaced the open details pane, discarding notes nobody saved"`
- `(true, false)` — `"Opened the details pane, replacing the one that was open"`

I swept all 16 status writes for the rest of the class. Two removals do not
qualify: `delete_task` takes the row itself, so the notes had nowhere to
survive, and `reload_task`'s "discarding its local state" is the wider claim
that already contains the pane's buffer. `dashboard.rs` has no member — its
panel is not a `Slot`, so Esc rereads it in place and a selection move refills
it, and in both the discard is on screen. The composed Esc removes the pane
itself, which is what makes it unreadable there.

Rows pin all three outcomes at both sites, and both new assertions fail under
either one-sided mutation (always claim the discard / never claim it).

fmt, typos, clippy (build features and `loom-core`), test (20 binaries),
test-doc and doc all pass on c11db27.

## On the stopping rule

Withdrawn, and you are right that I had no strong reason. Filing does not
remove the work or its cost; it adds the cost of finding it again, and I pay
both. Worse, the call was not mine to make — what I owe you is the estimate,
and I turned it into a threshold.

Filing is right when the decision lives outside the PR: a design change, another
contract, a pattern across files that a line here cannot fix. #363 and #364 met
that. These did not — they are strings this PR wrote, and everything needed to
settle them is in this PR.

The stopping condition stays what it was: zero findings.

## Verification, self-reported

Last round I rebuilt the verification script from memory and typed feature names
the crate does not have. `cargo test` and `cargo doc` both exited non-zero, and
the script printed "test ok: 0 binaries" and "doc ok" and exited 0 — the
pipelines ended in `grep -c` and `tail`, whose success is the status `set -e`
reads. Nothing was actually verified by those two lines.

The features now come out of the justfile rather than my memory, each step's
status is read directly instead of through a pipe, and the run above is the
first one on this branch I can say that about. `just test` is not what I ran:
it is `--all-targets`, which pulls in the bench and load harness, so this is the
CI shape (`--lib --bins --tests --examples`) plus the loom lint pass.

akiomik:
39de022. A is declined with the reason recorded and pinned, B needs nothing,
and the eviction question is now read off the log.

## A — declined, and my previous ground was wrong

You are right that the ground I gave last round does not hold: the task list
renders `title`, `done`, `ticks` and the syncing marker, not `notes`
(`render_tasks`), so nothing about the reload is legible on screen. That
argument is withdrawn.

The line that does hold is whether a save would have helped. `save_notes` is
the only thing that copies the buffer onto the task, so:

| removal | would Enter first have kept it | reports the buffer |
| --- | --- | --- |
| `CloseDetails` | yes — the task keeps its notes | yes |
| `open_details` over an occupant | yes — same | yes |
| `reload_task` | no — the row's notes become the server's copy | no |
| `delete_task` | no — the row is gone | no |

Naming the buffer in the bottom two would offer a remedy that does not exist:
a reader who had pressed Enter before `r` or `d` loses the notes just the same.
What those two discard is the row, which is what their status says.

So the fix is not the wording, it is that the line was in neither the code nor
the prose. The comparison is extracted to `unsaved_notes`, whose doc carries
which removals ask it and why the other two do not — both callers now share it
rather than repeating the closure, which is your suggestion minus the branching.
Each declining site says at the status it sets why it declines.

Prose alone would let this drift back, so a row pins both declining strings and
fails if either moves toward `CloseDetails`'s wording. Confirmed by mutation.

## B — declined, and the disposition is already in the repo

`Cargo.toml:31-33` states the policy the exclusion follows: the entry is named
file by file rather than by glob because a directory pattern "would also ship
`docs/migrations/README.md` — a maintainer policy index, not crate content."
`docs/composition.md` is the same kind of file, and `README.md` and
`src/lib.rs` reference `docs/` the same way. Nothing to add — a second record
of a policy already written is a second thing to keep true.

## `ACTIVITY_LIMIT` — no change, and it should stop coming back

Your arithmetic is right and matches the row's own comment. Four derivations of
the same fact is a cost the row was imposing, so it no longer asks a reader to
do it: before the comparison it asserts the log still holds its oldest line.

```rust
assert_eq!(
    activity.entries().first().map(String::as_str),
    Some("added: #1 alpha"),
    "the log still holds its oldest line, so nothing below is missing for want of room"
);
```

A line added anywhere above pushes that one out first, so this fires first and
names the buffer's capacity — instead of leaving the comparison below to fail
about its own contents, which is the failure mode the question kept being about.
`ACTIVITY_LIMIT` is untouched, so the pane still shows what it showed.

fmt, typos, clippy (build features and `loom-core`), test (20 binaries, 18 in
`dashboard_composed`), test-doc and doc all pass on 39de022.

akiomik:
616ad74. `/// Presents the details pane for a row.` is back above `open_details`,
and `unsaved_notes` opens with its own summary.

Mine, and the cause is exact: extracting the helper, I anchored the insertion on
the `fn open_details(` line, which put the new function between an existing doc
comment and the item it documented. Nothing about the result is malformed — two
summary lines in one block is valid rustdoc — so, as you say, examples being
outside `cargo doc` is not what hid it. Nothing would have read it.

I swept both example files for the same shape rather than fixing only the one
you named: every doc block whose first line completes a sentence and is followed
by more prose, which is what a merged block looks like. Zero, in either file, so
this was the only one.

There is no check to add here that I can see — rustdoc does not warn on a block
with two summaries, and `--examples` would not either. The guard belongs to how
I make the edit: anchor an insertion on the blank line before an item rather than
on the item, and run that scan afterwards. That is now part of my own routine
rather than something the repo has to carry.

fmt, typos, clippy (build features and `loom-core`), test (20 binaries),
test-doc and doc all pass on 616ad74.

akiomik:
a088f58, in both files.

The finding holds, and your wider reading of it is the right one — all three
clauses, not just `Enter`. I checked the decode rather than taking it: the
`Enter` arm for `Focus::Tasks` is at `:1264` and the editor's at `:1272`, so
with a pane open behind the task list the first one wins and `open_details`
runs over its own occupant. The replacement is a removal, so the key the line
offered as a save is the one that discards the buffer.

`dashboard.rs` had the same line over the same guards, and `render` already
takes `focused`, so it is fixed too rather than left as the asymmetric half of
a pair the PR is about. There `Enter` away from the panel is unanswered rather
than destructive, which makes it the milder case and not a different one.

Both now read `details_hint`, which names the three keys only at the focus that
has them and otherwise names where they have to land — the wording the empty
branch already carries for the inert version of this.

Prose would let it drift back, so each file pins both halves. The composed row
reads the decode instead of arguing it:

```rust
state.focus = Focus::Tasks;
assert!(matches!(bare(&state, KeyCode::Enter), Some(Message::OpenDetails(TaskId(1)))));
// ...then runs it and finds the discard in the status.
```

and `dashboard.rs`'s ends on `finish()` with no `receive_matching`, since away
from the panel the key produces `Command::none()` and a save would be a message.
Both fail under a mutation that returns the focused wording unconditionally.

I also re-ran the merged-doc-block scan over both files after these insertions,
since last round's defect came from exactly this kind of edit: zero.

fmt, typos, clippy (build features and `loom-core`), test (20 binaries — 17 in
`dashboard`, 19 in `dashboard_composed`), test-doc and doc all pass on a088f58.

akiomik:
6ab56b2. A is fixed in both files; B is declined, and the record it needs is
already in the repo.

## A

Holds. This PR made `q` text in the editor and added the `Ctrl+C` handler, and
never showed the second anywhere — so the one focus where `q` does not work is
the one focus with no quit named.

One correction to the framing: `Ctrl+C` is not the only way out of that focus.
`Tab` leaves the editor and `q` works again, and the header says so. It is the
only one that does not ask the reader to leave first, and under raw mode the
application is the only thing that can answer it, which is what makes it worth
naming rather than merely available.

Each header is now a function of the focus it is drawn for. The rows press the
key each branch offers instead of comparing strings: `q` outside the editor,
the chord inside it, and in the composed file the `q` that decodes to
`DetailsMessage::Input('q')` there — which is the reason the branch stops
naming it. Both fail under a mutation that drops the chord from the string.

## B — declined

Your correction to the framing is right and I checked it: the same reread runs
for `delete_selected` and for Up/Down, so `n` is not the last silent discard,
and `Reset` names its discard because there the discard is what was asked for.

The record already exists. `dashboard_composed.rs`'s module doc says the panel
is reread "whenever the selected task can have changed — a move, an add, a
delete — so any of those throws an unsaved edit away; here the occupant is
fixed to the task it was opened for, because a `Slot` holds an instance rather
than a view of whatever is selected." All three paths are named, and the
property is what the bullet exists to contrast against `Slot`. Making the three
announce a discard would blunt exactly that contrast, and nothing needs adding —
a second statement of a documented property is a second thing to keep true.

fmt, typos, clippy (build features and `loom-core`), test (20 binaries — 18 in
`dashboard`, 20 in `dashboard_composed`), test-doc and doc all pass on 6ab56b2.
The merged-doc-block scan is clean in both files after these insertions.

PR #362, opened 2026-09-05
docs: prose describing code is read again when that code changes

akiomik:
Blocks #360, which pauses until this is settled.

## What this adds

One sentence to the README's Contributing paragraph:

> An edit is not finished until the changed file's own documentation has been
> read again — the module doc, every doc comment under it, the comments in the
> body — and any disagreement with the code fixed on one side or the other,
> explicitly.

And one phrase in the Testing section, which is the whole of the rest of the
diff: "Worked, runnable tests ship with **the command examples**" named a
category one of the three listed is not. It says "these examples" now.
`docs/testing.md` is untouched.

## Why it is only that

#360 revised one example across many rounds. Sorting those findings, the prose
that failed was overwhelmingly prose whose truth lived somewhere other than the
lines it annotated — and the ones a *placement* rule could have prevented were a
minority. The rest were sentences describing code their author had changed
without reading them again.

Earlier revisions of this branch tried to write the placement rule anyway: a
table assigning each kind of claim an owner, then four rows, then two, then two
rules. Every version either denied a convention the tree uses or permitted what
it was written to forbid.

What survived is one obligation. The rounds since found a defect in every
sentence written *in support of* it — the scope, the cross-file method, the
pointer that replaced the method, and twice the account of what goes unchecked,
whose real boundary is a three-row table. Each fix was an explanation, and every
explanation is a claim. The support is gone; the instruction has drawn no
finding in fifteen rounds.

It sits in Contributing rather than in `docs/testing.md` because that is where
the reader it binds already is, and because a paragraph there already gives
instructions rather than only directions.

## What it deliberately does not cover

Prose in one file that pins code in another. That is #363.

## Verification

`typos` passes. Nothing here is compiled into the crate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,8 +289,8 @@ cargo test --example command_cancellation
 cargo test --example dashboard
 ```
 
-Repository-wide test conventions, and one about keeping prose true when the
-code under it changes, live in [docs/testing.md](docs/testing.md).
+Repository-wide conventions for tests, and for the prose around the code they
+cover, live in [docs/testing.md](docs/testing.md).
 
 ## Optional Features
 
@@ -351,8 +351,8 @@ Contributions are welcome! Please feel free to submit issues or pull requests.
 Design contracts and invariants live in [docs/rfcs](docs/rfcs/README.md).
 If you are writing or amending an RFC, run the
 [pre-review checklist](docs/rfcs/pre-review-checklist.md) before
-requesting review. Testing conventions are documented in
-[docs/testing.md](docs/testing.md), and the release procedure in
+requesting review. Conventions for tests, and for the prose around the code they cover, are
+documented in [docs/testing.md](docs/testing.md), and the release procedure in
 [docs/releasing.md](docs/releasing.md).
 
 ---
--- a/README.md
+++ b/README.md
@@ -289,8 +289,9 @@ cargo test --example command_cancellation
 cargo test --example dashboard
 ```
 
-Repository-wide conventions for tests, and for the prose that describes the
-code, live in [docs/testing.md](docs/testing.md).
+Repository-wide test conventions live in [docs/testing.md](docs/testing.md),
+along with the rule that prose describing code is read again when that code
+changes.
 
 ## Optional Features
 
@@ -351,9 +352,10 @@ Contributions are welcome! Please feel free to submit issues or pull requests.
 Design contracts and invariants live in [docs/rfcs](docs/rfcs/README.md).
 If you are writing or amending an RFC, run the
 [pre-review checklist](docs/rfcs/pre-review-checklist.md) before
-requesting review. Conventions for tests, and for the prose that describes
-the code, are documented in [docs/testing.md](docs/testing.md), and the
-release procedure in [docs/releasing.md](docs/releasing.md).
+requesting review. Testing conventions, and the rule that prose describing
+code is read again when that code changes, are documented in
+[docs/testing.md](docs/testing.md), and the release procedure in
+[docs/releasing.md](docs/releasing.md).
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -280,13 +280,11 @@ store.finish();
 its own paused time context) and does not execute subscription sources — it
 observes only the *declared* set via `subscription_ids`. See the
 [`tears::testing`](https://docs.rs/tears/latest/tears/testing/) module docs for
-the full contract, including deterministic time without `TestStore`. Worked,
-runnable tests ship with the command examples:
+the full contract, including deterministic time without `TestStore`. Several
+examples carry worked, runnable tests of their own:
 
 ```bash
-cargo test --example command_timeout_retry
-cargo test --example command_cancellation
-cargo test --example dashboard
+cargo test --example <name>
 ```
 
 Repository-wide test conventions live in [docs/testing.md](docs/testing.md).
--- a/README.md
+++ b/README.md
@@ -289,8 +289,8 @@ cargo test --example command_cancellation
 cargo test --example dashboard
 ```
 
-Repository-wide conventions for tests, and for the prose around the code they
-cover, live in [docs/testing.md](docs/testing.md).
+Repository-wide conventions for tests, and for the prose that describes the
+code, live in [docs/testing.md](docs/testing.md).
 
 ## Optional Features
 
@@ -351,9 +351,9 @@ Contributions are welcome! Please feel free to submit issues or pull requests.
 Design contracts and invariants live in [docs/rfcs](docs/rfcs/README.md).
 If you are writing or amending an RFC, run the
 [pre-review checklist](docs/rfcs/pre-review-checklist.md) before
-requesting review. Conventions for tests, and for the prose around the code they cover, are
-documented in [docs/testing.md](docs/testing.md), and the release procedure in
-[docs/releasing.md](docs/releasing.md).
+requesting review. Conventions for tests, and for the prose that describes
+the code, are documented in [docs/testing.md](docs/testing.md), and the
+release procedure in [docs/releasing.md](docs/releasing.md).
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -355,8 +355,9 @@ requesting review. Testing conventions are documented in
 [docs/releasing.md](docs/releasing.md).
 
 An edit is not finished until the changed file's own documentation has been
-read again and corrected where the change made it false: the module doc, the
-doc on the item, the comment over the line.
+read again — the module doc, every doc comment under it, the comments in the
+body — and any disagreement with the code fixed on one side or the other,
+explicitly.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -289,9 +289,7 @@ cargo test --example command_cancellation
 cargo test --example dashboard
 ```
 
-Repository-wide test conventions live in [docs/testing.md](docs/testing.md),
-along with the rule that prose describing code is read again when that code
-changes.
+Repository-wide test conventions live in [docs/testing.md](docs/testing.md).
 
 ## Optional Features
 
@@ -352,10 +350,13 @@ Contributions are welcome! Please feel free to submit issues or pull requests.
 Design contracts and invariants live in [docs/rfcs](docs/rfcs/README.md).
 If you are writing or amending an RFC, run the
 [pre-review checklist](docs/rfcs/pre-review-checklist.md) before
-requesting review. Testing conventions, and the rule that prose describing
-code is read again when that code changes, are documented in
+requesting review. Testing conventions are documented in
 [docs/testing.md](docs/testing.md), and the release procedure in
 [docs/releasing.md](docs/releasing.md).
+
+An edit is not finished until the file's own documentation has been read
+again and corrected where the change made it false: the module header, the
+doc on the item, the comment over the line.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -280,11 +280,13 @@ store.finish();
 its own paused time context) and does not execute subscription sources — it
 observes only the *declared* set via `subscription_ids`. See the
 [`tears::testing`](https://docs.rs/tears/latest/tears/testing/) module docs for
-the full contract, including deterministic time without `TestStore`. Several
-examples carry worked, runnable tests of their own:
+the full contract, including deterministic time without `TestStore`. Worked,
+runnable tests ship with three of the examples:
 
 ```bash
-cargo test --example <name>
+cargo test --example command_timeout_retry
+cargo test --example command_cancellation
+cargo test --example dashboard
 ```
 
 Repository-wide test conventions live in [docs/testing.md](docs/testing.md).
--- a/README.md
+++ b/README.md
@@ -289,7 +289,8 @@ cargo test --example command_cancellation
 cargo test --example dashboard
 ```
 
-Repository-wide test conventions live in [docs/testing.md](docs/testing.md).
+Repository-wide test conventions, and one about keeping prose true when the
+code under it changes, live in [docs/testing.md](docs/testing.md).
 
 ## Optional Features
 
--- a/README.md
+++ b/README.md
@@ -355,8 +355,8 @@ requesting review. Testing conventions are documented in
 [docs/releasing.md](docs/releasing.md).
 
 An edit is not finished until the file's own documentation has been read
-again and corrected where the change made it false: the module header, the
-doc on the item, the comment over the line.
+again and corrected where the change made it false: the module doc, the doc
+on the item, the comment over the line.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -354,9 +354,9 @@ requesting review. Testing conventions are documented in
 [docs/testing.md](docs/testing.md), and the release procedure in
 [docs/releasing.md](docs/releasing.md).
 
-An edit is not finished until the file's own documentation has been read
-again and corrected where the change made it false: the module doc, the doc
-on the item, the comment over the line.
+An edit is not finished until the changed file's own documentation has been
+read again and corrected where the change made it false: the module doc, the
+doc on the item, the comment over the line.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ its own paused time context) and does not execute subscription sources — it
 observes only the *declared* set via `subscription_ids`. See the
 [`tears::testing`](https://docs.rs/tears/latest/tears/testing/) module docs for
 the full contract, including deterministic time without `TestStore`. Worked,
-runnable tests ship with three of the examples:
+runnable tests ship with these examples:
 
 ```bash
 cargo test --example command_timeout_retry
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -41,7 +41,7 @@ Where each kind belongs:
 | why the code below is written this way | the comment | prose, as now |
 | a number some list in the same artifact has to agree with | — | do not write one |
 
-**Turning a claim about the file's own behaviour into a test is a conversion,
+**Turning a claim about the file's own behavior into a test is a conversion,
 not a deletion.** A comment claiming what happens three functions away is
 checked by a reader holding two places in their head at once; a test claiming
 it is checked by a run. Run the converted claim — `cargo test --example <name>`,
@@ -49,14 +49,18 @@ with whatever `--features` that example's own header names — and watch it pass
 then break the code it describes and watch it fail: a test nothing reaches is a
 comment with more syntax.
 
-Check that the runs you rely on reach it, rather than assuming they do. An
-example whose `required-features` are not enabled is left out without a word,
-so read the output for `Running unittests examples/<name>.rs` instead of for a
-green summary.
+Check that the runs you rely on reach it. Naming one example fails loudly when
+its features are missing — `cargo test --example <name>` either runs the test or
+says what to enable — while a wildcard selection does not: `--examples` and
+`--all-targets`, which `just test` and CI use, pass over an example whose
+`required-features` are not enabled and succeed anyway. So when what you
+converted sits in a gated example, check that the recipe you are relying on
+enables its features; `just test` takes them from `build_features` in the
+`justfile`.
 
 **A pointer to another example may say what it is for, not how it behaves.**
 "For a larger state-management example, see `x.rs`" keeps working when `x.rs`
-changes. "`x.rs` prints it from inside the run" is a copy of `x.rs`'s behaviour,
+changes. "`x.rs` prints it from inside the run" is a copy of `x.rs`'s behavior,
 and `x.rs` is free to change without knowing the copy exists. Where two examples
 exist to be compared, the comparison is stated once — in the one that exists
 because of it — and the other carries a pointer.
@@ -64,9 +68,9 @@ because of it — and the other carries a pointer.
 **A number that a list has to agree with is a fact a reader can check by
 counting**, and it goes stale in silence when the list grows: "three things"
 over four items is wrong and nothing fails. A measurement a run produced is a
-different thing and stays — the interleaving and hook-call counts under
-"Process-Global Panic Hook Tests" are load-bearing, and no later edit changes
-what a past run measured.
+different thing and stays: an interleaving count, a number of hook calls
+observed where one was expected, a number of consecutive green runs. Nothing a
+later edit does changes what a past run measured.
 
 These rules are about examples, which is where the failure showed up. Read the
 other way they are why rustdoc rarely has it: an item's documentation sits
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -25,44 +25,53 @@ the contract:
 ## What an Example May Assert
 
 An example teaches with prose as much as with code, and prose is claims. A
-claim keeps itself true only when what it describes is on the same screen.
-Every other kind has its truth somewhere else, and a copy of it in an example
-is a copy that will be wrong after the next edit to the thing it copied.
+claim stays true by itself when it describes the code it annotates. When it
+describes something else — a framework item, another example, a different part
+of the same file — its truth lives where that thing lives, and the copy in the
+example goes stale the next time that thing changes.
 
 Where each kind belongs:
 
 | Claim | Owner | Form in an example |
 | --- | --- | --- |
 | what a framework item does | that item's rustdoc | a link, not a restatement |
-| when and why to reach for a shape | the guide in `docs/` | a reference |
-| how two examples differ | the one that exists to be compared | nothing in the other |
+| when and why to reach for a shape | whichever document in `docs/` owns the subject | a reference |
+| how another example behaves | that example | a pointer to what it is for, and no more |
 | what this file does elsewhere in itself | a test | the test's name and its assertions |
 | why the code below is written this way | the comment | prose, as now |
-| a count of anything | — | do not write one |
+| a number some list in the same artifact has to agree with | — | do not write one |
 
-The fourth row is the one that does the work, and it is a conversion rather
-than a deletion. A comment claiming what happens three functions away is
+**Turning a claim about the file's own behaviour into a test is a conversion,
+not a deletion.** A comment claiming what happens three functions away is
 checked by a reader holding two places in their head at once; a test claiming
-it is checked by `cargo test`. The explanation is worth keeping — move it to
-where something fails when it stops being true.
+it is checked by a run. Run `cargo test --example <name>` on the converted
+claim and watch it pass, then break the code it describes and watch it fail: a
+test nothing reaches is a comment with more syntax. `just test` and CI reach
+every example's tests; confirm with those two rather than with whatever a
+narrower invocation happens to cover.
 
-The third row has a matching consequence: a file may not describe another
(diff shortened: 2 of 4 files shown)


akiomik:
All eight fixed in b226da6. Four of them were the section doing the thing it
describes, which is the most useful result this review could have produced: a
rule that cannot survive its own statement is not one I should be asking anyone
to apply.

**1 — "checked by `cargo test`".** Verified: `cargo test -- --list` names no
example target, CI reaches them with `--examples` and `just test` with
`--all-targets`. The paragraph now names `cargo test --example <name>` for one
and those two for all, and asks for the check — pass, then break the code and
watch it fail — rather than explaining what each invocation covers.

**2 — counts of an open PR.** Removed. My "closed record" defence was wrong on
its own terms: #360 is open, and this PR unblocks it, so both numbers were due
to move. The evidence reads the same without them.

**3 — "never carries anything the sentence needs".** Refuted by this file three
times, as you show. The distinction that holds is a number some list has to
agree with versus a number a run produced, and the panic-hook measurements are
named as the second so the rule cannot be read onto them.

**4 — unscoped prohibition.** Carries its scope in the sentence now.

**5 — in-tree violations.** `views.rs` turned out to be the useful case: it says
what `dashboard.rs` is *for*, not how it behaves, and it should keep working. So
the rule distinguishes those, and `views.rs` is cited in the commit as the
correct form rather than grandfathered. That left one real instance —
`command_timeout_retry.rs`'s "four kinds" over four bullets — fixed here rather
than excused, since a convention the tree contradicts on landing is not one.

**6 — a guide that does not exist.** The row says "whichever document in
`docs/` owns the subject", which asserts no particular one.

**7, 8 — positional row references, and "the same screen".** Both were the
section's own failure class: a positional reference is redirected by a later
insertion, and a viewport is not a property of the code. Entries are named by
subject, and the criterion is adjacency to what is annotated.

On the mis-citation: "the mirrors' eight rows" is in `justfile:173`, not
`testing.md` — and it is machine-enforced by that `grep`, which makes it the
strongest case for your version of the count rule rather than mine.

`typos` passes. `cargo test --example command_timeout_retry` still passes.

akiomik:
All four fixed in 4db42ce. Three were the previous commit's, and they share one
shape worth naming: I removed a defect and replaced it with a claim of the same
class, one level up.

**1 — a count replaced by a framework claim.** Correct on both counts. It is
false (a cancelled command produces no outcome; one with neither modifier just
resolves) and it is the row this section assigns to the item's own rustdoc.
Dropping a number never required promoting the sentence — "several" is local and
stays true.

**2 — coverage the sentence cannot carry.** Right, and the conditional is
invisible: today every example's `required-features` are inside what those runs
enable, and nothing keeps that so. I verified the failure mode you describe —
`cargo test --examples --no-run` without the features exits 0 having built
neither `websocket` nor `http_todo`. The paragraph asks for the observation now
(the example's own `Running unittests` line) rather than promising the runs
reach it.

**3 — a recipe that errors.** Reproduced:

```
$ cargo test --example http_todo --no-run
error: target `http_todo` in package `tears` requires the features: `http`
```

It now says to pass whatever `--features` the example's header names, and I ran
the corrected recipe on both a gated and an ungated example rather than only
reading it back.

**4 — the outcome claim that outlived its counts.** You are right that removing
the numbers did not remove the dependency. "Claims it moved into tests stayed
fixed" is a forward claim about a PR still under review, so it is gone; what
stays is the mechanism and the observation of it, which is about rounds that
have already closed.

On `views.rs`: agreed, and it is the reason the rule distinguishes a pointer to
what another example is *for* from a copy of how it behaves. Had it come out the
other way the rule would have been forbidding the one cross-file sentence in the
tree that works.

`typos` passes; `cargo test --example command_timeout_retry` and
`cargo test --example http_todo --features http` both run and pass.

akiomik:
All six fixed in 87f89e4, plus the body rewritten.

**1 — "without a word" false of the command above it.** Verified both halves:
`cargo test --example websocket` errors and names the features, while
`cargo test --examples --no-run` exits 0 having built neither gated example. The
warning was reaching the reader who would hit a hard error and missing the one
relying on the silent form. It names the forms now, and points at
`build_features` in the `justfile` for what a recipe enables.

**2 — cargo's output string quoted into a repository document.** Agreed, and it
is the #350 category exactly. Nothing here owns that format, so the instruction
would have broken with a toolchain and nothing failing. The check survives
without it.

**3 — the exemplar trimmed from a line the tree still had untrimmed.** Fixed in
`views.rs` rather than by widening the rule to fit it. Your scenario is the
argument: the description is of `dashboard.rs`'s structure, and the README's
example list already owns that description, so the pointer keeps the pointing
and the description stays in one place.

Noting the reversal you flag: it was in range last round and is not now, because
the section quoting a trimmed version of that very line changed what the tree
owes. That is a real cost of putting an exemplar in a rule, and the right call
is the one that leaves them agreeing.

**4 — the carve-out anchored to a heading.** Re-anchored to what the numbers
are. It was the stale cross-reference this section forbids two paragraphs above,
which is now the third time this branch has written its own failure class into
the rule describing it.

**5 — spelling.** Both were mine; the file's other nine win.

**6 — the body arguing about deleted text.** Rewritten against the current diff.
The second point I asked for argument on is now the real open question: the
section names a hazard in cargo's selection behavior, which is the kind of thing
this repository's docs avoid. It is there because the hazard is the reason for
the instruction beside it. If that trade is wrong, the sentence can go and leave
only the check.

`typos` passes. `cargo test --example command_timeout_retry`,
`cargo test --example views` and `cargo test --example http_todo --features http`
all run.

akiomik:
Took option 3 and applied all four in d026db5. Not option 2: the stopping rule
does not need relaxing if the thing generating findings is removed instead.

**Why 3 rather than 1.** Three of this round's four — the rustdoc aside, the
carve-out anchored to another section's title, the evidence on an open PR — are
not errors in the rule. They are the section holding claims *beyond* the rule,
which is the shape the rule describes, arriving for the third round running.
Round two had the same three (positional row references, "same screen", the
counts); round three had two more (the heading anchor again, the outcome claim
again). The table and the conversion paragraph have generated one finding
between them across four rounds. Everything else has generated the rest.

So the document had the defect it was written to name, and the fix is the one it
prescribes: cut the argument, keep the rule. The section is 28 lines shorter and
now asserts nothing outside itself except three repository facts and one about
cargo, which the body flags as the open trade.

I expect this to converge because the generator is gone, not because the
remaining text is more carefully written than the removed text was.

**1 — a check that cannot fail here.** Correct, and the misreading you describe
is the damaging one: a contributor concluding their gated test does not gate
merge, when it runs on every leg. `just test` and CI both pass `build_features`,
and `just test-doc-packaged` checks that list against the crate's feature table.
The paragraph names what is actually exposed — a bare `cargo test --all-targets`
— instead.

**2 — a prescribed form nothing resolves.** Right that `cargo doc` does not
document examples, so the link is text that keeps looking like a reference after
the item is renamed. The row asks for the item to be named rather than restated
and says nothing about syntax.

One correction: the tree does have the precedent. `examples/command_cancellation.rs`
uses `[`CommandId`]` and `[`CancelPolicy::CancelInFlight`]`, and
`examples/panic_hook.rs` uses `[`tears::install_panic_hook`]` and
`[`Runtime::run`]`. That makes your point stronger rather than weaker — those
four links are already unverified today, and the previous text would have
recommended more of them.

**3, 4 — the aside and the evidence.** Both gone with the rest of the argument.

`typos` passes.

akiomik:
All six fixed in cd18a80. Four by deletion, which is why I have not used the
disposition route you offered — recording "won't fix" would have cost more than
removing the sentences did.

**My hypothesis was wrong, and here is the correction.** I cut the *arguments*
last round and kept the *references*. The paragraphs that survived still had
`justfile`, the CI workflow, `docs/`, and a review history as their subjects.
Argument versus mention is not the variable; whether the subject is elsewhere
is. Four of this round's six are that, including `:38` returning through a
rewording exactly as you say — the phrasing changed and the hole did not.

What is left has one subject: the prose an author is about to write. The table,
the conversion instruction, the pointer rule, the count rule. Nothing in the
section now names a file, a recipe, a workflow, or a pull request.

**1 — the count row prohibited more than its failure.** `examples/websocket.rs`
is the useful case: a comment naming two areas directly above the two
constraints that make them is seen by whoever changes them. The row is about a
number that *heads* a list, which is the shape all four original failures had,
and once it says so the measurement carve-out is unnecessary and goes too.

**2 — the row pointing at a document that does not exist.** Dropped rather than
reworded again. The rows for a framework item and for the file's own behaviour
cover what it was reaching for, and #360 needing a guide is not a reason for
this table to name one before it lands.

**3, 4 — the tooling paragraph.** Gone. The instruction above it is complete
without it, and both halves lived elsewhere: which features a recipe enables is
in the `justfile`, and cargo's selection behaviour is cargo's.

**5 — the README list.** Fixed here, since this rule is what turns it into
debt. It no longer enumerates. Note this will conflict with #360, which edits
the same paragraph; I will rebase that branch rather than the other way round.

**6 — the failure a reader cannot follow.** The clause is gone; the scoping
sentence works without it.

On cost and convergence: 357k is real and I am not going to predict a round
count again. What I can say is what changed structurally — the section's
subjects are now all internal, which was not true of any previous revision,
including the one I claimed had removed the generator.

akiomik:
All four in 470169a. Finding 2 decided the shape of the rest.

**2 — the count rule blessing what this branch first deleted.** Deleted the
rule. Its boundary moved in four of six rounds and its current position
permitted "four kinds of asynchronous outcomes" over four bullets, which is
`src/kernel.rs`'s "two things this module deliberately does not have" over two —
in the tree, correct, and untouched. I verified both.

It was also never needed. "Three things stay at the root" over four paragraphs
is a claim about a different part of the same file, which the opening principle
covers. The row was a special case that spent five rounds failing to name its
own edges, and the `command_timeout_retry.rs` edit made under it is reverted:
the count was accurate, and nothing now says to remove it.

I would rather not keep proposing a boundary. If a count deserves a rule it
deserves one that survives being written down, and I have four rounds of
evidence that I cannot write this one.

**4 — the item row reaching the example's own subject.** Same failure mirrored:
`command_cancellation.rs` describes what `CancelPolicy`'s variants do, which is
the example's entire subject, and `panic_hook.rs` does the same with the hook it
is named for. The row is about an item picked up along the way and restated as
the reason for a line — the shape that went stale in #360 — and it says that
rather than reaching what an example exists to demonstrate.

**1 — the placeholder.** You are right that the unusable block is worse. Three
of nine examples carry tests, two of the rest do not build without features, and
the others answer `running 0 tests` with no way to tell why. The names are back
with a lead-in that no longer calls them the command examples, which `dashboard`
is not. Verified the three by counting `#[test]` across `examples/`.

**3 — the unnamed comment.** Gone with the count paragraph.

The section is four rows and three paragraphs now. Both rows that have generated
findings are removed; the two that have held — another example's behaviour, and
this file's behaviour elsewhere — are the two that came directly out of #360's
failures.

akiomik:
Taking neither option, and here is why.

Option 1 is the one I have already taken four times. The count row moved its
boundary in four rounds; the item row in three. Each narrowing was a fit to a
counter-example you had just found, and each produced the next one. Fitting the
rule to the convention a fifth time has no better prospects than the first four.

The thing that collides with the convention is not the wording of any row. It is
having a taxonomy at all. So the table is gone in 8d7d363, and what is left is
the two rules that held across all seven rounds without ever needing an edge
case:

- a file does not describe another file's behavior
- a behavioral claim a reader would act on is better held by a test

Plus one sentence that is not a placement rule, which is the part I got wrong
from the start. Of the prose failures in #360, the ones a placement rule could
have prevented are the cross-file ones. The rest — including the header bullet
you cite, "every child here has runs of its own", which was false about two of
the four children — were sentences describing code their author had changed
without re-reading them. That is an obligation at edit time. Writing it as a
question of where a claim lives was a category error, and it is why the table
kept having to exempt the places where description is exactly the point.

The header convention is now named as something neither rule touches, since a
reader arriving with the table would have concluded the opposite. That closes
findings 1 and 2 without either amendment you offered: with no row about a
framework item, nothing flags `command_cancellation.rs:254`, and with the
convention named, nothing flags the bullet lists. I verified all nine headers
use them.

**3** — the README count. Fixed; the list below carries the same information and
cannot disagree with itself.

**4** — the scope. It says what is outside now: nothing here is a rule for
`src/`.

The PR body has a new section asking the question I think is actually open: two
rules and an obligation may be too little to be worth writing down. That is a
fair reading and I would rather hear it than pad the section back out.

akiomik:
All four plus (a) and (b) in 6a5b2cc.

**(a) — you are right that it was not adjudicated, and it stands.** I read your
finding as being about the example-to-example pointer and answered that instead;
the question was an example describing `src/testing.rs`, which the rule as
worded — "a file does not describe another file's behavior" — flags outright.
`command_cancellation.rs:254-256` is that, and it is an example describing the
library, which is what an example is for.

So the rule says "another *example*" now. That gives up its coverage of an
example restating a library mechanic, deliberately: drawing the line between a
mechanic an example demonstrates and one it merely uses is the thing I failed at
three times, and the re-reading obligation is what carries it instead.

**(b) — correct, and the claim is wrong.** `examples/signals.rs` opens with
prose and its bullets are under "Then try:". The grep I checked "all nine" with
matched those. It is in a pushed commit message on a branch under review, so I
have noted it in the PR body rather than force-pushing over your review anchors.
Nothing in the section repeats the count.

**3 — the scope line cutting off the obligation.** This is the one that
mattered, and your reasoning is the reason I took it: re-reading is the failure
placement never reached, it is the largest group in the review this came from,
and it is not example-specific. Someone changing a runtime and leaving a false
sentence in its module header is exactly who it is for. The rules stay scoped,
the obligation does not, and the section counts three norms.

**1 — the authority.** `Cargo.toml`'s `required-features` is what binds; the
headers agree with it today and nothing keeps them agreeing. Named.

**2 — the counter-example with no antecedent.** Self-contained now, and the pair
carries the criterion — a pointer names the other file and what it is there for,
a copy states something that file does — rather than leaving it to be inferred
from a case only #360's readers saw.

**4 — "larger".** Both the exemplar here and `views.rs` say "another", which
neither file can falsify.

**On your answer to the body's question:** taken, and the second reason is the
one I had not weighed properly. I had been treating the obligation as the weakest
of the three because it is not enforceable, when it is the one covering what the
enforceable ones cannot. The body now asks the question I think is left: whether
the two rules should be unscoped as well.

akiomik:
All three in 40b4aa3, and here is the convergence answer with the analysis behind
it.

## Where the 43 findings came from

- claims about something outside the section (cargo, `justfile`, CI, `docs/`,
  #360, tree files, headings) — about 20
- a rule whose boundary does not match the tree — about 12
- the section committing the failure it describes — about 5
- hygiene — about 4
- reachability — 1

## Why it has not converged

One sentence drew five of those across four rounds: how to run an example's
tests.

| round | finding | what I did | what happened |
| --- | --- | --- | --- |
| 2 | "`just test` and CI reach every example" is conditional | removed it | |
| 4 | the replacement check cannot fail | removed the paragraph | |
| 8 | features authority is `Cargo.toml`, not the header | named `Cargo.toml` | |
| 9 | `Cargo.toml` has entries for two of nine | — | here |

Remove the anchor and it reads as vague; add one and it is a claim about
something I do not control. That is an oscillation, not a defect being fixed,
and the sentence was never a norm — invoking cargo is not what this section is
for.

Meanwhile the three norms themselves have not drawn a boundary finding since
round 7, when the taxonomy went. Round 8 was about their scope and round 9 about
their placement; neither was about what they say.

## What I did

The operational half is gone. The point it carried — a converted claim has to be
able to fail — survives without an invocation, and this is the document about
running tests.

The obligation is its own section with the document's opening carrying it, which
closes finding 3 structurally rather than by widening a sentence: it was
repo-wide and reachable only through a heading about examples.

`views.rs` borrows `dashboard.rs`'s own words, per your finding — a fact about
that file's self-description rather than a category I invented. "Another
state-management example" classified `views.rs` as one when nothing does.

## Prediction

The example section is now two norms and nothing operational; no sentence in it
names cargo, a recipe, a workflow, a manifest, or a pull request. I verified
that with a grep rather than by reading it back.

If round 10 finds more, that is worth having: findings on the two norms
themselves would say the norms are unwritable, which is different information
from the last nine rounds, all of which were about what surrounded them.

akiomik:
Took the third option, and finding 1 is why rather than the round count.

**1 decided it.** The rule forbade an example describing another example's
behavior and its exemption required exactly that — "the comparison is stated
once, in the one that exists because of it" is one example describing another.
And the exemption serves a pair that is not in the tree. A rule that cannot be
applied as written, with an escape hatch for a case that has not arrived, is not
one I can ask anyone to follow.

Reading the ten rounds back: the section went from a table of six rows to four,
to two norms, to two rules, each reduction because that much was defensible and
no more. Following that curve to its end leaves one paragraph. I would rather
land where the curve was going than stop one round short of it and call the
remainder issues.

**2 is fixed and is the reason the paragraph is worth keeping.** You are right
that it declared itself over prose wherever it describes code and then asked
only for the file's own documentation. This document pinning `load.rs`'s
`[3, 2, 1, 0]` and `producer.rs`'s gauge trace is the case that proves it —
nobody editing those files is brought back here. It asks for the search that
reaches them now.

**3, 5, 6** go with the section. **4** is fixed: the README pointer says what the
document holds.

**On the PR body citing `docs/composition.md`.** Correct, and it was the same
defect as the rest — an argument resting on a file no reviewer here can open.
The body is rewritten against what the diff now is, and the PR is retitled,
since "What an Example May Assert" is no longer in it.

**On your read of the convergence shape.** I agree with it, and it is why I did
not take option 2. The seams you identify — four of this round's six created by
last round's restructuring — are the same generator as #360's, reproduced in a
document written to prevent it, at one tenth the size. That is not an argument
for one more round of careful writing. It is an argument that the thing being
written was not writable in that form, which the surviving paragraph is not,
because it asks for an action rather than drawing a line.

If this is still too much, deleting it is a fair answer and the PR becomes two
line fixes. I do not think it is: it is the only thing here that addresses the
largest group of the failures it came from.

akiomik:
All four in 187c797.

**1 — the method the repo had already rejected.** This is the sharpest finding
in eleven rounds, because it is not about my wording: I prescribed a search by
name, and `pre-review-checklist.md` records from #240 and #246 that a name is
exactly what those searches miss — a stale restatement usually shares no
vocabulary with the sentence that made it stale. I wrote a weaker version of
something the repository had already learned, without looking for whether it had
learned it.

I did not take your first option. Restating the checklist's method here would be
one file copying what another owns, in the paragraph that exists to say not to,
and the checklist prunes and grows, so the copy would go stale by design. It
points at it and says what it is there for. What is left is the same-file half,
which needs no method: the documentation is in front of whoever is editing the
file.

**3 — the count I introduced.** Right, and it was the same class as the rest:
a number describing another file's sections with nothing to notice when a second
arrives. Both README entries now describe what is there without counting it.

**2 — the Contributing pointer.** Fixed with the same wording, so the two
entries cannot drift from each other either.

**4 — one of two identical descriptions.** Took the third answer and reverted
`views.rs`. Its trim rested on the cross-example rule, which is gone; on the
obligation alone the same reasoning reaches the README's example list, whose job
is to describe each example. Applying it to one of the two was the inconsistency,
not a fix, and exempting the index would have been a boundary — which is the
thing eleven rounds have shown I should stop drawing.

The PR is now one section and two pointer lines. The body is rewritten against
that, and no longer claims an in-tree line as a rule application.

akiomik:
All five in 0efc24b.

**3 is the one that decided the shape.** My fix for last round — point rather
than copy — produced a new instance of the class the paragraph is about, and
your reading of why is exact: the checklist says of itself that it retires
examples once one principle holds them, so "records which searches found stale
restatements" was written to go stale. Pointing is not a way out of making a
claim about another file; it is a smaller one.

So the cross-file half is gone rather than aimed somewhere else a third time.
That leaves the case uncovered, which I would rather say plainly than paper
over: the search that finds it is one the RFC checklist worked out, and I have
now been wrong about it twice.

**4 matters more than its severity.** The same-file half had been clean for
eleven rounds and I had been treating it as the settled part. It was not — it
carried an absolute the repository's own conventions refute, since a change
history in Git is exactly how a reader finds out what was true when. The
sentence claims what it can: the build and the tests do not catch it.

Taken together with 3, the generator is not the cross-file half. It is
explanation. Every round's fix added a sentence in support of the obligation,
and every sentence in support is a claim. Twelve rounds have now found a defect
in each one. The support is gone; the obligation is two sentences and asserts
nothing beyond what an editor can see.

**1** — the pointers limited the document to prose around tested code, and the
obligation is not limited to tested code. That reader — someone editing a
runtime — is who the section was moved out of an examples heading to reach.
Fixed in both places with the same words.

**2** — the checklist link is gone with the half it served. The audience
mismatch you found between the two pointers goes with it.

**5** — rewrapped to the paragraph.

The body now states what the section deliberately does not cover, rather than
implying the cross-file case is handled.

akiomik:
Both in d05a507; they were one overreach in two places, as you say.

**1.** Verified before fixing: `cargo doc --no-deps` runs under
`RUSTDOCFLAGS: -D warnings` in the doc job, `src/` holds 563 intra-doc links,
and the doctest job's own comment says that without it an API rename could break
every snippet and leave CI green — which asserts that with it, one does not.
My premise was false of both classes, in a section about sentences that quietly
stop being true.

The fix makes the contrast instead of overstating past it: a link that stops
resolving fails a build and an example that stops compiling fails a test, and a
sentence is neither. That is narrower, true, and it states the reason the
obligation exists rather than a slogan around it.

**2.** Same overreach plus the attachment — "prose that no test can carry" reads
as prose, and doctests carry part of that prose. It is the obligation no test
can carry.

Recording what this round means for the read on convergence. Twelve rounds
found defects in the section's structure: what it was, how much it claimed,
where it lived, what it pointed at. This round found none of that. Both findings
were the accuracy of one factual sentence, and both were mine to check against
the repository before writing — the CI workflow says so in its own comment,
which I had not read.

That is a different failure from the previous twelve, and a smaller one.

akiomik:
Took the recommendation. b3ad649 drops the two explanatory sentences; 1, 2 and 3
go with them, and 4 is fixed.

Verified all three before cutting, since the cut rests on them being right:
`.github/workflows/` holds `ci.yml`, `release.yml`, `spelling.yml` and no link
checker, so the two Markdown links this change adds break in silence; fourteen
`rust,ignore` blocks are compiled by nothing; and `tests/api_surface.rs`
enforces two prose rules from `docs/api-guidelines.md` mechanically, described
sixty lines below in the same file I was writing the absolute into.

Your table is the argument. Three rows, two of them with a checked and an
unchecked side, and I tried twice to state it in two sentences. The instruction
has drawn no finding in fourteen rounds and now stands alone. A reader who wants
the reason can work it out; what they cannot do is act on a reason that is
wrong.

**4** — the pointers claimed coverage of prose describing the code, and the
obligation is a file's own documentation. Both quote the section heading now,
which is the most stable thing there is to point at.

On the pattern you name: this is its fourth appearance at the same position, and
the first time I have applied it without being shown the instance. The three
before — rounds 5, 8 and 12 — I only saw after a finding pointed at them. That
is the difference worth recording, more than the count.

akiomik:
All four in 5fdfdda. 1 and 2 turned out to be one problem and it had a move as
its answer rather than an edit.

**1 + 2.** Read together they say the obligation has no home in
`docs/testing.md`: the one non-test item in a document titled for tests, above
the section on test placement, whose reader — someone editing a runtime — has no
reason to open it, so its discoverability rested on two pointers, one of which
advertised a contributor's obligation to downstream users learning to test their
own applications.

So it moved to the Contributing paragraph, which is where that reader is and
which already gives an instruction rather than only a direction — it says to run
the RFC checklist before requesting review. `docs/testing.md` is back to
untouched, both pointers are back to what they said, and there is no longer a
description of another document to keep true. That last part removes the class
that produced findings in rounds 9, 10, 11, 12 and 14.

**3.** Real gap and it is the obligation's own. Someone who re-read a header
their change had falsified, decided against touching it, and pushed had
satisfied the sentence as written — which is the failure it exists to close. It
asks for the correction now.

**4.** Filed as #363 rather than written into the section, for the reason you
give: I wrote a boundary sentence for exactly this in round 11 and round 12
found it. The issue records what is known — including that the checklist's own
entries say a search by name or by changed sentence goes straight past these —
and states what would close it without deciding the design. `docs/testing.md`
being the largest instance is in there, with its own note that an edit adding
`last()` to `Readings` would pass the whole run.

On your last point: you flagged that 3 and 4 both add words to a section I have
been cutting for four rounds. That was the right thing to flag. The answer was
that 3 belongs to the obligation and is one clause, 4 belongs in an issue, and
1 and 2 together meant the section should not exist where it was. The diff is
five lines in one paragraph.

akiomik:
Both in 7fc86d1 and the body.

**1.** Fixed. Checked the vocabulary rather than taking it: `module docs` is in
this same file seventy-six lines above the rule, and `module doc` is in two
RFCs. The rule's own subject was drifting from the tree, which is the thing it
is against.

Your correction to the agent's count holds — `module header` appears twice, the
other being `src/kernel/conformance/observability.rs:348`, a code comment about
a different thing. That does not change the finding and it does change what a
contributor's grep returns, which was the finding's mechanism.

**2.** The body said "Nothing else" and the diff had a second hunk. It now names
it: `:284` said the three listed examples are the command examples, and
`dashboard` is not one. That line has been in the diff since `cd18a80`, and the
classification rule that motivated it was removed two commits later, so it has
been sitting there unattached to any argument — which is exactly why "Nothing
else" was able to hide it.

Also corrected: one sentence, not three. It is three lines.

The diff is now the sentence and that word.

akiomik:
Fixed in 5cf0448: "the changed file's own documentation".

The antecedent problem is worth naming for what it is. The sentence was moved
into that paragraph last round precisely because the paragraph is where
contributors are — and the move put it after four document names, which is the
one context that inverts its range. A rule about keeping prose true, made
ambiguous by where it was placed to be found.

Also corrected in the body: the Testing hunk is a phrase, not a word. That was
yours from last round and it was right; I let it through when I fixed the rest
of that paragraph.

The diff is one sentence and one phrase.

akiomik:
Both in e71c46f.

**1.** Verified before fixing: `docs/rfcs/README.md` says one or the other is
fixed explicitly when code and an accepted RFC disagree, and doc comments under
`src/` cite `INV-` on 256 lines and `RFC 00` on 475. So this is not a corner —
it is most of what the rule would be applied to, and "corrected" pointed the
contributor at the sentence rather than at the disagreement.

It says "fixed on one side or the other, explicitly" now, which is the RFC
directory's own wording. Borrowing it rather than paraphrasing is deliberate: a
reader who has seen that rule will recognise this one as the same rule reaching
further, and there is no second phrasing to drift.

This is the second time an existing repository rule turned out to be both more
accurate and already written down — round 11 was the pre-review checklist's
search method. Both times I wrote a weaker version without looking for whether
one existed. That is the check I should be running before adding any rule here,
and it is cheaper than the round it costs.

**2.** Right that a field's `///` is neither an item's doc nor a comment over a
line, and that this tree uses "item" strictly. The list reads every doc comment
under the module doc, and the comments in the body, which has no gap to fall
through.

